PR #13, opened 2017-08-20
Synthesis AI -- Pathfinding Robots

pfhgetty:
## **Synthesis AI**
**The Problem**
At times when fast turnarounds and continuous integration are required, as happens so often in the chaotic environment of FRC competitions, FRC teams frequently need to test robot designs as quickly as possible and make changes accordingly. Testing things like defense and maneuverability, however, usually requires more than one robot in the simulation, and LAN/online multiplayer systems can sometimes be overkill or simply not feasible in times such as when a single defense mechanism needs to be tested a couple times, or when there is no internet connection to test the robot with another player.

**The Solution**
AI pathfinding robots solve this unique problem. When resources, team sizes, and time get low, AI pathfinding robots offer nearly automated testing solutions at a low expense. Having an AI robot chase and ram into your team's robot can be used to test your robot's sturdiness, while ramming into an AI robot to knock it off a pre-specified course can be used to test your robot's defensive capabilities. This feature adds realism and utility to the gameplay of Synthesis at very low expense. It should be noted that our AI pathfinding robots were not created _to replace_ Synthesis's upcoming multiplayer features, but to simply fill a gap and increase quality of life for many FRC teams which would benefit from being able to test their robots against autonomous agents for faster and more automated experimentation. This feature helps better emulate real gameplay of FRC matches, **because what's an FRC match if the field only has you**?

synthai pathfinding demo
*SynthAI extends Synthesis gameplay by adding autonomous agents to test robot designs against*

**Synthesis AI**
Synthesis AI is a lightweight pathfinding extension, created for the Synthesis Code Sprint. It was made to be customizable, flexible, and powerful without compromising performance. Currently, the add-on **adds only 0.5 seconds of loading time** to most scenes, and more than 90% of runtime lag is caused solely by the physics calculations required to add another robot into the scene. Simply put, SynthAI is a fast solution to a big problem. It makes Synthesis gameplay highly interactive while still maintaining performance.

Synthesis AI is highly extensible and neatly integrated into the existing Synthesis architecture. In creating Synthesis AI, **as little of the original architecture was touched as possible**, adding only a couple method definitions for UI and control purposes, and adding another few lines to tag field elements. Most of the work for Synthesis was in creating new code on top of the existing architecture, rather than in changing the code that was already there. Synthesis AI code was also **created solely by students** -- not a single mentor touched the code for Synthesis AI.

**Current Features**
- Ability to set AI robot spawn point to active robot's current position.
- Spawn any robot in robot list as an AI robot*
- Chase Behaviour which has AI robot chase the player's robot in an attempt to ram and stop them.
- Roam Behaviour which has AI robot move to random locations around the field.
- Dummy Behaviour which has AI robot not do anything at all.
- Auto pathing using Unity NavMesh allows AI robot to path around obstacles which have the SynthAITag component
- Quick setup AI manager component which finds the field object tagged by the field tag name and auto generates NavMesh on field. Should be attached to StateMachine.
    --> Customizable max speed and look ahead distance for robot pathfinding are also within this component.
- Auto regeneration of NavMesh during runtime.
- (On Scene.unity) SynthAI UI elements including a "Clear ALL Robots" button.
*Some robots have bugs due to the Unity forward axis pointing in the wrong direction. More details below.
## **Integration and Extensibility**
**Highly Extensible**
We defined a very specific scope for our project, in order to allow us to finish it by the end of the Code Sprint, but by using object oriented principles, we made our code highly extensible for any coder to be able to add their own custom behaviour with minimal effort. The abstract *BaseSynthBehaviour* lays the groundwork for an AI robot, simply extending from it and pointing the Nav Agent towards a direction on the *AIUpdate()* loop will have the robot pathfind towards that location. Because of this object oriented implementation, it is very easy to define your own behaviour, from having the robot pathfind its way towards goals, or making it avoid the player at all costs. Three behaviours, the *Chase, Dummy*, and *Roam* behaviours, have been coded as examples of how to extend the *BaseSynthBehaviour* class to make your own AI behaviour.

**Integration Details**
In order to **reduce coupling**, our *IControllable* interface is used as the communication layer between each AI Behaviour and each AI Robot. AI Behaviour components automatically search for an *IControllable* component on the game object it is attached to in order to send commands to the robot game object. The *IControllable* interface defines methods which allow the AI Behaviour to move and turn the robot. **Any robot with a working implementation of the IControllable interface can be controlled by the AI Behaviour**.

In the current version of SynthAI, an *AIRobot* component extends from the *Robot* component and implements an *IControllable* interface. This *AIRobot* component is what is spawned by the *SynthAIManager* when "Spawn AI Robot" is pressed. The *AIRobot*'s current implementation of the *IControllable* simply sets a virtual "Horizontal" and "Vertical" input using a Dictionary (C#'s version of a Hashmap). In the Fixed Update loop, instead of using *DriveJoints.UpdateAllMotors*, a custom *DriveJoints.MoveAIMotors* method takes in the virtual input, and moves the robot accordingly.

TL;DR
Through our "*IControllable*" interface, SynthAI can be easily implemented into the current version of Synthesis without being invasive to the existing code structure.

**Known Issues**
- Robots with Unity forward vectors facing the wrong direction will not steer correctly. This is the case for some robots, but the Dozer robot, with its forward vector facing in the direction it accelerates, works well with SynthAI. CAD models that are output correctly should not have a problem.

**The Future of SynthAI**
Although the scope of this project was large, we hope to make SynthAI even larger. With the existing code base, it is possible to have AI robots test other robots by maneuvering around fields and finding their way around obstacles. However, our code base can be easily extended and used for other things. Features that could be implemented (1-2 months) include:
- Having the robot use its other mechanisms, not just its wheels.
    --> This would actually be very easy to implement, as the *IControllable* interface could be used as a Robot controller. Simply adding a *MoveManipulatorJoint()* method to the *IControllable* would allow behaviours to access a robot's mechanisms and increase the applications of SynthAI to an almost infinite possible cases.
- Implement behaviours for specific fields
   --> This would also be fairly trivial to implement, as behaviours are already highly customizable.
- Fix the forward vector bug
    -->  This is a problem that could technically be solved by users simply outputting their robots facing forward, but for ease of use, there could possibly be a way to change the forward vector to the direction the robot moves when the virtual forward key is pressed.

Thanks for reading, we hope you have a great summer!
**Synthesis AI - IRS 1318**

For any comments, questions, or issues, please contact:
*Preston Lee - pfhgetty1337@gmail.com*

SynthAI short example
*Chase behaviour on Dozer robot. Robot on the right is pathfinding its way to the player, on the left.*

## **Working Build**
[This working build](https://drive.google.com/file/d/0B_R43AvwZu_Fd3haNUpQRFI2QUk/view?usp=sharing) can be downloaded from Google Drive. It is the full build of our project as of August 20th, 2:00 PM.

synthai ui button
The AI button is used to open the SynthAI panel.

synthai ui panel
In the SynthAI panel, the user can select which robot they want to spawn with which behaviour they want it to have. They then spawn that robot by clicking "Spawn AI Robot". The user can also set the spawn position to wherever they would like by moving their robot to the position they want to move the spawn point and clicking on "Set Spawn Point". The "Clear ALL Robots" destroys all AI controlled robots. 

**Extra Class Details**
_SynthAIManager.cs_ - To be added to StateMachine. Configures NavMesh area, and auto updates NavMesh. Also handles spawning and clearing AI robots.

_SynthAITag.cs_ - Any object with a SynthAITag component is taken account in the creating of the dynamic NavMesh. The AI robot's steering agent will attempt to steer on the NavMesh and around obstacles.

_IControllable.cs_ - An interface which operates between AI Behaviours and AI Robots.

_BaseSynthBehaviour.cs_ - The abstract base behaviour of all AI robots. Any AI behaviour should extend from this class. Holds a reference to an IControllable component.

_AIRobot.cs_ - An example of an IControllable implementation. Extends from Robot.cs, and implements the IControllable methods **void Accelerate(float acceleration)**, **void Turn(float direction)**, and **Vector3 GetPosition()** in order to allow SynthBehaviours to manipulate robot. Also handles its own initialization, and uses **DriveJoints.UpdateAIMotors()** to move, rather than **DriveJoints.UpdateAllMotors()**.

_ChangeBehaviourScrollable.cs_ - Much like the ChangeRobotScrollable.cs. Used to show user all existing behaviours, so that they can choose one. Because it uses reflection to find all existing behaviours, it can be very slow, so we only scan on OnEnable(), and not on OnGUI().


tRidha:
Thanks for the submission! We'll review all the submissions over this week and the next and let you guys know the results of the Code Sprint soon.

- Thariq

PR #22, opened 2018-06-19
Disabled HEL UI Resizability

matthew17754:
Can't see any reason why the Code Viewer would need to be sizable as nothing is scaled. 
This is just some quick formatting fixes to prevent the window from being resizable.


HiceS:
Branch deleted as per cleanup procedure.

PR #26, opened 2018-06-21
Export directory

NWalker1208:
Changed the default directory for exporting robots to "User\AppData\Roaming\Synthesis\Robots" instead of "User\Documents\Synthesis\Robots". This way it matches the default directory that Synthesis checks.

Stores a version number is settings so that users that are updating the plugin will automatically have the export directory overridden to the new folder.


HiceS:
Looks good

PR #35, opened 2018-06-22
Fixed Advanced Exporter Exception

NWalker1208:
Added an invalid data check when opening the dialog in the advanced exporter.
https://jira.autodesk.com/browse/AARD-515


HiceS:
Reviewed by Matthew, I'm gonna call that good.

Very small changes

PR #36, opened 2018-06-22
Exporter Workflow Improvements

NWalker1208:
This branch resolves JIRA issues AARD-489, AARD-490, and AARD-491. The branch name is only representative of AARD-490, as the purpose of this branch has expanded.

The wizard is now the only available exporting option. The back-end for One-Click still exists, and the advanced exporter functions as the final step where modifications can be made before the full export is done. (AARD-490)

Because of this, the wizard is now opened when the robot exporter is started. (AARD-489)

To improve workflow, the majority of the processing of the mesh has been moved to the end of the process, when the user finished the exporter. (AARD-491)

These modifications have introduced an issue that should dealt with on a separate branch. The option at the end of the wizard to open Synthesis still needs to be moved to the moment before the user clicks finish.


HiceS:
Once Braden reviews I will merge and delete branch.

HiceS:
This is a larger change so QA will for sure have to approve this before I get to merge it.

PR #39, opened 2018-06-25
Exporter progress bar improvements

NWalker1208:
Made it so that the progress bar of the exporter actually shows how much has completed and how much is left. Removed unnecessary info being presented to the user, such as the exact number of faces that have been processed.

Resolves AARD-473


MackinnonBuck:
Should I delete this branch?

PR #42, opened 2018-06-25
Exporter save settings window

NWalker1208:
This branch provides a prompt to enter robot name, field name, and full color options before saving the robot and the long exporting process. It is also compatible with the existing save buttons on the toolbar. One issue that is currently present in this code is that color settings that are chosen in the save settings prompt do not apply until the next time the exporter is run. Either that can be solved as a separate issue on another branch, or this pull request can be delayed until that issue is solved.

This branch applies to issue AARD-492.


NWalker1208:
Color settings issue has been resolved

exporters/BxDRobotExporter/BxDRobotExporter/BxDRobotExporter/StandardAddInServer.cs
```diff
@@ -334,8 +334,11 @@ private void EndExporter()
             // Export mesh as exporter is finished
             if (Utilities.GUI.SkeletonBase != null)
             {
-                Utilities.GUI.ExportMeshes();
-                Utilities.GUI.RobotSave();
+                if (Utilities.GUI.PromptSaveSettings(true, true))
```
MackinnonBuck: I don't mean to sound nitpicky but this `if` chain could be replaced by `&&`s. Functionally the same but might look a bit nicer.

PR #55, opened 2018-06-27
Invalid Joint Error Message Improvements

NWalker1208:
Changed the message displayed to the user when they attempt to export a robot with a linear joint without limits. It is now more readable to the user and displays the name of the joint. This resolves AARD-519.


NWalker1208:
Unfortunately, I can't figure out how to abort the exporter add-in after it has started, which is what should happen when the robot joint is invalid. I can't find documentation for inventor add-ins. I made it so that the exception isn't thrown to the user anymore, but we still need to implement something to stop the exporter entirely.

PR #56, opened 2018-06-27
Export wizard wheels

JoshGrace:
removed the basic robot setup page
moved drive train selection and mass to wheel setup
made wheel setup drag and drop based


HiceS:
Alright ill close the request, @NWalker1208 can either make the changes or @JoshGrace can depending on what @kirac2093 thinks.

JoshGrace:
I've fixed the typo issues and the PWM ports not being filled correctly. Also, I've fixed an issue with the PWM values being set with a minimum of 1 instead of 0.

JoshGrace:
@BradenMccoy @NWalker1208  can one of you check this on Monday?

LaurenChoquer:
I can read over it on Monday as well.

-Lauren

NWalker1208:
The exception I pointed out still exists. Also, instead of adding try-catch structures everywhere, try to find the source of the problem. I suspect that you need to check if the selected item is null, and then stop attempting to drag the item if it equals null.

HiceS:
@BradenMccoy build and test his changes, if you encounter a exception then let josh know.

HiceS:
about time

PR #60, opened 2018-07-02
Implemented multiple state machine functionality

MackinnonBuck:
It is now possible to have more than one state machine instance. A "scene global" state machine instance can be defined, and this instance can be accessed from everywhere.


MackinnonBuck:
This was a relatively simple change, most of it was back-end refactoring.

PR #61, opened 2018-07-02
Removed unused settings

NWalker1208:
Removed settings from the settings menu that the user no longer needs:
- Robot directory (will always be AppData)
- Fancy colors (set in the save menu)

This resolves AARD-546.


NWalker1208:
Will return with more commits to remove other items from toolbar.

NWalker1208:
Removed other toolbar items to resolve AARD-555 as well.

MackinnonBuck:
@NWalker1208 Could you resolve the merge conflicts so the merging can proceed?

HiceS:
@NWalker1208 take the diff.

PR #62, opened 2018-07-02
Gamepiece rotation

ambercmerrill:
Fixed bug AARD-530 (https://jira.autodesk.com/browse/AARD-530). This was a user-reported bug where gamepieces don't rotate with the robot after getting picked up. (This was is only apparent in games like 2015 and 2018 where the game pieces are not round).


MackinnonBuck:
@ambercmerrill There appear to be some merge conflicts. Could you resolve these if possible?

PR #63, opened 2018-07-02
Improved exception handling for empty assembly

NWalker1208:
An error message is no longer shown. The user enters the exporter, but there are no joints to modify. Leaving the exporter does not result in an error. This resolves AARD-529.


MackinnonBuck:
@NWalker1208 Would you mind resolving the conflicts so that the merging can proceed?

HiceS:
Resolve the conflicts that probably just came from josh's branch then ill merge

PR #72, opened 2018-07-03
Auto Generated Weight

NWalker1208:
- Improved weight configuration window
- Refactored most instances of the word "mass" to "weight" to make it consistent with the UI
- Added option to use auto generated weight in the weight window. This uses the same calculations as the old material based system
- To avoid merge conflicts, I have not modified the wizard's weight settings, but leaving the weight in the wizard at zero will cause the exporter to use the material based weights. This could be clarified on a branch focused on the wizard itself. I recommend removing the weight option from the wizard entirely, which would cause the weight to default to the auto-generated value.
- This resolves AARD-579


exporters/BxDRobotExporter/robot_exporter/JointResolver/ControlGUI/SynthesisGUI.cs
```diff
@@ -66,7 +66,7 @@ public static RuntimeMeta CreateRuntimeMeta()
     public RigidNode_Base SkeletonBase = null;
     public List<BXDAMesh> Meshes = null;
     public bool MeshesAreColored = false;
-    public float TotalMass = 120;
+    public float TotalWeightKg = -1; // Negative value indicates default mass should be left alone
```
MackinnonBuck: Generally public variables like this are made properties (the C# equivalent of adding a getter and setter). Not a huge issue but it would be nice to follow C# conventions.

exporters/BxDRobotExporter/robot_exporter/JointResolver/EditorsLibrary/SetWeightForm.cs
```diff
@@ -0,0 +1,69 @@
...
+            UnitBox.SelectedIndex = IsMetric ? 1 : 0;
+        }
+
+        private void SaveButton_Click(object sender, EventArgs e)
+        {
+            IsMetric = UnitBox.SelectedIndex == 1;
+
+            if (CalculatedWeightCheck.Checked)
+                TotalWeightKg = -1;
+            else
```
MackinnonBuck: This could be simplified:

```
if (CalculatedWeightCheck.Checked)
    TotalWeightKg = -1;
else if (!IsMetric)
    TotalWeightKg = (float)WeightBox.Value / 2.20462f;
else
    TotalWeightKg = (float)WeightBox.Value;
```
Again, not a huge deal though.

PR #73, opened 2018-07-03
Improved Clarity of Node Names in Wizard

NWalker1208:
Moved file name prepping to the end of the process. This means that nodes maintain their part names throughout the exporter process so that the user can tell which node corresponds to which part more easily. This resolves AARD-569.


MackinnonBuck:
@NWalker1208 @BradenMccoy Any updates on this? Should this pull request be closed or re-evaluated?

PR #79, opened 2018-07-05
Saving/Exporting Workflow Improvements

NWalker1208:
Resolved AARD-590: User is no longer prompted to export their robot when closing Inventor, but instead are only asked if they want to save the current configuration. This can be expanded upon later by adding a check to see if any changes have been made. Currently they are asked if they want to save the config no matter what.

Resolved AARD-596: Exporting and saving the configuration are now separate processes. Naming conventions both in the UI and in the code have been improved. The old save button is now a robot export button, and a new button was added to save the configuration. The following internal names were changed:
- "RobotSave" is now "ExportRobot"
- "RobotSaveAs" no longer exists, and instead may be achieved by using "PromptExportSettings" before calling "ExportRobot"
- "JointDataSave" is now "SaveRobotData"
- The private recursive function "JointDataSave" is now "SaveJointData"
- "JointDataLoad" is now "LoadRobotData"
- The private recursive function "JointDataLoad" is now "LoadJointData"
- "ExportMeshes" is now "LoadMeshes"
- "ExportSkeleton" is now "LoadSkeleton"

Resolved AARD-603: The robots mass and name are now stored in the assembly file. To improve backwards compatibility, the loading system was changed to continue loading even after some data could not be found. After https://github.com/Autodesk/synthesis/pull/72 is merged, the saving of isMetric can be saved as well.


exporters/BxDRobotExporter/BxDRobotExporter/BxDRobotExporter/StandardAddInServer.cs
```diff
@@ -55,16 +55,14 @@ public bool PendingChanges
 
         //Ribbon Pannels
         RibbonPanel SetupPanel;
-        RibbonPanel FilePanel;
         RibbonPanel SettingsPanel;
+        RibbonPanel FilePanel;
```
MackinnonBuck: By C# naming conventions, private fields are camel cased. Just something to keep in mind going forward.

exporters/BxDRobotExporter/BxDRobotExporter/BxDRobotExporter/Utilities.cs
```diff
@@ -13,7 +13,7 @@ namespace BxDRobotExporter
 {
     internal static class Utilities
     {
-        public const string SYTHESIS_PATH = @"C:\Program Files (x86)\Autodesk\Synthesis\Synthesis\Synthesis.exe";
+        public const string SYNTHESIS_PATH = @"C:\Program Files (x86)\Autodesk\Synthesis\Synthesis\Synthesis.exe";
```
MackinnonBuck: Constants should use PascalCasing according to C# naming conventions.

exporters/BxDRobotExporter/BxDRobotExporter/BxDRobotExporter/StandardAddInServer.cs
NWalker1208: Diff groups have different conventions, to avoid refactoring all of the code, we should probably stick to the conventions currently used in the code, which is pascal case for private fields, and all caps for constants.

PR #85, opened 2018-07-06
Improved Weight Setting Consistency

NWalker1208:
Resolves AARD-586:
- The weight controls in the wizard now more closely resemble those in the set weight form.
- Values for weight and units are synchronized between the wizard and the set weight form.
- Moved the back-end storage of weight from GUI to RMeta to make it more consistent with robot name.
- Chosen weight units are now saved to the assembly file along with the weight of the robot.


MackinnonBuck:
@NWalker1208 please resolve the merge conflicts when you can 😃

NWalker1208:
Fixed conflicts

PR #91, opened 2018-07-06
Feature movement arrows

MackinnonBuck:
I've added functionality that allows users to reset the robot by dragging arrows. This gives CAD users a familiar method of moving objects. The WSAD controls have been left in, however, in case some users find that to be easier.

Testing should involve setting the spawn points with many various robots using the arrows.


MackinnonBuck:
@BradenMccoy I'm going to merge this right now so others can work on the feature involving saving the spawn point. Feel free to review it later so I can fix any bugs that might pop up.

PR #94, opened 2018-07-06
Converted arrow meshes to .3ds files

MackinnonBuck:
This change prevents an error from showing up at the start letting the user know they don't have blender installed.


MackinnonBuck:
No functionality changed here so no testing is required.

PR #92, opened 2018-07-06
Exporter Wizard Flow Improvements

NWalker1208:
Resolves AARD-613:
- When no joints are left, the other moving parts page is skipped. To allow this to happen, some of the wizard flow management was altered.

Resolves AARD-614:
- When the wizard is completed, the existing configuration is erased before the wizard data is applied.

Resolves AARD-621:
- First page of wizard has been removed.

Other:
- Background of wizard was changed to white to match other dialogs.

PR #95, opened 2018-07-07
Exporter node name fix

JoshGrace:
Fixes
AARD-594
When deleting nodes. the wheel wizard should put the node back into the correct order in the node view
AARD-593
The wheel wizard should allow users to reorganize wheel nodes within the view
AARD-592
The wheel wizard should allow users to drag and drop wheel nodes between windows
Most of AARD-552
Remove or edit the useless menus that exist within the exporter (First and last guided export page, "saved" popup, etc.)
The red text in the exporter should refer to required items only, or walk the user through which items MUST be selected to continue.

Implements drag and drop between left and right wheel groups and reordering within the groups themselves. Also, this allows the user to place wheel nodes in whatever order they please


NWalker1208:
Coding format is much improved from last time I looked at it, good job Josh!
Also, some of my pull request https://github.com/Autodesk/synthesis/pull/92 has some overlap with AARD-552. We may want to merge that first, then check for conflicts.

JoshGrace:
The bug is outside of the scope of the pull request

NWalker1208:
Please merge https://github.com/Autodesk/synthesis/pull/92 before this request

PR #96, opened 2018-07-09
Identical part names error

NWalker1208:
Added protection against errors in models with multiple parts of the same name. Unfortunately, Inventor protects against this situation well enough already that I could not test these changes fully. I know it doesn't break anything, and it should resolve the issue. This resolves AARD-623.


HiceS:
I don't see any issues.

PR #98, opened 2018-07-09
Exporter non-joint components hider

JoshGrace:
Gives the wizard the ability to read through the model and find any components not associated with a joint and disables them for the duration of the exporting process.


JoshGrace:
also, this fixes AARD-633, an issue with crashes in models with more than 20 joints

exporters/BxDRobotExporter/BxDRobotExporter/BxDRobotExporter/StandardAddInServer.cs
```diff
@@ -306,6 +309,38 @@ private void StartExporter()
             
             EnvironmentEnabled = true;
 
+
+            jointRelatedOccurences = new ArrayList();
+            foreach (ComponentOccurrence c in ((AssemblyDocument)MainApplication.ActiveDocument).ComponentDefinition.Occurrences)
```
HiceS: It's best practice to make object lists item and store the data as a copy before going into a iterative loop while doing casting and function calls.

exporters/BxDRobotExporter/BxDRobotExporter/BxDRobotExporter/Wizard/Components/DefinePartPanel.cs
```diff
@@ -43,7 +43,7 @@ public DefinePartPanel(RigidNode_Base node)
             PortOneUpDown.Minimum = 3;
             PortTwoUpDown.Minimum = 3;
 
-            PortOneUpDown.Value = WizardData.Instance.NextFreePort;
+            PortOneUpDown.Value = WizardData.Instance.NextFreePort < 19 ? WizardData.Instance.NextFreePort : 19;
```
HiceS: This seems like you are saying as long as the port is less than 19 its the port, but if its over 19 then it will be 19 no matter what .
so
 
`50 < 19 ? 50 : 19` is **19** and `40 < 19 ? 40 : 19` is **19**

PR #112, opened 2018-07-11
Gamepiece property sets

rasberrypai:
Intake interactor looks for similarities in property sets between the defined gamepiece and the collision object rather than the keyword since all keywords are different because of Inventor.

JIRA: https://jira.autodesk.com/browse/AARD-481

### DIFF
--- a/engine/unity5/Assets/Scripts/DriverPractice/Interactor.cs
+++ b/engine/unity5/Assets/Scripts/DriverPractice/Interactor.cs
@@ -7,6 +7,7 @@ using BulletUnity.Debugging;
 using System;
 using Synthesis.BUExtensions;
 using BulletUnity;
+using Synthesis.Utils;
 
 namespace Synthesis.DriverPractice
 {
@@ -22,11 +23,12 @@ namespace Synthesis.DriverPractice
         public string[] collisionKeyword = new string[2];
 
         public List<GameObject> heldGamepieces = new List<GameObject>();
-
+        
         private void Awake()
         {
             if (GetComponent<BMultiCallbacks>() != null)
                 GetComponent<BMultiCallbacks>().AddCallback(this);
+            
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


engine/unity5/Assets/Scripts/DriverPractice/Interactor.cs
```diff
@@ -37,13 +39,11 @@ private void Awake()
...
         {
-            string gamepiece;
             for (int i = 0; i < collisionKeyword.Length; i++)
             {
-                if (collisionKeyword[i] != null)
+                if (collisionKeyword[i] != "NOT CONFIGURED" && collisionKeyword[i] != "NOT DEFINED" && collisionKeyword[i] != null && GameObject.Find(collisionKeyword[i]) != null)
                 {
-                    gamepiece = collisionKeyword[i];
-                    if (other.UserObject.ToString().Contains(gamepiece))
+                    if (Auxiliary.FindObject(Auxiliary.FindObject("Field"), collisionKeyword[i]).GetComponent<BRigidBody>().mass == ((BRigidBody)other.UserObject).mass && Auxiliary.FindObject(Auxiliary.FindObject("Field"), collisionKeyword[i]).GetComponent<BRigidBody>().friction == ((BRigidBody)other.UserObject).friction && Auxiliary.FindObject(Auxiliary.FindObject("Field"), collisionKeyword[i]).GetComponent<BRigidBody>().rollingFriction == ((BRigidBody)other.UserObject).rollingFriction)
```
MackinnonBuck: This is a lot to process every time a collision occurs. Especially since the game element's GameObject and components are being searched for when comparing each individual property. I think a more efficient version of this method would be to find the required components once and then check the properties using that established reference. Also, what if two different property sets have all the same friction and mass values but contain different field elements? Wouldn't this `if` statement pass? I think maybe a different approach might be to somehow attach the property set ID to the field elements on field generation. Then you could compare these IDs when the collision occurs. That would have a lot less overhead. Every GameObject has a "tag" property so you could probably set that property to the property set ID and check the tag when the collision occurs.

PR #119, opened 2018-07-13
Exporter gear ratios fix

JoshGrace:
this PR statisfies the exporter portion of AARD-625
"The user is able to define gear ratios for drivers in the exporter, but this information is not used in the backend functions of the robot exporter in order to transfer that information to the simulator."

I changed the Schema of the BXDJ file to include the gear ratio data required for gear ratios to be imported into the simulator THIS WILL CAUSE ANY ROBOT EXPORTED WITH THIS CODE TO NOT WORK IN PREVIOUS BUILDS OF THE ENGINE. This means that devs must build the simulator from source to be able to run robots exported until a new release build of synthesis can be made.

This PR also takes the gear ratio data and writes it to the JointDriver data type, and writes the gear ratios to the Inventor PropertySets so it can be saved from session to session. Finally, it writes the gear ratios to the XML file for use with the exporter.

The next steps would be to determine if there is anything else we want to include in the .BXDJ, and incorporate that into this PR so we only have to majorly modify the Schema once. 
On the simulator side, we need to implement the usage of the gear ratio into the simulator itself.


JoshGrace:
@MackinnonBuck the schema currently allows for version numbers to be included in the .BXDJ if I modified the new BXDJs to have a 3.0 version, and have the SimulatorAPI import the 2.0 with the old method and the 3.0 with the new method would that help to prevent compatibility issues?

PR #120, opened 2018-07-13
Dpm physics

MackinnonBuck:
Implemented real physics simulation for driver practice mode game elements when held by the robot. This means that the game pieces will exert forces on the robot when held, and can collide with other objects in the simulation. This means we can finally create a stack of totes like an actual robot would :D

Testing should involve reviewing all aspects of driver practice mode to make sure nothing got broken by these major physics changes. Here are a few things to keep in mind:

1. The objects are now attached to their parent robot node via bullet physics constraint. This means that if the user sets the release position to be somewhere way outside of the robot, the robot will behave as if there is something actually connecting the robot to that game piece (this is why some robots will tip if you do this).
2. Some field elements were exported with masses set too high. For example, the 2014 field's exercise balls were *far* too heavy. So when the Simbotics 2014 bot, for example, picks up an exercise ball, it essentially becomes immobilized, and moving the manipulator results in the base of the robot rotating around the ball, not vice versa.
3. There is still an aspect of DPM that isn't physically accurate: Robots picking up more than one game element. This was an exception where the game pieces shouldn't collide with each other.


MackinnonBuck:
Update: I just realized I didn't have anything handling robot reset. @BradenMccoy I'll fix this in the next commit.

MackinnonBuck:
@BradenMccoy Just updated the reset code. Please review when you can 😃

PR #127, opened 2018-07-18
Mesh Exporting Optimization

NWalker1208:
Makes steps towards resolving AARD-495:
- Exporting is now faster (By about 30%) and multithreaded
- Exporting without color now gives more comparable output to color exporting
- Filtering for small parts now works differently and is more efficient
- Meshes are now lower resolution to reduce impact on frame rate in the engine. Visual quality has not been hugely impacted.

More multithreading investigation is being done on mesh-optimization-experimental branch, but it is not currently usable. If someone takes on this project in the future, they should work off of that branch. Currently, robots exported on that branch do not behave properly in the engine. This pull request is stable.


HiceS:
im not going to worry about a squash, we have other important stuff to focus on

PR #128, opened 2018-07-18
slider joint fix

JoshGrace:
fixes AARD-669
"Slider joints do not appear in the wheel selector, this results in the next page in the wizard being skipped so the user cannot define the joint in exporter setup."

Wizard now counts all joints in the skeleton and compares them to the number of joints with assigned wheel sides, allowing us to properly determine if there are more joints that need a Driver


HiceS:
code looks good to me

PR #130, opened 2018-07-20
Implemented CER Management Utility

matthew17754:
This is implemented in its own Solution directory and has not yet been implemented into any components. This is simply the class library implementation of the Crash Reporter.

To run a test/sample crash report, one must perform the following steps:
1. Right-click on the _CER_Utility_ Project in Solution Explorer and click _Properties_.
2. Change output type from _Class Library_ to _Windows Application_ and then Save.
3. Join the #failures Slack Channel.
4. Replace `string urlWithAccessToken = "INSERT WEBHOOK LINK HERE";`
5. Click _Start_

This Pull Request should resolve AARD-543, though it still needs to be implemented into some Synthesis components.


HiceS:
niiiiiiiiice

PR #143, opened 2018-08-01
Updated NSIS Installer GUI & Assets

matthew17754:
Optional: Although there's no easy way to test this right now, if you want to test it, you'll need to download the NSIS Compiler and get the Synthesis installer assets from me. QA on this Pull Request is optional.


HiceS:
Personally saw this work and build and oversaw development, did not include binaries so a okay to go!

PR #167, opened 2018-08-07
Fusion Exporter UI Improvements and Sensors

NWalker1208:
- Added a save configuration button to the exporter so that the user can save without exporting
- Added an edit sensors button to each joint
- Create the sensors menu, allows user to add encoders to joints (AARD-778)
- Exports encoders according to the 4.0.0 BXDJ schema


NWalker1208:
Okay, I added the back-end needed to work with the new schema. No drive train selection in UI yet, but since the data is not used, it isn't really necessary at this point.

PR #171, opened 2018-08-08
Fusion Exporter Improvements

NWalker1208:
- Added support for rigid groups in robot models, which should fix many broken models as long as they are properly jointed
- Many bug fixes, JavaScript is now much more reliable
- Optimized writing of BXDA files
This needs to be merged AFTER https://github.com/Autodesk/synthesis/pull/167 as it is based off of that branch.
- Update: Added fix for new Schema, also changed the schema so the names "port1" and "port2" are now "Port1" and "Port2." This improves capitalization consistency. New DLL is also present in the engine.


NWalker1208:
Also added slightly improved color processing
Still doesn't work very well though.

NWalker1208:
Now ready for review

PR #177, opened 2018-08-10
fixed edge case with elevator type being set weird

JoshGrace:
fixed issue when elevator type is set to 255 for some reason


HiceS:
simple change reviewed by shawn

PR #179, opened 2018-08-11
Fix raycast wheels

MackinnonBuck:
Fixed the issue where robots exported sideways don't always drive properly.

Testing should involve ensuring that other robots have not been broken by this change, as well as verifying that all robots exported sideways drive properly.


MackinnonBuck:
Merging with master since we need to get things moving.

PR #190, opened 2018-08-14
Emulation Rework

copernicium:
Replace the old WPILib-based emulation solution with an FPGA-based solution. As integrated with the engine, supports PWM and CAN outputs and joystick, match info, and encoder inputs; though the emulator can support much more.


HiceS:
**This PR is now ready to merge into master with the approval of QA** 

@BradenMccoy

PR #218, opened 2018-08-20
Field Config Save

NWalker1208:
Added saving of field configuration to Inventor assembly. Resolves AARD-896.
Required small changes to SimulatorAPI, so DLL was rebuilt.
Uses a new C# JSON library.

Includes changes from https://github.com/Autodesk/synthesis/pull/215


NWalker1208:
I think a little more optimization needs to be done. There is currently a huge bottleneck when loading configs.

NWalker1208:
Issues have been fixed. Ready for review

PR #224, opened 2018-08-20
added software tracking to the BXDJ and added motor types to the schema

JoshGrace:
added motor types to the BXDJ and added what kind of Autodesk software was used to export the robot


HiceS:
@JoshGrace  FIX DIS

PR #243, opened 2018-08-24
Network integration

MackinnonBuck:
Finished integrating network multiplayer with the new UI.

Testing should involve making sure that everything still functions in the normal simulator. Network multiplayer mode itself is also ready to be started testing.


MackinnonBuck:
Merge conflicts have been resolved.

HiceS:
@BradenMccoy can you test this before testing anything else to make sure we are on track for the release.

MackinnonBuck:
That should be fixed now. It's a Unity bug that sometimes adds GameObject components to the prefabs themselves. We'll just have to make sure the NetworkElement prefab doesn't have any NetworkMesh components attached to it when we do the final build.

HiceS:
@BradenMccoy

BradenMccoy:
Multiplayer is currently LAN-only, the possibility of global multiplayer is being investigated.

BradenMccoy:
@HiceS Through testing we discovered that network multiplayer does not work (except for LAN), @MackinnonBuck is currently looking into something that would allow for global multiplayer.

MackinnonBuck:
Just fixed some merge conflicts, now doing some final testing to make sure nothing broke. Then I'll merge the PR.

PR #246, opened 2018-08-27
Engine code clean

rasberrypai:
Documentation, interactor fix (cant score when holding gamepiece, cant steal gamepiece from other robots), spawn fixes (AARD-883 and gamepiece clones wont automatically be intaked - use synthepark for quick testing)
JIRA: AARD-883


HiceS:
thats a heck of a lot of cleanup

PR #251, opened 2018-08-31
Added EmuInstaller Cross-Architecture Support

matthew17754:
Emulator should now install on both 32bit and 64bit OS and detect either Synthesis.

Though personally, I've never actually seen the emulator run on a 32 bit system.

I feel this is something we should test before our next release @BradenMccoy


matthew17754:
We've decided that we're no longer going to support emulation on x86 hardware

PR #252, opened 2018-09-05
Added check for WiFi connection

KiraCorbett:
Allows users to launch and use Synthesis without an internet connection. @matthew17754 Can verify if this implementation would work with other back-end code. A very similar code snippet from a forum states this may not be the most reliable solution. 

Tested in airplane mode and by changing the integer value within the script to verify the auto-updater is still called. 

JIRA Issue: [AARD-880](https://jira.autodesk.com/browse/AARD-880)

### DIFF
--- a/engine/unity5/Assets/Scripts/GUI/MainMenu.cs
+++ b/engine/unity5/Assets/Scripts/GUI/MainMenu.cs
@@ -66,25 +66,50 @@ namespace Synthesis.GUI
 
         private void Awake()
         {
-            WebClient client = new WebClient();
-            ServicePointManager.ServerCertificateValidationCallback = MyRemoteCertificateValidationCallback;
-            var json = new WebClient().DownloadString("https://raw.githubusercontent.com/Autodesk/synthesis/master/VersionManager.json");
-            VersionManager update = JsonConvert.DeserializeObject<VersionManager>(json);
-            updater = update.URL;
-
             string CurrentVersion = "4.2.0.1";
-            Auxiliary.FindObject(gameObject, "ReleaseNumber").GetComponent<Text>().text = "Version " + CurrentVersion;
 
-            var localVersion = new Version(CurrentVersion);
-            var globalVersion = new Version(update.Version);
-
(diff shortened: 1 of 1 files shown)


engine/unity5/Assets/Scripts/GUI/MainMenu.cs
```diff
@@ -66,25 +66,50 @@ public bool MyRemoteCertificateValidationCallback(System.Object sender, X509Cert
...
+            }
+            else
             {
-                Auxiliary.FindGameObject("UpdatePrompt").SetActive(true);
+                Auxiliary.FindObject(gameObject, "ReleaseNumber").GetComponent<Text>().text = "Version " + CurrentVersion;
+            }
+        }
 
-            }//client.DownloadFile(update.URL, @"C:\Users\t_moram\Downloads\Synthesis Installer.exe");
+        public static bool CheckConnection()
```
matthew17754: This probably shouldn't be `static` as that'd make the method become defined at compile time, something which isn't necessary here.

engine/unity5/Assets/Scripts/GUI/MainMenu.cs
```diff
@@ -66,25 +66,50 @@ public bool MyRemoteCertificateValidationCallback(System.Object sender, X509Cert
...
-
             string CurrentVersion = "4.2.0.1";
-            Auxiliary.FindObject(gameObject, "ReleaseNumber").GetComponent<Text>().text = "Version " + CurrentVersion;
 
-            var localVersion = new Version(CurrentVersion);
-            var globalVersion = new Version(update.Version);
+            if (CheckConnection())
+            {
+                WebClient client = new WebClient();
+                ServicePointManager.ServerCertificateValidationCallback = MyRemoteCertificateValidationCallback;
```
matthew17754: This Remote Certificate Validation Callback is what allows us to make checks on the server over Unity. So this will need to  be added to the `CheckConnection()` method as well in order for us to make that callback.

PR #257, opened 2018-09-24
Added Fusion Plugin To Installer

matthew17754:
Installer should now install the necessary plugins for Fusion.

Please let me know if the extractions look correct or of if anything needs changing @HiceS


HiceS:
looks good from what I can see

PR #287, opened 2019-06-18
Game piece controls disappearing from the controls menu

KyroVibe:
If you went into any field/any robot, then left and went into a different field the control listings for interaction with the game pieces for that specific field would be gone and can't be reset and changed. Basically someone added in code to remove the field specific controls but never had it update to the new controls for the new field. I just ran the function that set these controls when the control menu tried to grab the controls
Issue 939 : https://jira.autodesk.com/browse/AARD-939


matthew17754:
For future reference, be sure to squash your commits for small changes like this @KyroVibe

PR #295, opened 2019-06-24
928 custom binding fix

KyroVibe:
The gamepiece controls were never properly re-added in a new field. I had a very general fix for it which caused a large bug. I removed the original fix and replaced it with a more precise one that didn't cause the same errors.
Jira issue 928: https://jira.autodesk.com/browse/AARD-928


KiraCorbett:
@matthew17754 Do you want to test this just to make sure it is not a local issue with key registries?

KyroVibe:
I think it should finally work. PLEASE test extensively before approving. It works just how it should on my computer.

KyroVibe:
@matthew17754 @KiraCorbett

KiraCorbett:
@KyroVibe There was a bug in master so I updated your branch from master. Can you confirm on your end that your branch is behaving as expected? Engine team is testing.

KiraCorbett:
I think we should hold off on merging until we fix a separate Unity bug.

PR #299, opened 2019-06-25
Restructured File Paths For Cross Platform Support

matthew17754:
Changed the formatting of all file path references using the built in .NET `Path.DirectorySepatatorChar` rather than a standard backslash which is only recognized on Windows.


matthew17754:
Note: You will likely need to manually re-add the SimulatorAPI.dll for testing this branch until it is merged.

Also, the UserMessageManager path prompts will likely be removed later on, but I'd like to keep them there for now.

PR #303, opened 2019-06-28
Disabling Network Multiplayer

KiraCorbett:
Disabled a user's access to network multiplayer from the main menu and commented out network multiplayer references due to the Unity 2019 upgrade. These were commented out to allow for further development until we address network multiplayer in sprint 3.


KiraCorbett:
Switched back to monobheaviour so should be tested well.

PR #330, opened 2019-07-08
597 flat plain

KyroVibe:
Made a flat plane to spawn into. Someone else should disable the main menu so it loads into the scene "Scene" first. You can mess about in the blank plane or load a field and all is good. Also for first timers they default to Dozer.
Jira issue 597: https://jira.autodesk.com/browse/AARD-597


matthew17754:
We've decided to hold off on this until we have a meeting to further discuss the main menu. Closing for now.

PR #334, opened 2019-07-09
Add preview icons to Fusion Joint Editor

wangl5:
### Add preview icons to Fusion Joint Editor
- When the joint editor button is clicked, a snapshot is taken for each joint and saved in a temp directory
- The path to the temp dir is sent to the palette in the JSON string
- The js loads the appropriate icon for each joint

### Improved joint highlighting
- Joint highlight triggered on mouse hover for 0.5 seconds instead of clicking on the small link in the corner of the card
- In addition to moving the camera, the zoom of the camera is set to a constant
- When clicking on the joint setup button, the camera moves to show the whole robot and the highlighted objects are cleared

NewJointEditor

HighlightZoom

### Jira
- Issue 1028

| Code      | ❌   | Kira Corbett/Alex Carter |   1028|       |

### DIFF
--- a/exporters/FusionExporter/Source/AddIn/CustomHandlers.cpp
+++ b/exporters/FusionExporter/Source/AddIn/CustomHandlers.cpp
@@ -71,7 +71,7 @@ void ShowPaletteCommandExecuteHandler::notify(const Ptr<CommandEventArgs>& event
 void ReceiveFormDataHandler::notify(const Ptr<HTMLEventArgs>& eventArgs)
 {
 	if (eventArgs->action() == "highlight")
-		eui->highlightJoint(eventArgs->data(), true, 1);
+		eui->highlightJoint(eventArgs->data(), false, 1);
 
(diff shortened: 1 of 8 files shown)


wangl5:
> 1. When a user minimizes the _edit joints_ window, then attempts to re-open the window by clicking the edit joints button on the toolbar, is it possible to pop open the side window again? See image. Does it reload the entire joint editor? @victoryforphil Thoughts?

I didn't see a way to collapse and expand palettes through the API, although I could definitely be missing something. 
I could grey out the joint setup button so users don't click it again and expect something to happen?

> 2. We may want to reduce the hovering time slightly unless that is similar to what Fusion already incorporates.

Hover time reduced to 0.2 seconds.

KiraCorbett:
Don't worry about expanding/collapsing panels as I forgot about the API palettes. Looks good! We can merge after conflicts are resolved.

PR #341, opened 2019-07-10
Removed Main Menu

KyroVibe:
Yes the title of the branch is off but I did a lot of stuff in this branch.
I:
- Removed main menu from build settings
- Created a grid to be the default "field"
- Removed the help buttons
- Created a settings panel
- Made a new scene that handles the fatal errors
- Made changing the directory for robots and field possible in the sim
- Fixed a quality settings bug

Now this is fine for the Beta but before release we need to:
- Fix the FTC fields because they are below 0
- Re-implement the global updater

Jira Issues completed in this PR:
1046: https://jira.autodesk.com/browse/AARD-1046
1047: https://jira.autodesk.com/browse/AARD-1047
967: https://jira.autodesk.com/browse/AARD-967
597: https://jira.autodesk.com/browse/AARD-597

PR #348, opened 2019-07-12
Broken Link Engine

KiraCorbett:
Two things that needed to get added to the analytics:
- Broken tutorial link in the control panel is now fixed
- Tracking what type of robot a user changed to (Mix & Match or Exported)


matthew17754:
Due to the fact that this PR contains the commits from #341 I think it would be best if we add these changes when we redo that PR.

PR #351, opened 2019-07-15
Removed Main Menu 2

KyroVibe:
So after lots of commiting and merging the branch should be finally complete.
There is no main menu, a default grid in place of the field. Made settings and change robot/field directory available from inside the simulator. Analytics are implemented everywhere for the most part. Also the loading text cycles through dots to show active loading and not a frozen game
The Jira issue no longer exist as it was in the previous sprint


KiraCorbett:
@KyroVibe Can you test that the field/robot directories now open? Everything looks good other than that. Nice job re-implementing all this work!

KyroVibe:
Testing and working. We need to fixed the quality of the loading panel but that will come later as we should remake it to match the rest of the simulator

KyroVibe:
Changes finished @matthew17754

KyroVibe:
So can I merge the pull request or am I not allowed to do that? @KiraCorbett

PR #356, opened 2019-07-17
Created new setting TrackAnalytics and made brief survey with prompt

nicolasespinoza:
### Created new setting TrackAnalytics and made brief survey with prompt
- Gives insight into why people use Synthesis
- To be merged & integrated with Alex's analytics branch

analytics

### Jira
- Issue 1043

| Code      | ❌   | Alex Carter |   1043 |       |


wangl5:
Nice survey questions! However, would it make more sense to put the survey in the unity application rather than in the exporters? Also, you might want to make it clearer to users that the survey is optional, maybe with `skip survey` and `submit survey` buttons.

nicolasespinoza:
> Nice survey questions! However, would it make more sense to put the survey in the unity application rather than in the exporters? Also, you might want to make it clearer to users that the survey is optional, maybe with `skip survey` and `submit survey` buttons.

I thought about that, but from my experience on my team, a lot of people download the actual Synthesis application just to try it out and see what it's like. They didn't necessarily have an intent, or even a robot CAD file. By putting it in the exporter, it's established a team or user wants to use Synthesis for a specific purpose, and I think it would be super insightful to capture that through our analytics.

As for the button ideas, I like it! I'll do that now.

PR #369, opened 2019-07-19
Add DOF Color Key

nicolasespinoza:
### Add DOF Color Key
- Explains the meaning of each joint highlight
- Opens and closes automatically with the DOF Viewer

Capture

### Jira
- Issue 1075

| Code      | ❌   | Kira Corbett / Liam Wang / Mat Moradi |   1075 |       |


nicolasespinoza:
> Not sure why this needed to be added to it's own separate pane. DOF already exists on the Guide Form.

As far as the separate pane goes, I think you're right and I like your suggestion. So I've updated it to look like [the Fusion 360 one](https://github.com/Autodesk/synthesis/pull/361).

The idea was originally suggested by Shawn at our Biweekly, so I added it to both exporters. I think it makes sense because eventually you can't add _all_ the information to the guide, and you have to split it up to keep it concise. I think the previous exporter had a few issues of trying to pack too much information into one place, and we don't want our guide to turn into that.

I think we should get some more feedback before we dismiss it, but if not, I can close both PR's at your request.

victoryforphil:
Have to agree, if anything, remove the key from the guide. This looks like the best place for this key.

nicolasespinoza:
> Looks great! You might consider adding some extra text in the window above the color key for both Inventor and Fusion explaining what the DOF tool should be used for, maybe something like "View the status of joints on the robot".

Good idea. I'll do that right now.

wangl5:
@matthew17754 @nicolasespinoza This is caused by not disposing of the EmbeddedKeyPane in Utilities.cs::DisposeDockableWindows(); you are attempting to re-use the ID of the EmbeddedKeyPane on a new instance the next time the exporter is opened.

PR #371, opened 2019-07-22
Fixed tab highlighting and added custom impulses panel

KyroVibe:
I made all the tabs properly highlight and also added a panel to apply custom impulses to the robot to test its ability to recover from impacts
JIRA issues:
https://jira.autodesk.com/browse/AARD-1080
https://jira.autodesk.com/browse/AARD-916


KyroVibe:
@KiraCorbett all the panels do that its just that the point impulse panel is set to render of the loading panel. Fortunately its a very simple fix

KyroVibe:
Sometimes the cursor is there sometimes it isn't. It is a bug inside of unity that we can't fix.

matthew17754:
Perhaps make the text disappear?

KyroVibe:
Again, that's handled unity side

KyroVibe:
I limited the max value to 6 characters. Anything 4 or above isn't even useful.

PR #372, opened 2019-07-22
Fixed JSON exporting

nicolasespinoza:
### Fix JSON exporting
- Merge Liam's cleanup (#367) which removed DLL reference embedding from Costura, resolving Newtonsoft dependency issues
- Enable Alex's JSON exporting with the following change:
https://github.com/Autodesk/synthesis/blob/bcd8501a520495f0e9252dcdbcf8dd0fa6090626/exporters/BxDRobotExporter/BxDRobotExporter/ControlGUI/DepricatedStandaloneGUI.cs#L334-L337


nicolasespinoza:
A note for testers: to tell whether this PR worked for you is to navigate under `C:\Users\<username>\AppData\Roaming\Autodesk\Synthesis\Robots`, and then go to the directory of the robot you just exported. If you open the `skeleton` file, it should be formatted in JSON. However, unless Engine has implemented JSON support, checking the "Open in Synthesis" box will not actually import the robot yet.

PR #374, opened 2019-07-23
Fixed no dozer failsafe

KyroVibe:
I made it so in the case that Dozer isn't found then it will go through all the folders in the default robot directory and search for any robot. If no robot is found it just throws an error.


KyroVibe:
Ok finally added the change robot panel to the error screen in the case that it can't find the selected bot

KyroVibe:
Did you fix it because it works for me? @matthew17754

KiraCorbett:
Is it possible to update some UI things?

- Lighten the highlighting on the directory scroll panel?
- Convert the "back" button into an exit button?

I can take care of some of these too. Just let me know which ones so we don't work on the scene at the same time.

PR #381, opened 2019-07-25
Rework Control Profiles

copernicium:
Rework how we handle controls to be more maintainable by cutting out unnecessary complexity and duplicate code. It also now manages saved controls in Player Prefs using JSON. From the user perspective, the main fix is they can now control outputs on PWM addresses 0 and 1 directly without need for our custom drive base control handling. 

The old engine `unity4` and `engine_research` folders were also deleted.


copernicium:
Instead of adding control options for basic drive base type controls, I added matching control editing. Right clicking on a key mapping now selects all fields with that same key so you can edit them all at once. This will make it easier for users to modify controls while preserving the new joystick input to PWM output percent conversion. 

I also separated player and global controls in the backend, but UI changes will be needed to edit global controls more cleanly. Currently, they appear under all players, but changing them modifies those controls for everyone. 

Additionally, I fixed the Ctrl+H hot key to toggle hiding the main ribbon.

copernicium:
Add macro fields for editing the necessary key mappings for common FRC drive schemes into the control options UI.

copernicium:
Clicking on a control that's waiting to be re-mapped now returns the control to what it used to be. 

I added some simple instructions to the control settings panel in addition to improving the color scheme of the profile drop down. I know some UI changes are in the pipeline for the controls settings, so I tried not to change too much.

I also improved the free roam camera controls to be a little more intuitive and fixed a bug with the overview camera when using the simple plane.

PR #389, opened 2019-07-29
Fixed visibility of text on selectables

KyroVibe:
I made font color of selectables black so they are more visible
JIRA: https://jira.autodesk.com/browse/AARD-1119


KyroVibe:
Sorry I thought the selected text was what it was referring to

PR #396, opened 2019-07-30
Engine loading screen message

ambercmerrill:
Fixed the loading screen in the error scene to be consistent with the new coloring and added a message to all the loading screens per issue #1130 (https://jira.autodesk.com/browse/AARD-1130) so users are less surprised by long loading times.


matthew17754:
Looks good to me. Not sure if we want to keep the `.` at the end, as most user message statements don't have a period at the end.

PR #394, opened 2019-07-29
Joint editor should not take focus from main window

wangl5:
### Fix issue: Joint editor should not take focus from main window
- Merge #392 
- Show joint editor as a new window instead of a modal to allow users to move the robot view, and fix associated issues with the advanced joint editor

|    DOD    | Done |    Reviewer    | Issue | Notes |
|-----------|------|----------------|-------|-------|
| Code      | ❌   | Nicolas Espinoza/Alex Carter|  1123|       |
| Unit Test | N/A  | N/A            |       |       |
| Merged    | ❌  |                |       |       |

PR #401, opened 2019-08-02
Fix Synthesis opening regardless of "Open Synthesis" checkbox at export

nicolasespinoza:
### Fix Synthesis opening regardless of "Open Synthesis" checkbox at export

https://github.com/Autodesk/synthesis/blob/256e9d45becb12b8c741da3fd3278c8ecca62502/exporters/InventorRobotExporter/InventorRobotExporter/Utilities/Synthesis/SynthesisUtils.cs#L15-L18

### Jira
- Issue 1139

| Code      | ❌   | Liam Wang / Alex Carter / Kira Corbett |  1139 |       |

**This will merge #394**


wangl5:
@matthew17754 Fixed

WideFinish

wangl5:
@matthew17754 Good idea, I increased the font size.

I think that even if synthesis is opening, the user should know that the export was successful and the location of the output directory. Are you sure it isn't necessary?

nicolasespinoza:
> @matthew17754 Good idea, I increased the font size.
> 
> I think that even if synthesis is opening, the user should know that the export was successful and the location of the output directory. Are you sure it isn't necessary?

I think the reason he suggested it is because when you open Synthesis it covers the entire screen, so the user wouldn't even notice the dialog.

wangl5:
@nicolasespinoza Okay, now it will always display on top

PR #407, opened 2019-08-05
Color coordinated start and stop buttons on emulation toolbar

KyroVibe:
Made the start code button green and the stop code button red.
JIRA: https://jira.autodesk.com/browse/AARD-1137


copernicium:
Looks good to me.

PR #417, opened 2019-08-08
Fix Inventor analytics

wangl5:
### Current analytics events
- Each editor/subeditor logs a `pageview` event with the corresponding page title
- Events when the environment opens/closes
- Event when the exporter completes an export or is canceled
- Event when the exporter launches Synthesis

| Code      | ❌   | Nicolas Espinoza/Alex Carter|  1079|       |


wangl5:
@KiraCorbett Are there other events that the analytics should be logging?

nicolasespinoza:
> @KiraCorbett Are there other events that the analytics should be logging?

Yes, the custom fields for the analytics survey

wangl5:
@nicolasespinoza Should we make a separate PR adding the backend to the analytics survey because we also have to figure out when the form should show to the user?

wangl5:
> * Toggle Robot Export Guide button
> * Toggle Degrees of Freedom View button
> * Finish Robot Export button
> * The button that says "Yes" or "No" if a user decides to export their robot

Sounds good, I'll add those before we merge this PR.

> Everything looks great! Do these analytics also track that this is an Inventor CAD model?

In theory, the analytics hits are all including that the app is inventor, as well as the version and build number:
https://github.com/Autodesk/synthesis/pull/417/files#diff-f9971d17ac486b39e78c202217240a3eR45-R48
However, I wasn't able to find this info in the analytics dashboard. Anyone more familiar with google analytics?

nicolasespinoza:
> @nicolasespinoza Should we make a separate PR adding the backend to the analytics survey because we also have to figure out when the form should show to the user?

Already implemented 😉 

https://github.com/Autodesk/synthesis/blob/1256aea44b90efdb2d1e4c4df93cb33d43d26e28/exporters/InventorRobotExporter/InventorRobotExporter/Managers/RobotDataManager.cs#L164-L169

KiraCorbett:
How are the analytics being tracked? Most of the data should be under "Behavior" in the Google analytics.

wangl5:
@KiraCorbett 
For the analytics, I was looking at the `(Last 30 min)` tab of the `events` and `content` categories under `realtime` to watch the events come through.

wangl5:
@KiraCorbett 
> * Toggle Robot Export Guide button

Added (it's now in exporter settings). Also added one for changing the joint highlight color.
> * Toggle Degrees of Freedom View button

Tracked as the "Joint View" pageview
> * Finish Robot Export button

I think this is the same as "environment close" events
> * The button that says "Yes" or "No" if a user decides to export their robot

Isn't the value for `yes` equal to the number of successful exports? And `no` would be the number of environment closes. I could add separate events for these though if that would be more readable in the analytics console.

PR #418, opened 2019-08-08
Merge `fusion-devel` into master

wangl5:
### Merge `fusion-devel` into master

- Currently, no analytics have been added `(AARD-1056)`
- The following are included in the UI but are **not** written to the JSON file `(AARD-1163)`
    - Per-joint weight
    - Drivetrain weight
    - Gearing

PR #422, opened 2019-08-09
Fix `fusion-devel` merge conflicts with `master`

wangl5:
### Merge `master` into `fusion-devel` and resolve conflicts

@matthew17754 I needed to do this because `fusion-devel` is protected so I can't merge directly
Alternatively, we could make `fusion-devel` not protected temporarily...


matthew17754:
#418 has been unprotected

PR #430, opened 2019-08-12
Fix analytics by using client ID instead of user ID and update GUID generation

wangl5:
- Use client ID instead of user ID as hits with a user ID were not showing up in the analytics dashboard
- Use randomly generated user ID instead of the hash of the username

| Code      | ❌   |Kira Corbett|  1079|       |

### DIFF
--- a/exporters/InventorRobotExporter/InventorRobotExporter/Properties/Settings.Designer.cs
+++ b/exporters/InventorRobotExporter/InventorRobotExporter/Properties/Settings.Designer.cs
@@ -166,5 +166,17 @@ namespace InventorRobotExporter.Properties {
                 this["ExportCount"] = value;
             }
         }
+        
+        [global::System.Configuration.UserScopedSettingAttribute()]
+        [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
+        [global::System.Configuration.DefaultSettingValueAttribute("firstRun")]
+        public string AnalyticsID {
+            get {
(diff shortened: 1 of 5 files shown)


victoryforphil:
Wont this generate a new GUID on each run? Removing all user tracking? IIRC the user id was working fine in the dashboard.

KiraCorbett:
I think we have to generate a new GUID so that we are not tracking personal clients.

wangl5:
@victoryforphil The GUID is saved to the settings XML file so it isn't reset each run. The user id hash version worked but was only anonymous one way. If someone's username is known, it is possible to hash the username and see their corresponding analytics.

victoryforphil:
If the GUID is saved then that's a good compromise.

wangl5:
@victoryforphil 
> IIRC the user id was working fine in the dashboard.

The hits were showing up in the realtime dashboard view but they didn't show up in the `Audience` or `Behavior` tabs for me. Are you sure you saw the hits in either of those tabs?

KiraCorbett:
@wangl5 Is this okay to merge? I checked the dashboard this morning and saw some user hits on various pages in the Behavior > Site Content section as well as Audience > User Explorer.

wangl5:
> @wangl5 Is this okay to merge? I checked the dashboard this morning and saw some user hits on various pages in the Behavior > Site Content section as well as Audience > User Explorer.

@KiraCorbett Yes, this is ready to merge

PR #438, opened 2019-08-13
Disabled Errornote when there is no selected bot

KyroVibe:
If the registry key is blank it makes the error note disable as to not confuse the user that an error occurred.

### DIFF
--- a/engine/unity5/Assets/Scripts/States/MainState.cs
+++ b/engine/unity5/Assets/Scripts/States/MainState.cs
@@ -162,6 +162,13 @@ namespace Synthesis.States
             //If a replay has been selected, load the replay. Otherwise, load the field and robot.
             string selectedReplay = PlayerPrefs.GetString("simSelectedReplay");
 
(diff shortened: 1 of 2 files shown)


engine/unity5/Assets/Scripts/States/MainState.cs
```diff
@@ -162,6 +162,13 @@ public override void Start()
             //If a replay has been selected, load the replay. Otherwise, load the field and robot.
             string selectedReplay = PlayerPrefs.GetString("simSelectedReplay");
 
+            if (PlayerPrefs.GetString("simSelectedRobot", "").Equals(""))
+            {
+                AppModel.ErrorToMenu("ROBOT_SELECT|FIRST");
+                Debug.Log("ASKJHASDFKJHDSAKJHGDSAF");
```
matthew17754: Looks good, but what is this debug log for?

KyroVibe: Honestly can't even remember

PR #441, opened 2019-08-13
Fix License Compliance

matthew17754:
Updated copyright license for license compliance


HiceS:
Looks good let's get this in

PR #452, opened 2019-08-16
Fix fusion UI palette heights

wangl5:
- Fix fusion UI palette heights
- Fix bug where palette and resources folders wouldn't get copied on every debug session


wangl5:
Addin settings save directory changed to the AppData folder.

PR #463, opened 2019-08-22
Fix Freeze Bug During User Program Upload

copernicium:
- Better manage the lifetime of the robot print-out tasks
- Add checks and limits on user program types and sizes
- Fix bug which prevented camera and control freezing


KiraCorbett:
Emulation changes are working good. We may push one more analytics change after verifying it shows up in the analytics dashboard tomorrow morning.

PR #470, opened 2019-10-12
Sticky Hatches

KyroVibe:
You can label a goal as "Sticky" and it will make the gamepiece stick to the goal. Please test this with your own setup to see if it is compatible with existing field goals.
(I couldn't find the JIRA issue I thought existed)

### DIFF
--- a/engine/Assets/Scene.unity
+++ b/engine/Assets/Scene.unity
@@ -76182,7 +76182,8 @@ MonoBehaviour:
     m_HorizontalOverflow: 0
     m_VerticalOverflow: 1
     m_LineSpacing: 1
-  m_Text: This is a beta feature that may cause glitches and other adnormal behaviours
+  m_Text: This is a beta feature that is in testing and may cause glitches in the
+    simulator.
 --- !u!222 &1425890473
 CanvasRenderer:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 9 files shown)


KyroVibe:
@matthew17754

KiraCorbett:
The sticky addition is pretty neat! I was actually able to get a robot to lift a few inches off the ground through a chain effect by sticking the gamepiece on the object -> then picking up the gamepiece from the sticky point which caused the robot to "climb." 

A few things to double check:
- When a gamepiece is stuck to a different object, it doesn't seem to increment the scoreboard. 
- Do you have a consistent way of getting a gamepiece to stick? It seemed to work well in the 2019 field with a ball but I couldn't get a hatch panel to stick. If you have a few steps you recommend, let me know so I could re-test. The smaller gamepieces were kind of hard to deal with. 

For now, let's double check the scoreboard and since the sticky feature is an improvement we can probably merge it and add the task to improve smaller pieces (like the hatch later.)

KiraCorbett:
@KyroVibe I forgot to add this image. When I created a giant goal, where does the sticky action take place? On the edge of the defined goal? 
image

KyroVibe:
@KiraCorbett Yes, the goals are supposed to match the shape and position of the hook and loop tape so where ever the gamepiece hits, that is where it should stick to.

KiraCorbett:
Messaged @KyroVibe and we are thinking of doing more testing to make sure we get consistent results before merging into master.

KyroVibe:
@KiraCorbett I added a little warning message that will popup. Basically it will display whenever the sticky feature is enabled and then will go away semi-permanently when you click the "I Understand" button. It will, however, reset after you restart synthesis. If you would like it to go away permanently, I can change it.

PR #474, opened 2019-11-13
Master Bug with Global Controls

KiraCorbett:
There is currently a null reference to the global controls on master. To test this fix, attempt to change your global control configurations by the following:
1. Attempt to change your controls by clicking on the "Menu" button -> Global Controls 
2. On first time generation, no controls were produced which resulted in a null reference error. This fix should now allow a user to change their controls.


KyroVibe:
Looks good. I can see the controls now and can change them and they save just fine.

PR #478, opened 2019-12-26
Fixed change MaM button adding a new bot instead of changing

KyroVibe:
Just something that was probably overlooked before hand
JIRA: https://jira.autodesk.com/browse/AARD-886


KyroVibe:
@matthew17754 I fixed the order issue. Could you verify?

PR #479, opened 2019-12-31
Fix standalone field exporter dependency issues and export bug

liamjwang:
Updated NuGet dependencies to use the [new PackageReference format](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files) which installs the dependencies globally instead of in the project `packages` folders.
The old method of requiring the dependencies within projects was causing issues with the `Aardvark-Libraries` projects which were used in multiple solutions.

The standalone inventor field exporter and the inventor robot exporter should now both build by following the instructions in the respective README files without any additional changes.

~The functionality of the standalone field exporter was not tested.~


liamjwang:
The standalone inventor field exporter is now tested and working with the 2019 field.
I was able to add the walls/rockets/cargo ship as static colliders and the cargo balls as sphere colliders, export the field, and open the field in the unity app.
My field colors and all the colliders seemed to be working as expected.

In https://github.com/Autodesk/synthesis/pull/479/commits/dc8d3c7687dd4cf751da29cf2d71c4ba5f1ba802 I fixed a bug that was causing my export to fail. 
As I haven't previously worked on the field exporters I'm not sure if there are other bugs that might cause similar crashes, so feedback from others who have is welcome.

liamjwang:
@matthew17754 If visual studio is run as an administrator (necessary for copying the executable in the post-build task) it can't attach to the Inventor process which was not launched as administrator.
If you run visual studio normally and remove the post-build task the exporter is able to attach to Inventor. This might be good to note in the readme.

KiraCorbett:
I may have missed this but is there a fix for this error? 
image

PR #493, opened 2020-04-14
Synthesis 2020

KyroVibe:
This is a basic start for the new synthesis. Exporter will probably need to be completely remade along with the protobuf files but it's a start. Also if you could figure out how to make this not be 5k+ commits that would be wonderful.


KyroVibe:
Im going to leave this on its own branch until we can figure out what to do about the 5k+ commits and until it has a working build

PR #496, opened 2020-06-01
Integrating the Integration of Continuous Integration

matthew17754:
CI is here! 🎉 

This is somewhat experimental at the moment as most of this technology has only come about within the last few months. 

Not to be a downer, but this first test is probably gonna fail. Just so ya know. :/

Note: Hold off on this until `aardvark-2020` is merged, as this refers to paths within that directory.

### DIFF
--- a/.github/workflows/EngineBuild.yml
+++ b/.github/workflows/EngineBuild.yml
@@ -1,0 +1,48 @@
+name: Engine
+
+on:
+  pull_request: {}
+  push: { branches: [master] }
+
+env:
+  UNITY_EMAIL: ${{ secrets.UNITY_EMAIL }}
+  UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
+  UNITY_SERIAL: ${{ secrets.UNITY_SERIAL }}
+
+jobs:
+  buildForSomePlatforms:
+    name: Build for ${{ matrix.targetPlatform }} on version ${{ matrix.unityVersion }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        projectPath:
+          - engine/CoreEngine
+        unityVersion:
+          - 2019.3.14f1
+        targetPlatform:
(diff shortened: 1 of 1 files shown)


KiraCorbett:
This is excellent! Looking forward to testing it soon with `aardvark-2020`.

HiceS:
https://github.com/marketplace/actions/unity-builder

After looking at the action you are using why are we limiting the unity version?

.github/workflows/EngineBuild.yml
```diff
@@ -0,0 +1,48 @@
...
+  buildForSomePlatforms:
+    name: Build for ${{ matrix.targetPlatform }} on version ${{ matrix.unityVersion }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        projectPath:
+          - engine/CoreEngine
+        unityVersion:
+          - 2019.3.14f1
```
HiceS, line 22: This needs to be a environment variable in the repo that we can change as the modern unity gets updated. If you don't want to target the beta unity version this makes the most sense.

.github/workflows/EngineBuild.yml
```diff
@@ -0,0 +1,48 @@
...
+  push: { branches: [master] }
+
+env:
+  UNITY_EMAIL: ${{ secrets.UNITY_EMAIL }}
+  UNITY_PASSWORD: ${{ secrets.UNITY_PASSWORD }}
+  UNITY_SERIAL: ${{ secrets.UNITY_SERIAL }}
+
+jobs:
+  buildForSomePlatforms:
+    name: Build for ${{ matrix.targetPlatform }} on version ${{ matrix.unityVersion }}
```
HiceS, line 14: matrix.unityVersion artifact seems like what the build target should look at.

.github/workflows/EngineBuild.yml
matthew17754: Now updated to latest ✔️

PR #531, opened 2020-06-18
Add empty classes for Synthesis Core managers

copernicium:
Also begin working on the virtual file system.


copernicium:
You can test the features that have been added to the virtual file system by running the TestApi console project.

Issue #521, opened 2020-06-16
Engine - Saving Preferences


KyroVibe:
|    DOD    | Done |    Asignees    | Branch | PR |
|-----------|------|----------------|-------|-----|
| Status    | ❌ | @nicolasespinoza | 521-saving-prefs | <number> |

Create a central system for saving data such as keybinds and what not.

PR #538, opened 2020-06-23
Saving Preferences

KyroVibe:
I created a new manager called PreferenceManager. Each GUID has their own dictionary where they can store values with a given key. By using the Save function, all the preferences that have been stored will be saved to file currently placed in whatever directory FileSystem.BasePath is set to. By using the Load function, you can load that file into Synthesis using the AssetManager and then load all the stored preferences into PreferenceManager. You can use the TestApi to test the PreferenceManager. @NickBurrell and @copernicium I would like either of you to take a look at the modifications that I made to Directory and FileSystem.
Issue #521


copernicium:
This looks great to me, though there are a few things I'd like to talk to you and @NickBurrell about:
- When removing an IResource from the virtual file system, should we call its Delete method first? And for the types that are built into the Core (i.e. those that exist now), should those Delete methods be empty like they are currently?
- We should also talk about how we want to manage the preferences:
        - I don't think we should remove them from the virtual file system when we save them
        - Should they be saved simply into the "/modules" directory or somewhere else?
        - We should consider making and using a Resource to load and save the preferences instead of importing as a JSONAsset and manually saving to disk

PR #542, opened 2020-06-24
NUnit Packages added and converted Preference Manager tests to NUnit tests

KyroVibe:
Added the NUnit framework and NUnit test adapter. You can now run the unit tests inside of Visual Studios. You may need to install the NUnit test adapter extension for it to work inside your IDE. I'm unsure because I added that while trying to get this to work.
Issue #541


KyroVibe:
@copernicium do you want me to switch your tests as well or do you want to do that?

copernicium:
> @copernicium do you want me to switch your tests as well or do you want to do that?

@NickBurrell and I are in the middle of a Virtual File System / Asset Manager refactor, so we should hold off on doing that for now.

Issue #525, opened 2020-06-17
Engine - Event Dispatcher


KyroVibe:
|    DOD    | Done |    Asignees    | Branch | PR |
|-----------|------|----------------|-------|-----|
| Status    | ❌ | @annanixon | <branchName> | <number> |

Create a central class for handling all events in the engine.

PR #557, opened 2020-06-30
525 event bus

annanixon:
Added basic pub/sub functionality for Event Bus
#525


controller/Api/EventBus/IEvent.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Text;
+
+namespace SynthesisAPI.EventBus
+{
+    public interface IEvent 
+    {
+        public string EventType { get; }
+
+        public byte[] getData();
```
KyroVibe, line 10: I would get rid of this. Any data that would be passed in a IEvent type would just be in the form of properties that would be declared inside the newly defined IEvent type.

NickBurrell: Check my new changes I've pushed. I added general parametric support for arguments.

PR #561, opened 2020-07-01
OSX Installer Updates

### LINKED ISSUE #495: post-install.sh on MacOS shouldn't rely on homebrew
**Is your feature request related to a problem? Please describe.**
Most users don't have homebrew installed, so post-install.sh shouldn't rely on it.

**Describe the solution you'd like**
Use the tools that are part of unmodified MacOS installs.

**Describe alternatives you've considered**
1. Remove the "Sanity check for brew" block.
2. Replace `wget https://github.com/Autodesk/synthesis/releases/download/v4.3.3/SynthesisSampleAssets.zip -q --show-progress` with `curl -LO https://github.com/Autodesk/synthesis/releases/download/v4.3.3/SynthesisSampleAssets.zip`

rasberrypai:
Updated MacOS installer.

Installer now uses clickable bash scripts rather than drag and drop file folders. Both uninstall and install scripts are included in one dmg and the dmg also includes data files as well as the application. 

Closes #495


installer/OSX/README.md
```diff
@@ -1,62 +1,27 @@
...
-2. Put the now built Synthesis.dmg into the folder
-
-3. Copy post-install.sh into the new folder
-
-4. Copt instructions.txt into the new folder
-
-5. Zip that folder and publish
-
-
-## Build steps (Not Working) :
```
matthew17754, line 30: Let's move these steps to another directory

matthew17754:
Might also be a good idea to add a compressor to the `.dmg` package.

matthew17754:
Moving to `.pkg` installer instead

PR #567, opened 2020-07-06
Add module callback and system support

copernicium:
Rudimentary module loading now works.
Also update the Api.dll to be current.


modules/.gitignore
```diff
@@ -0,0 +1,2 @@
+*/bin
+*/obj
```
matthew17754, line 2: Consider adding more of the `.gitignore` essentials here, as it will override all global ignores.

copernicium: This doesn't seem to be an issue for me

PR #585, opened 2020-07-20
UI for API and Core

KyroVibe:
The entire UI system done and ready for use. You can use UXML files to create UI for panels and toolbars. Currently we have no uniformity over UI elements so whoever creates a UI element is responsible for maintaining this uniform style. I recommend making UI using the UIBuilder but you will have to modify the UXML to work with the VirtualFileSystem. The default UI has buttons and images for some default capabilities but are not setup inside the core. The engine module will have to bind these buttons separately.


api/Api/AssetManager/AssetManager.cs
```diff
@@ -360,6 +360,22 @@ private Inner()
                            throw new Exception("Import of image/sprite asset: wrong number of arguments");
                        return new SpriteAsset(name, perm, sourcePath);
                    });
+
+                RegisterAssetType("image/texture", new[] {".png", ".jpeg"},
```
copernicium, line 364: Both SpriteAsset and TextureAsset are registered to use these file extensions, so you will have to remove one or the other or there may be problems with how it automatically deduces which type of asset to import a file as.

I made SpriteAsset a while ago to be a part of the UI stuff, but it seems like it is not used. Should we delete it? Or do you think it might be useful for anything.

api/Api/AssetManager/SynVisualElementAsset.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public SynVisualElementAsset(string name, Permissions perms, string sourcePath)
+        {
+            Init(name, perms, sourcePath);
+        }
+
+        public SynVisualElement GetElement(string name)
+        {
+            if (_document == null)
+                Load(File.ReadAllBytes(FileSystem.BasePath + SourcePath));
```
copernicium, line 22: I don't think this is going to work because assets aren't really backed by files but by entries in a zip archive. So we'll need to find a different solution.

api/Api/Runtime/ApiProvider.cs
```diff
@@ -50,5 +52,16 @@ public static List<Component> GetComponents(uint entity)
 		}
 
 		public static Component? AddComponent(Type t, uint entity) => Instance?.AddComponent(t, entity);
+
+		public static T? CreateUnityType<T>(params object[] args) where T : class => Instance?.CreateUnityType<T>(args);
```
copernicium, line 56: We shouldn't expose any Unity code in the API so users won't have to install Unity to make modules.

api/Api/UIManager/UIManager.cs
```diff
@@ -1,12 +1,182 @@
...
+            newButton.clickable.clicked += () =>
+                EventBus.EventBus.Push("ui/select-tab", new SelectTabEvent(tab.Name));
+            ApiProvider.Log("Created new tab");
+            Instance.TabContainer.Add(newButton);
+            ApiProvider.Log("Added tab to top bar");
+            
+            RecursivePrint(Instance.TabContainer);
+        }
+
+        public static void RecursivePrint(VisualElement e, int depth = 0)
```
copernicium, line 37: We should probably rename this function to be more obvious what it does.

api/Api/UIManager/UIManager.cs
```diff
@@ -1,12 +1,182 @@
...
+            {
+                VisualElement elm = LoadedPanels[panelName].Ui.GetElement($"panel-{panelName}").VisualElement;
+                Instance.PanelContainer.Add(elm);
+                LoadedPanels[panelName].BindFunc(elm.GetSynVisualElement());
+            }
+            
+            // TODO: Maybe some event
+        }
+
+        private static void ClosePanel(string panelName)
```
copernicium, line 116: Should this be public?

api/Api/Runtime/IApiProvider.cs
```diff
@@ -19,5 +21,14 @@ public interface IApiProvider
 		Component GetComponent(Type t, uint entity);
 		TComponent GetComponent<TComponent>(uint entity) where TComponent : Component;
 		List<Component> GetComponents(uint entity);
+		T CreateUnityType<T>(params object[] args) where T : class;
+		VisualTreeAsset GetDefaultUIAsset(string assetName);
+
+        #region UI
+
+        TUnityType InstantiateFocusable<TUnityType>() where TUnityType : UnityEngine.UIElements.Focusable;
```
copernicium, line 29: Should this be a part of the API provider?

api/MockApi/ApiInstance.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using SynthesisAPI.EnvironmentManager;
+using UnityEngine.UIElements;
+
+#nullable enable
+
+namespace MockApi
+{
+    public class ApiInstance : IApiProvider
```
copernicium, line 15: This class should be deleted. We already have a mock API provider in `MockApi.cs`

engine/.gitignore
```diff
@@ -36,6 +36,10 @@ ExportedObj/
 *.pidb.meta
 *.pdb.meta
 
+# SynthesisAPI Library
+Assets/Packages/Api.dll
```
copernicium, line 40: The Api.dll needs to be tracked by git.

engine/Assets/Scripts/Testing/UIParseTest.cs
```diff
@@ -0,0 +1,98 @@
...
+using SynthesisAPI.UIManager.UIComponents;
+using SynthesisAPI.VirtualFileSystem;
+using Unity.UIElements.Runtime;
+using UnityEngine;
+using UnityEngine.Networking;
+using UnityEngine.UIElements;
+using Directory = SynthesisAPI.VirtualFileSystem.Directory;
+
+
+public class UIParseTest : MonoBehaviour
```
copernicium, line 25: Should probably make this a unit test

api/Api/AssetManager/AssetManager.cs
KyroVibe: Im gonna actually get rid of TextureAsset just because a sprite is a texture with a little more information and we can just grab the texture from a sprite.

api/Api/Runtime/ApiProvider.cs
KyroVibe: ApiProvider is going to be available to the modules?

KyroVibe:
@copernicium I either deleted or changed the things you requested. However I'm not sure what to do about the testing script. I think I might just remove it from the branch

api/Api/Runtime/ApiProvider.cs
copernicium: Yes, it's a part of the API

PR #589, opened 2020-07-23
Env manager refactor

rasberrypai:
Refactored Environment Manager and Input Manager

Environment Manager now loads entities and components into the Unity Scene. (Modules should use Environment Manager to spawn entities and components instead of ApiProvider). Input Manager's workflow is shifted toward modules (Separated from Preference Manager and inputs are generalized).


copernicium:
I updated it to master and verified that the API and CoreEngine unit tests all still pass and that everything still works.

PR #600, opened 2020-07-31
Configured API Package Details For NuGet

matthew17754:
Configured `csproj` parameters necessary for publishing a pre-release package to NuGet.
Note: Squash on merge due to unnecessary image file added to source control.


copernicium:
I assume this is all fine, but I'm not sure how to verify that it can be published to NuGet.

PR #605, opened 2020-08-05
Begin C, Java, and Python Controller projects

copernicium:
- Create C, Java, and Python projects to interface with the native Controller API (implemented in Rust)
- Update Controller module to use the JSON RPC error format
- Fix encountered bugs


.gitignore
```diff
@@ -1,7 +1,13 @@
 .vs/
 .DS_Store
 **/*/.idea/
-*.dll
-*.dll.meta
```
matthew17754, line 5: Any particular reason this was removed from the global?

copernicium: They're still in there. I just moved them down a bit with the other binary file types.

PR #607, opened 2020-08-07
Toast feed

### LINKED ISSUE #540: Engine/API - Add Adaptive Logging System
|    DOD    | Done |    Asignees    | Branch | PR |
|-----------|------|----------------|-------|-----|
| Status    | ❌ | <assignee> | <branchName> | <number> |

Create a way to create a popup text used for debugging and logging while in game (i.e. Preferences failed to load, Module doesn't exist)

copernicium:
Add a toast feed to the Synthesis engine. It is just another logger that is registered, so anyone can use it with calls to `Logger.Log("Test debug log", LogLevel.Debug)`, for example. To do so, I made it so we can register multiple loggers, and messages are sent to all of them.

Closes #540

⭐  Also as a part of this PR, I added the module metadata files to git, and I made many parts of the API that use Unity types internal so that modules cannot access them.

image


copernicium:
No they're not being used. I added a few extras that I was tinkering with. I can remove them, or we could keep them around and see if we come up with a use case for them later.

PR #622, opened 2020-08-15
Physics, Motors, etc.

KyroVibe:
Added support for rigidbodies, fixed joints, hinge/revolute joints, and added a MotorManager to SynthesisCore that adds a wrapper on top of the joint system to control them with specified motor settings.
[SynthesisCore.zip](https://github.com/Autodesk/synthesis/files/5079406/SynthesisCore.zip)


KyroVibe:
The module loader test is giving me trouble so I'm gonna use the CI to see if it will build.

engine/Assets/Tests/PlayMode/TestModuleLoader.cs
```diff
@@ -1,121 +1,121 @@
...
-using Engine.ModuleLoader;
-using System.IO;
-using System.IO.Compression;
-using System;
-using UnityEngine.SceneManagement;
-using SynthesisAPI.AssetManager;
-
-namespace Tests
-{
-    public class TestModuleLoader : IPrebuildSetup
```
copernicium, line 14: Add this test back and make it work.

KyroVibe: So I can't replicate the axis error. Could it be that your using the wrong module? Because I did get that error when testing and it was caused by an exception being thrown inside the Setup for sample system.

KyroVibe: Ok just re-added the test and for whatever reason the test finally determined that Api.dll was a thing.

KyroVibe: Also regarding the Axis error, did you add the Test.glb file to the module?

copernicium: I will take another look 👍

api/Api/EnvironmentManager/Components/Rigidbody.cs
```diff
@@ -0,0 +1,207 @@
...
+        ContinuousSpeculative
+    }
+
+    public enum ForceMode
+    {
+        Force = 0, Acceleration = 5,
+        Impulse = 1, VelocityChange = 2
+    }
+
+    public enum RigidbodyConstraints
```
copernicium, line 192: You may consider adding the `Flags` attribute to this enumeration which allows an enum value to contain many values. That way we could put multiple different constraints on things at once.

api/Api/EnvironmentManager/Components/Transform.cs
```diff
@@ -9,51 +11,51 @@ namespace SynthesisAPI.EnvironmentManager.Components
...
 		public delegate Vector3D PositionValidatorDelegate(Vector3D position);
 		public delegate Quaternion RotationValidatorDelegate(Quaternion rotation);
 		public delegate Vector3D ScaleValidatorDelegate(Vector3D scale);
 
 		public PositionValidatorDelegate PositionValidator = (Vector3D position) => position;
-		public RotationValidatorDelegate RotationValidator = (Quaternion rotation) => rotation;
+		public RotationValidatorDelegate RotationValidator = (Quaternion rotation) => rotation.Normalized;
 		public ScaleValidatorDelegate ScaleValidator = (Vector3D scale) => scale;
 
+		public event PropertyChangedEventHandler PropertyChanged;
```
copernicium, line 24: Can we make these private?

KyroVibe: internal at most

copernicium: Ok, that makes sense.

PR #624, opened 2020-08-17
Create mouse click event that is blocked by UI elements and UI bug fixes

copernicium:
- Create mouse click event that is blocked by UI elements
- Make Selectable and MeshCollider2D use the event bus
- Add a function to unassign digital inputs in the InputManager
- Cache created toolbars in their tab struct so we do not recreate them or register multiple button callbacks
- Cache panel elements and add unique id to button event tags


nicolasespinoza:
Looks good, I appreciate the unique IDs to button tags 👍

nicolasespinoza:
While I still do approve of these changes, I just realized while scrolling down a ListView (namely Environments and Entities windows), it does still zoom the camera.

This can be tackled later, but I thought it was something worth mentioning

copernicium:
Yeah, I see that too. We can probably fix that later by providing a UI-blocked axis for the scroll wheel in the InputManager that is zeroed out when the cursor is over a UI element. That shouldn't be too hard to do by mimicking how I fixed things in this PR.

Issue #633, opened 2020-08-22
Engine - Integrate Motor Configuration Functionality with UI


KiraCorbett:
|    DOD    | Done |    Asignees    | Branch | PR |
|-----------|------|----------------|-------|-----|
| Status    | ❌ | <assignee> | <branchName> | <number> |

To configure the joints in the engine, the motor type, motor gear, and motor count will need to be configured and connected to the UI.

Issue #632, opened 2020-08-22
Engine - Provide a Visual Highlighting Representation of Joints 


KiraCorbett:
|    DOD    | Done |    Asignees    | Branch | PR |
|-----------|------|----------------|-------|-----|
| Status    | ❌ | @KiraCorbett | joint-ui | <number> |

When a user clicks on the "Highlight" button, there should be a visual sphere that surrounds the highlighted joint. Currently, the camera redirects its position towards that specific joint but it is difficult to see this joint unless you are very zoomed in.

![image](https://user-images.githubusercontent.com/16855850/90949246-b15a4100-e3fb-11ea-9162-da679ad62a65.png)

PR #634, opened 2020-08-22
Joint Editor Window UI

KiraCorbett:
This is the base UI for joint configuration in the engine. 

This PR does include:
:heavy_check_mark: Joint Editor UI with functionality from @copernicium per #633 
:heavy_check_mark: "Highlighting" a current joint when the highlight button is selected. Added highlighting improvements per #632 
:heavy_check_mark: Custom dropdowns from @rasberrypai per #642 
:heavy_check_mark: Organization of joints per entity #635 

image


KiraCorbett:
Pending on some feedback for this API error. After an update from master and updating the dll many times, I am still not able to correct this.
image

copernicium:
> Scrolling or moving the window with dropdowns open causes the dropdown to disconnect from the window and remain open.

@rasberrypai Would you be able to work on fixing this bug?

matthew17754:
Any particular reason why right-clicking force-closes the form?

KiraCorbett:
> Any particular reason why right-clicking force-closes the form?

The right-click deselects the entity so that will naturally close the windows associated with that entity.

matthew17754:
Hitting the red **X** throws `Joint broken, cannot set`. Not sure if this is related to the Joint Editor but it is critical all the same.

copernicium:
> Hitting the red **X** throws `Joint broken, cannot set`. Not sure if this is related to the Joint Editor but it is critical all the same.

This is an issue with the Joint implementation. I think it should be fixed as a part of a separate PR, one that refactors how we handle joints.

matthew17754 (approved):

PR #654, opened 2020-09-07
Restored deleted Dialog assets and implementation

nicolasespinoza:
Makes Dialogs work again


KiraCorbett (approved):
Thanks for the quick fix. Works in my merged branch and can be tested using the sample implementation removed .

matthew17754 (approved):
Flex Grow

PR #655, opened 2020-09-07
Disabling Internal Collisions within designs

KyroVibe:
I added a collision layer field to mesh colliders. Also fixed some bad anchor parsing.


matthew17754 (approved):

KiraCorbett (approved):
Looks good. Do you want to keep the toast logger about how many triangles parsed for now or remove that?

KyroVibe:
@KiraCorbett You can remove it, it's only for debug purposes

PR #658, opened 2020-09-11
Implemented Splash Screen

matthew17754:
- Adds splash screen to engine startup (matches previous configuration)
- Adds icons to engine application executable
- Adds controller module extraction to shell (unrelated - test unnecessary)


KiraCorbett (approved):
Looks good and verified on a Windows build. Missed the Synthesis splash screen. Note for testers: I believe the Unity logo can only be removed with Pro builds.

KyroVibe (approved):
Tested great on windows. We'll just have to make sure however builds it disables the unity logo in the player settings

PR #659, opened 2020-09-12
Misc bug fixes

KyroVibe:
- Removed the error thrown by the joint adapters after the joints have been broken.
- Fixed error that was thrown if you tried to select a new entity after deleting one
- Removed triangle count toast


KiraCorbett (approved):
Looks good and glad to have the remove entity issue fixed. @KyroVibe Should we make an issue about dealing with broken joints for long term?

matthew17754 (changes requested):
The items in the entity toolbar UI don't gray out after the entity has been deleted. This should be done to indicate that the entity has been deselected.

KyroVibe:
@matthew17754 Well a week later and I'm finally done with the toolbar fix

matthew17754 (approved):
Looks good. Please resolve the merge conflict and I'll merge it.

PR #667, opened 2020-09-20
Updated Exporter Default Values

HiceS:
#### Reason 

When I saw the default values for the materials in the exported GLTF models I noticed lots of reflectivity indicating incorrect default mappings for Metallic properties in a default normal material so I updated the the material properties to reflect more realistic values.

#### Changes 

See Below for references

#### ScreenShots
**OLD DEFAULT**

old

**NEW DEFAULTS**

new

#### Notes
I also used the PR to investigate the exception that happens during the export but I could not find no good solution.

I tried isolating the section of the code throwing the runtime exception as well as attempted to force Pass Through exceptions but none of that worked and in general the exception happened after the current function had already been completed which indicates some strange malpractice to investigated later.

My entire IV Robot has an exception for each of the materials - Not sure TBD later

### DIFF
--- a/exporter/SynthesisFusionGltfExporter/gltf/utils/FusionUtils.py
+++ b/exporter/SynthesisFusionGltfExporter/gltf/utils/FusionUtils.py
@@ -68,12 +68,8 @@ def getPBRSettingsFromAppearance(fusionAppearance: adsk.core.Appearance, exportW
     metallicFactor = 0.1 # never 0 and certainly never 1
     roughnessFactor = 0.1
 
-    try:
-        roughnessProp = props.itemById("surface_roughness")
-        roughnessFactor = roughnessProp.value
-    except: # This is a pass-through (usually bad practice but it looks like this might be a fusion bug of some kind)
-        pass 
-
+    roughnessProp = props.itemById("surface_roughness")
+    roughnessFactor = roughnessProp.value
     baseColor = None
 
     modelItem = props.itemById("interior_model")
--- a/exporter/SynthesisFusionGltfExporter/gltf/utils/FusionUtils.py
+++ b/exporter/SynthesisFusionGltfExporter/gltf/utils/FusionUtils.py
@@ -65,9 +65,14 @@ def getPBRSettingsFromAppearance(fusionAppearance: adsk.core.Appearance, exportW
 
     transparent = False
     emissiveColorFactor = None
-    metallicFactor = 0.0
-    roughnessProp = props.itemById("surface_roughness")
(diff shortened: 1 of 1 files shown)


exporter/SynthesisFusionGltfExporter/gltf/utils/FusionUtils.py
```diff
@@ -65,23 +65,25 @@ def getPBRSettingsFromAppearance(fusionAppearance: adsk.core.Appearance, exportW
 
     transparent = False
     emissiveColorFactor = None
-    metallicFactor = 0.0
-    roughnessProp = props.itemById("surface_roughness")
-    roughnessFactor = roughnessProp.value if roughnessProp is not None else 0.1
+    metallicFactor = 0.1 # never 0 and certainly never 1
+    roughnessFactor = 0.1
```
HiceS, line 69: Added default values to the low end of metallic,

Roughness was modified as a side effect of testing for the exception, no actual change happened except for the null check.

exporter/SynthesisFusionGltfExporter/gltf/utils/FusionUtils.py
```diff
@@ -65,23 +65,25 @@ def getPBRSettingsFromAppearance(fusionAppearance: adsk.core.Appearance, exportW
...
 
+
     if matModelType == 0:  # Opaque
         baseColor = props.itemById("opaque_albedo").value
         if props.itemById("opaque_emission").value:
             emissiveColorFactor = fusionColorToRGBAArray(props.itemById("opaque_luminance_modifier").value)[:3]
-    elif matModelType == 1:  # Metal
-        metallicFactor = 1.0
+    elif matModelType == 1:  # Metal - even metal should be set at most to about .2-.5
+        metallicFactor = 0.3
```
HiceS, line 86: #### The Actual Change

Metallic property should really never be 1 unless this is a special type of mirror with unique shaders.

Defaulted to 0.3 because it looked good.

HiceS:
Another note is this was done with a custom renderer and not inside of Unity.

Worthwhile to test some more with real unity.

KiraCorbett:
fusionFixes-mesh
fusionFixes-synthesis

HiceS:
@KiraCorbett I already mentioned this as a separate issue, it has nothing to do with this pr but rather a bug in general. We already discussed this.

KiraCorbett:
@matthew17754 I can confirm that the bug I was dealing with is separate from Shawn's changes. Due to the bug, I can't test the new changes but the file changed were updates to specific values.

KiraCorbett (approved):

matthew17754 (approved):

PR #668, opened 2020-09-23
Implemented Native File Explorer Calls

### LINKED ISSUE #602: Engine - Open File Explorer Button
Create a button that will open the file explorer to the default `appdata` directory for user content. This will need to use the built-in .NET file path libraries to open the equivalent on all operating systems. We've decided to go with this solution instead of the overhead of a standalone file browser.

matthew17754:
Replaces the `Import...` button on all forms with an `Open Files` button which opens the operating system's native file browser directly to the specified %appdata% directory, a.k.a. `Environment.SpecialFolder.ApplicationData`. 

This simply passes a shell execution argument to the .NET `Diagnostics.Process.Start` library, specifying `open` as the verb parameter and allowing for cross-platform support. Closes #602 

UXML Forms Affected:
- Modules
- Entities
- Environments


KiraCorbett (approved):
Looks good. Tested on Mac and works well for Environments.

nicolasespinoza (approved):

PR #672, opened 2020-10-12
Fixed module assembly load order

KyroVibe:
A quick bug fix that loads all assemblies into the current domain then loads existing systems and what-not.


matthew17754 (approved):
Load order fix seems fine to me. @NickBurrell @copernicium can you confirm?

NickBurrell (approved):
Builds, and looks good to me.

PR #691, opened 2020-12-10
Restored Fusion Robot Exporter

matthew17754:
This restores the Fusion360 Robot Exporter for performing bug fixes in the winter sprint. 
Not sure if there are any dependencies needed from `Aardvark-Libraries` but inform me if there are.

In terms of the field exporter we'll need to discuss if we want to ship the standalone field exporter as-is or if we want to provide a more integrated solution within the existing Inventor/Fusion plugins, in which case I'll restore those as well.


NickBurrell (approved):
As far as I can tell, this restores all missing components. Approved.

matthew17754:
Closing due to deprecations in the build dependencies.

PR #692, opened 2020-12-13
Added Module API Readme

matthew17754:
Added basic user and build documentation for the module API. This probably won't be critical for another 9 months but it's good to make sure we have it anyway. If anyone wants to add further usage documentation for getting started with the API feel free. Note that this is subject to change once we ship the official release of the API.


NickBurrell (approved):
Documentation is a good starting point. In the future, we need to add an architectural overview, but for initial documentation this is good.

PR #717, opened 2021-06-24
Unity Resyncer Testing and Modification

victorxheng:
A script called SampleTest added testing capabilities to test UnityResyncer by creating a new thread and comparing it with the main thread. Running this script executes this test and prints it into the logs. 

UnityResyncer was also modified so that it would be able to work and move tasks into the main thread. The SampleTest script was run with the modified UnityResyncer and it was able to function properly.

AARD-1264:
[Test Unity Resyncer](https://jira.autodesk.com/browse/AARD-1264)


matthew17754 (changes requested):
Resolve Merge Conflicts

KyroVibe (changes requested):
Make you test scripts unit tests

PR #721, opened 2021-06-28
Added Tcp server to deserialize json objects

Aaron7317:
A TCP server that receives API Packets as JSON objects. The APIPacket and ModuleOutput classes were added as well. After receiving the Json Objects it generates a schema to check whether or not the Json is a valid APIPacket. If it is valid, the API packet name and list of ModuleOutputs are added to the dictionary in the JsonListener class.

AARD-1274
[JSON TCP Listener](https://jira.autodesk.com/browse/AARD-1274)


matthew17754 (approved):
Looks good. We'll also need to remember to link to the `UnityEngine.dll` once @KyroVibe moves the importer to its proper directory.

api/Api/Utilities/JsonListener.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Reflection;
+using System.IO;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Schema;
+using Newtonsoft.Json.Schema.Generation;
+
+namespace SynthesisAPI.Utilities
+{
+    public static class JsonListener
```
KyroVibe, line 15: So a few things with this class:
1) The JsonListener should be standalone. Instead of a Listen function, there should be some kind of Start function that will create the server, spawn a thread, and continuously listen. Then make sure you have the ability to forcibly interrupt the server whenever.
2) If the IDs are persistent and you are getting data more that once, using `Packets.Add()` is going to throw exceptions because it assumes that key doesn't exist. Use `Packets[key] = value`. This will work whether or not a kvp exists already.
3) If the packets you are receiving are longer than 256 bytes (which because its json it likely will be) currently your parser cuts the packet off at 256 bytes. Consider expanding the maximum sized byte buffer or add some logic to find the end of a given packet (I think you can look for a line termination character if you serialize the json correctly. If you use this approach I'd suggest just making your life easier and using a `StreamReader` which has a `ReadLine()` method just for this). This could also try to parse 2 packets written since the last read as a single packet.
4) Not sure why you are re-sending the packet back to the client. If the client tries to deserialize it, the client will likely have issues with all the variable names being altered by doing `.ToUpper()`

api/TestApi/TestJsonServer.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading;
+using NUnit.Framework;
+using SynthesisAPI.AssetManager;
+using SynthesisAPI.Utilities;
+
+namespace TestApi {
+    [TestFixture]
+    public static class TestJsonServer {
+        [Test]
+        public static void TestReceivingData() {
```
KyroVibe, line 14: Please use this unit test function I've set up for you to test your server and make sure it can receive and deserialize your json data.

KyroVibe (changes requested):

Aaron7317:
The TCP Server has been changed to accept multiple clients and run on a separate thread. It also deserializes the data asynchronously. It was also changed to use protobuf instead of JSON.

NickBurrell (approved):
LGTM, approved.

matthew17754:
Fix merge conflicts

NickBurrell (approved):
Fixed all merge conflicts

KyroVibe (approved):

KyroVibe (changes requested):
It looks like you accidentally undid the updates to sum of the nuget packages. Re update and restore the nuget packages in the solution and re commit. That should fix it

KyroVibe (approved):

PR #732, opened 2021-07-23
Proto api integration

Aaron7317:
Created ControllableState and RobotManager classes. ControllableState stores a signal layout and holds values for those signals. RobotManager stores a Dictionary of ControllableStates and has a ConcurrentQueue for updating them. I changed UpdateMessage to UpdateSignals and changed it to match the format of the mirabuf Signals. I also cleaned up the TcpServer a bit. It now doesn't need locks and its still thread safe (I think). It now has a SetTargetQueue method which sets the queue it will send the updates it recieves to. I made a couple of tests as well but commented out one of them because it uses the TcpServer.


api/Api/Proto/Assembly.cs
```diff
@@ -0,0 +1,3280 @@
+// <auto-generated>
```
KyroVibe, line 1: Add all the generated protobuf files to the gitignore

api/Api/Utilities/ControllableState.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Mirabuf.Signal;
+using Mirabuf;
+
+namespace SynthesisAPI.Utilities
+{
+    public class ControllableState
+    {
+        private Signals? _currentSignalLayout;
```
KyroVibe, line 12: If you could add some documentation to the controllable state that would be great. Just like a sentence or two of what it's supposed to do at the top

KyroVibe (changes requested):

Aaron7317:
Made requested changes and changed the Update method in ControllableState to make sure each Signal being updated is part of the current layout

NickBurrell (approved):

KyroVibe (approved):

NickBurrell:
Forgot to add a comment to my previous message, but everything looks good. The unit tests are still broken, for somewhat unknown reasons, but this is an issue that we can investigate at a later time (as we planned to).

matthew17754 (changes requested):
For some reason CI checks are stalling and there's still protobuf generated C# files present.

api/Api/Proto/Joint.cs
```diff
@@ -0,0 +1,2132 @@
+// <auto-generated>
+//     Generated by the protocol buffer compiler.  DO NOT EDIT!
+//     source: joint.proto
+// </auto-generated>
```
matthew17754, line 4: `Joint.cs` should be removed

matthew17754 (changes requested):

KyroVibe (approved):

KyroVibe:
I added an init script so instead of manually setting up the project, init will initialize the mirabuf submodule and generate all the proto files. Matt if you want you could try doing that for your CI

matthew17754 (approved):

PR #733, opened 2021-07-24
Transformational and Rotational Gizmo for objects

victorxheng:
The Gizmo is saved as a prefab in Assets > Prefabs > Game Objects > Gizmos. 
In Scripts > Gizmos, there are all the scripts associated with this. The GizmoManager static class manages gizmos and makes sure there is only one spawned in the scene at a time. To spawn a gizmo, you call the spawn function on this class along with defining the transform that you want to move.

The GizmoTester attached to the Tester object in the hierarchy allows you to test the gizmo on a cube. The temporary Gizmo button on the home bar will allow you to test it by activating the movable cube and spawning a Gizmo as a child.
image

This will spawn the cube:
image

Drag the axis arrows for axis movement.
Drag the planes for planar movement.
Drag the rotation arrows to rotate the object.
Drag the point marker for planar movement relative to the camera.

Press Control for snapping via axis arrows and rotational arrows. The snapping amount is configurable in the MoveArrow class. Currently, it is set to round to the nearest 0.5 units for axis arrows and nearest 15 degrees for rotations.

Press R to reset the rotation.
Press Escape to destroy the gizmo. The camera will then refocus on the original object and physics will be reactivated. 

Currently, bounds are set to 50 units in all directions (if the object goes beyond that point, it will be teleported back) and must be above y = 0. The maximum the object can move on one movement is 20 units.

Comments have been added for future reference.

To test the gizmo on any game object, just move the Gizmo object from the movable cube onto the transform you want to move. To do this by code, use the GizmoManager class. For the escape key to delete the gizmo, you must use the GizmoManager class.

You can now move around and drop dozers to your heart's content!
image


NickBurrell (approved):
Controls for the block function and handle exactly as expected, though the monochromatic UI makes it a bit difficult to look at at times. Additionally, in the future, it could be good to see rotations handled in a almost "gimble"-like layout, similar to how other Autodesk (specifically the animation suite) handles things. Aside from those comments, everything looks good. Approved.

KyroVibe (approved):
Fantastic Job

PR #734, opened 2021-07-26
Local Multiplayer Implementation

khuntetad:
Created a Local Multiplayer panel which allows the user to spawn in robots and also switch between different robot cameras. In the example below, you can see the camera at different positions when different multiplayer buttons are selected. In order to access the Local Multiplayer feature, hit the Multiplayer tab and then the controller icon.

Screenshot (5)
Screenshot (6)
AARD-1322
[Local Multiplayer](https://jira.autodesk.com/browse/AARD-1322)


victorxheng (approved):

NickBurrell (changes requested):
Functionality wise it works, however, I do have some issues with the UI panel itself. The buttons aren't labeled, aside from one, who's label is merely "label". Its a bit hard to discern which button does which thing. 

If you could pick icons for the buttons that are more suitable, and label the buttons on the panel appropriately, this would be best.

KyroVibe:
@khuntetad As an alternative to Nicks comment, could you just move the multiplayer button to the main tab with the other buttons we are using rn. Also add a label under the button saying "local multiplayer" or something. Btw merge conflicts now

KyroVibe (approved):

NickBurrell (approved):

PR #740, opened 2021-07-29
New User Interface Design for Toolbars

victorxheng:
Completely redid the navigation bar, rewriting the previous Bar script into one single NavigationBar script that manages tabs and panels with the LayoutManager. 

The new nav-bar is saved as a prefab that can be dragged in in case of merge conflicts in the scene. The tab switching uses instantiation and destroying tabs onto the tab transform.

image


KyroVibe (changes requested):
Everything is great, except you forgot to change the style of the dropdown in the game pieces tab

victorxheng:
Made the Dropdown look like the figma

KyroVibe (approved):
Looks great, we'll update the drop down again when the mock ups get updated

khuntetad (approved):
Looks good

PR #743, opened 2021-08-05
Populate Settings Panel with settings

victorxheng:
The settings panel now has a series of settings that the player can modify for their own preference. The look will be updated when the figma mockups come out.

image

Settings are either dropdowns or toggles. 

Three screen modes will be: full screen, maximized window, or windowed. There doesn't seem to be much of a difference between maximized window and full screen, I could consider only leaving full screen and windowed.

Currently, the player can only change the resolution when in windowed mode. Otherwise, the resolution is set to the native resolution. The resolution options change based on changing screen sizes, and if the player selected an option that is the native resolution, when screen resolution changes, it will automatically change to the max resolution for them. 

Currently, a bool called useImperial in the Preference static class is what is changed when the user swaps from imperial to metric. 

In Analytics, the PostDate() function is modified when the user turns off data gathering.


KyroVibe:
image
Changed a lot of the visual aspects of the panel. Also integrated it with the Layout Manager by making the SettingsPanel script actually inherit panel. Still have to check how it's spawned in and verify saving works. @victorxheng It looks like when you hit the reset button, it doesn't update any of the fields until after you exit the panel and re open it. Also you should add a way of detecting if there have been changes to any of the fields and to only make the Save button usable when changes are detected.

KyroVibe (changes requested):
Make it so the reset button will actually modify the current values in the panel (You can jank this by just closing and reopening the panel). Make it so the save button is disabled when nothing has changed. Make sure the button to open the settings panel uses the LayoutManager. Give your own pass on the styling of the UI (or use the one panel in the figma mock up to style it)

victorxheng:
Changes made

KyroVibe (changes requested):
Can you make it so the screen resolutions are available if windows mode is selected but not saved?

victorxheng:
image

New panel design matching figma.

victorxheng:
@KyroVibe looks good. Merge?

victorxheng:
Merged successfully. Added the API implementation script that was causing errors.

victorxheng:
Ready to merge

KyroVibe:
> 
> 
> Ready to merge

Please resolve the merge conflicts

matthew17754 (approved):

KyroVibe (approved):
Looks good

PR #746, opened 2021-08-12
Fix missing method implementation

NickBurrell:
Fix issue where there was a missing method for the `IApiProvider` interface on the `API` class.


matthew17754 (approved):
Resolves the issue on a fresh clone

KyroVibe (approved):

PR #756, opened 2021-08-27
1363 simulation manager server integration

Aaron7317:
The servers work with Synthesis running. They do use a bit of memory which might be something to address in the future


NickBurrell (approved):
Noted several minor issues upon review. Issues added to JIRA backlog, but functionality is present without bugs. Approved.

PR #760, opened 2022-01-17
Tcp improvements

Aaron7317:
I modified the TCP server to use sockets directly instead of the TCPListener class allowing for cleaner and more readable code. I also added functionality to release individual resources from a socket instead of having to disconnect completely. Finally, I added a couple of methods to change the tcp and udp ports.


matthew17754 (approved):

KyroVibe (commented):
Looks good. Unit tests pass and I updated it to use the newer mirabuf commits. I'm gonna hold off on approving until import-parsing-fixes goes through just because this branch is a lot smaller.

KyroVibe (approved):

PR #766, opened 2022-06-15
Added god mode

PepperLola:
The user can now hold <kbd>G</kbd> and left click and drag to move an object in the scene. There is a God Mode script on the Simulation Runner object that will add a configurable joint to an object that the user clicks on and updates the connected anchor position as they move the mouse. It recursively finds a parent object that has a non-kinematic rigidbody component to add the joint to. It can be a bit jittery, especially if the object is moved quickly, because it's setting the connected anchor location rather than using forces, so if there's a better way to move the object I'd love to hear it.

I also stopped ModuleLoader from being used when not on windows because Ionic doesn't work on mac, and commented out the ModuleLoader usages in Api so it would compile.

Jira Issue
[1372](https://jira.autodesk.com/browse/AARD-1372)


victorxheng:
Looks good, works great. Scroll no longer laggy, but slow at the moment, serialize the zoom sensitivity field? Able to move objects under the ground.

KyroVibe (approved):
Nice job on your first task

victorxheng (approved):
Looks good, comments made above, ready to merge

PR #767, opened 2022-06-16
1371 main menu

victorxheng:
Created new menu scene
-Added 2 new scenes: GridMenuScene is based on the figmo, LogoMenuScene was the one I originally made before the figma
-Temporary folder in scenes section contains some assets used for the LogoMenuScene, which can be removed later
-In the GridMenuScene, the figma is replicated and can be swapped out easily with new button images from figma
-In the LogoMenuScene, the camera orbits an autodesk logo and synthesis logo, and also, the robot configuration selection was drafted as well

image

Jira Task:
[1371](https://jira.autodesk.com/browse/AARD-1371)


KyroVibe (approved):
Looks good. @matthew17754 could you check and see whats up with the CI?

victorxheng:
Tweening looks good. Possible to make buttons pinch when mouse is held down instead of let go?

KyroVibe:
> Tweening looks good. Possible to make buttons pinch when mouse is held down instead of let go?

We can do that later when we review more UX

NickBurrell:
LGTM. Approved.

NickBurrell (approved):

PR #768, opened 2022-06-19
Replay Mode

KyroVibe:
This PR was taking too long so I got basic functionality wrapped up but a lot of the UI/UX is very temporary. We'll do another pass of this in the coming weeks.

By pressing Tab, the simulation will be paused and you'll be able to scrub through the last 5 seconds. When hitting tab again, you'll be able to unpause the simulation and it will continue from where ever you are at. Contact Markers are going to be disabled for the time being, they need to be optimized because for some reason there are like 4 collisions per actual collision, plus a ton of actual contact points per collision. Also all the integration with the rest of the UI isn't really there mainly because all the UI is gonna go bye bye.

[1376](https://jira.autodesk.com/browse/AARD-1376)


victorxheng:
image

Contact points in replay mode

victorxheng (approved):
Tested with replay functionality and collision marker points: works well and is able to scrub backwards in time effectively

PR #769, opened 2022-06-22
Gamepiece scoring

PepperLola:
I added scoring zones that have a gameobject, an alliance, a number of points, and whether or not to destroy gamepieces that enter them associated to them. I also created functions to create gamepieces from a single gameobject or a list of them and a draggable scoreboard panel that displays the current score. I added the functions to the FieldSimObject as well as the test class for the scoring zones because that's where I think it will go in the future, but those ones aren't used and can be removed if that's incorrect. There is also a property to change the visibility of the scoring zones (because they're visible but transparent by default), but it isn't connected to any preferences yet.


victorxheng (approved):
Code is well written and works. Functionality works and is clear in how to use the game piece system. Panel system and dragging functionality using the draggable panel script is effective. Well done!

KyroVibe (approved):
During the re-eval process and push for beta, we'll revisit this

PR #776, opened 2022-06-28
Practice Mode

PepperLola:
Added an escape menu with buttons to reset the robot, the gamepieces, the field, and all of them at the same time, as well as a dropdown to select a gamepiece and a button to spawn it in. There is also a button to configure the gamepiece spawnpoint using the gizmo. Created a popup menu type that's a copy of modal but that only has a cancel button that's centered horizontally, but it would be very easy to switch back to using a modal if this isn't preferred.


KyroVibe (changes requested):
Popups are literally just modals with extra modifications so could you just make a modal with said modifications. Also in the escape menu, the sizing of the dropdown is off from the rest of the buttons. Also when you split the content in the escape menu, you make the left content bigger than the right via the dimensions you pass into the function. Sorry, github isn't letting me highlight specific lines

victorxheng:
Mostly, the gizmo no longer works, might want to take a look at any changes that might have affected that. Might want to consider changing the look of the escape menu modal, maybe dividing into sections etc to make it less confusing / mixed up. Have load field link to the new field loading modal, maybe remove all game pieces and current field whenever swapping fields? 

Is the spawnpoint setting a y axis only thing? It seems that whenever I spawn in game pieces, it keeps its initial x and z axis but falls from above where it originally is (ie gamepiece from left side will fall down on the left side when spawned).

victorxheng (changes requested):
Above:
Mostly, the gizmo no longer works, might want to take a look at any changes that might have affected that. Might want to consider changing the look of the escape menu modal, maybe dividing into sections etc to make it less confusing / mixed up. Have load field link to the new field loading modal, maybe remove all game pieces and current field whenever swapping fields?

Is the spawnpoint setting a y axis only thing? It seems that whenever I spawn in game pieces, it keeps its initial x and z axis but falls from above where it originally is (ie gamepiece from left side will fall down on the left side when spawned).

Final thing, watch for Missing Reference Exception and other errors when sometimes repeatedly pressing the reset buttons - see whats being called and not having the reference.

victorxheng (changes requested):
Wow, amazing improvements! It works really well - the only question I have is if it is still supposed to spawn in primitive spheres for gamepiece_0; sometimes, at the start when I spawn in gamepiece_0, it spawns in spheres, but when I switch then switch back it spawns cubes.

Resetting seems to work. Some things we might want to reconsider (these are not requested changes): whether or not we show the modal at the start to spawn in both field and robot at the same time; and also whether or not an escape menu is the best idea for gamepiece spawning.

PepperLola:
Okay, I fixed the gamepieces defaulting to spheres. When I asked Hunter about choosing the field and the robot at the same time, he said that for now we can just choose the robot but that we might decide differently later. The user might want to test robot functions without a field just to see if it works (and if they don't want to take the time to wait for a field to load). I agree that the escape menu might not be the best option for gamepiece spawning (maybe we could have a button in the navbar like before?), but I think it's probably okay for now until we have the UI design more finalized.

victorxheng (approved):
Tested all functionalities and they seem to be working now, no explicit errors. Still don't think you need a reset game pieces button separate from reset field, but it works.

KyroVibe (approved):
Approved, a lot of integration to do with my branch

PR #777, opened 2022-07-18
Feature/thumbnail

HiceS:
# Exporter Changes

## Thumbnails

This change is basically to enable thumbnails to be taken when we are exporting Fusion Files.

Right now this is part of the [Mirabuf Assembly Thumbnail Field](https://www.mirabuf.dev/#mirabuf.Thumbnail)

I have set it to become Transparent since we can add any background we want to it later.

The changes to the total file size are present (~10kb)- right now 200px by 200px seems like a pretty large image.

We should really reduce that to about 60px

### Defaults

- Extension png
- Size: 200px x 200px
- [x] Anti_Aliasing
- [x] Transparent 

#### Example

Thumbnail_Shelfv5
Thumbnail_Shelfv5
Screen Shot 2022-07-18 at 2 19 04 AM

## README

Made a generic readme so that anyone new can run it without too much problems - plus we really should have one

## Deps

So on OSX we cannot run Subprocess with the `creationflags` argument

However, it seems like setting `buffer_size = 1` works just as well for some reason

Also I added a check to see if the pips downloader already exists so we don't have to constantly download it when something changes for development

## All other 

Apparently we have been forgetting to run the formatting command for python which is `black` in this case, see README for information on how to use it but it keeps everything in a consistent style.

# Image Size Comparison - Very Large Shelf File

## NO IMAGE
Screen Shot 2022-07-18 at 2 34 06 AM

## 60 px IMAGE
Screen Shot 2022-07-18 at 2 34 09 AM

## 200px IMAGE

Screen Shot 2022-07-18 at 2 34 11 AM


HiceS:
The color issue may have something to do with combining the anti-aliasing with the transparent feature, may have to move back up to 200px by 200px if that is the case.

KyroVibe (approved):
Still works for me. Didn't test the thumbnail because I don't want to add a png parser dependency atm and can't be bothered to write my own.

HiceS:
That's fine the thumbnail works I have already tested it, Im going to play around and see why it's giving me black models and then ill merge this when everything looks good.

HiceS:
Changed default size to 250px,

The creation flag error may represent an error if it used on OSX, need to be able to test this and verify it and fix it in another PR entirely.

response

PR #778, opened 2022-07-18
Importer Updates, Motor support, and more

KyroVibe:
### Tasks
- [x] Reviewed Importer and identified weak spots [JIRA](https://jira.autodesk.com/browse/AARD-1420)
- [x] Added support for motor data specified in Fusion [JIRA](https://jira.autodesk.com/browse/AARD-1444)
- [x] Modified how motors function and added inertia solving
- [x] Updated and tested importer to allow for dynamic fields with more convienent Gamepiece placement in Fusion
- [x] Fixed changing fields so it doesn't freak and throw an error
- [x] Added Support for objects that shouldn't have colliders [JIRA](https://jira.autodesk.com/browse/AARD-1440)
- [x] Fixed Gizmo tool
- [ ] Special wheel colliders

The wheel colliders are in the codebase but are not being used at the moment. They need to go through a lot more testing and I think keeping wheels the way they are is fine for the summer.

### Known Issues
The wheels, if set to a high enought speed, never seem to move the bot at the speed they should be. Likely caused by the wheels slipping as they still use default friction values. Could also be because of weak joints. Not of much concern for the time being.

### Tasks
- [x] Changed default values of joint motor speed and torque to be semi reasonable
- [x] Fix bug with motor references not being set
- [x] Changed joint motor speed to be a ValueInput and not an AngleValueInput to allow values more than 360 degrees

### Known Issues
Some selected meshes I used when testing were a little weird and don't render properly likely due to something wrong with them. When they enter the Engine is causes PhysX to silently complain about them and fail to generate a convex hull for them. This is likely a user issue they will have to mitigate.

Capture
It's much more noticeable when the camera is moving.

Transparency with the appearances aren't working. No clue why. The data fusion is giving has the same alpha channel value for every base color.
FusionTransp
Appearance working in Fusion

UnityTransp
Appearance is completely opaque in Unity

```
"Glass (Clear)_PrismMaterial-086_physmat_aspects:Prism-152": {
    "info": { ... },
    "albedo": {
        "R": 246,
        "G": 246,
        "B": 243,
        "A": 255
    },
    "roughness": 0.9,
    "metallic": 0.3,
    "specular": 0.5
}
```
Appearance as saved in the mirabuf file


NickBurrell (approved):
LGTM.

Found inconsequential bug, if you load in a field three times in a row (for 2018 field), limits are broken. Not relevant to this PR, but noted while testing

victorxheng (approved):
Tested Engine side changes. Dynamic fields work, gizmo is fixed, motor functionality works. Is functioning, besides the issues Nick discovered, irrelevant to this PR.

PR #779, opened 2022-07-20
Gizmo Improvements

KyroVibe:
# Engine
### Tasks
- [x] Update gizmo backend
- [x] Created intake configuration
- [x] Created ejection trajectory configuration
- [x] Added Victor's shooting stuff
- [x] Made colors update to colors.json
- [x] Updated new HUD for the Beta 

### Changes
Modified the gizmo back-end to be simpler to use.
You can now spawn in a gizmo without having the GameObject already.
You also now have more control over how the gizmo is utilized by specifying how it updates and starts off.

image
Added a test button at the top to move the robot after spawning it in.

image
Panel to configure intake trigger.

image
Panel to configure ejection trajectory.

image
image
Color changes made from the colors.json file in the AppData folder.

image
image
New HUD to be more minimalist. Mostly just want to add it to the beta to see if people like the vertical style as apposed to the horizontal style. The red circles are supposed to be the icons but i didn't like any of the icons we had so I kept them blank circles. The square is supposed to be the collapse/expand button and again, I don't have a good icon for that atm.

image
image
In trying to make the HUD very minimal, I moved most of the options and commands into option modals, where the user is presented a category of options to choose from. I think they are took big atm but we can fiddle with it later.

Fun fact: The Tab Drawer will actually hide options that aren't available, such as configuring your robot when there is no robot.

### Known Issues
The gizmo doesn't pause the simulation so it could cause potential issues.

Match mode is broken but I'm gonna have Victor fix it in another branch


NickBurrell (approved):
Everything works as expected. Approved.

Aaron7317 (approved):
No issues for me. Approved

PR #781, opened 2022-07-25
PhysicsManager Bug Fixes

NickBurrell:
Fixed several issues regarding freezing/unfreezing of objects related to the PhysicsManager. In particular, several bugs were fixed:

1. If you load a field, then a robot, grab a gamepiece, then a robot, the robot spawns underneath the field
2. If you load a field, then a robot, then a field, an exception is thrown
3. If you load a field, then a robot, grab a gamepiece, then a field, reproducing the above bug, then load a new robot, everything breaks.
4. If you load a robot, then a field, the robot spawns underground.

Each of these bugs have now been fixed.


Aaron7317 (approved):
Looks good to me.

PR #783, opened 2022-07-26
Camera modes

PepperLola:
Added freecam and overview cam, and a button in the new HUD menu to change between them (although it still uses the plus icon, and we might not want it to be there in the future). There is currently no driver station view because the fields aren't currently set up for that, and I don't know whether or not the driver station camera locations will be fixed or vary per field.

The camera mode automatically switches to orbit mode when the gizmo is instantiated and follows it by default. We might add a setting later to disable that functionality, but at the moment if the user *really* wants to be in another camera mode when they're moving the gizmo they can change it using the button.

The vertical position of the camera is clamped between 0 and 100 for freecam and overview mode so that the user can't go under the map and so far above it that they can't see it. 

The controls should be the same as in old Synthesis, except the arrow keys move the camera in freecam, because WASD move the robot (the opposite of old Synthesis). Holding down RMB is also not required to zoom in overview mode. Moving forward and backward also move in the direction the camera is pointing, including vertical movement, whereas the camera only moved along the horizontal axes (parallel to the ground) in old Synthesis.

Moving down in freecam is bound to shift, which conflicts with rotating a motor in reverse, which was also an issue in old Synthesis. Additionally, rotating yaw clockwise is bound to E, which conflicts with intaking a gamepiece, and moving up is bound to space, which conflicts with ejecting a gamepiece.

P.S. Sorry for the messed up commit history; I forgot to branch off of master, which is something I will keep in mind when making future branches.


NickBurrell (commented):
Everything appears to be working well so far! I need to do more testing, but is it possible to go back to the default "follow robot" camera? I don't believe I saw the option. There is also some keybind collisions in free cam. I'd recommend setting the default key for "raise camera" to shift, and lower camera to "left control", as these are fairly standard defaults. The other option is changing the throwing key to something like "G", "F", or "E", as those are also standards for "interact" buttons.

KyroVibe:
@NickBurrell Thats my bad actually. Keys like shift, control, alt, and menu are reserved as modifiers and can't be used as their own separate key at the moment. Maybe we should add a way to have an empty key with only a modifier attached to it?

PepperLola:
The orbit camera mode was meant to follow the robot, the focus point was just not getting set back to normal after the gizmo was exited. I set the raise camera keybind to shift and lower to left control as Nick suggested, although shift still conflicts with the default keybind to move an arm in reverse. The user will be able to work around this when the input panel is finished by rebinding the arm moving keybinds, and I'm not sure if there are any other binds for the camera that would still be intuitive. I worked around the modifier key issue by setting the keybinds to be left shift with the left shift modifier as well.

PepperLola:
I reversed the direction of the camera rotation in freecam when using the mouse because it was the opposite of most games and how the user would probably expect. I also froze the robot movement (in the ArcadeDriveBehaviour) when the user is in freecam so that they can use WASD to move the camera instead of the arrow keys. I removed `input/` from the freecam keybind names so that they won't show up in the input menu when that's merged as well.

KyroVibe (approved):
Looks good. Like the approach you took to adding in new camera modes. The grid gets a little wonky when you line up perfectly with the lines. Also gotta try something else out with the new way of disabling behaviours, but I'll tackle that with the controller integration.

NickBurrell (approved):
Looks good to me. Approved.

PR #785, opened 2022-07-29
Linux appimage

Aaron7317:
Synthesis can now be packaged as an AppImage. I added some documentation on the process of doing so. In my tests I was able to create and run an AppImage on an Ubuntu system with no apparent bugs. When it actually comes to packaging for release, we only need to provide the created AppImage file. The steps are pretty simple and listed on the docs. Using an up to date linux system, you make the init.sh script in the linux installer directory executable and run it, providing a path to your synthesis build as well as robots and fields. Then you run the appimagetool utility on the .AppDir directory and get the AppImage file.


matthew17754 (approved):
Works on Ubuntu

KyroVibe (approved):
Works on Ubuntu

PR #787, opened 2022-08-02
New Input UI

PepperLola:
Moved the keybinds UI over to the new dynamic UI system. Includes both robot-specific keybinds (such as movement) and global ones (such as god mode, ejecting, intaking, etc.) side by side using split content. When a robot isn't loaded, the robot inputs are replaced by a label saying that the robot isn't loaded.

Reassigning a keybind works the same as it did before: click the button for the keybind you would like to rebind, then press the key that you want to bind it to. It won't allow inputs to be bound to mouse buttons (nor did the previous input panel) because it's difficult to differentiate between the user clicking the button and them pressing Mouse0 to bind an input.

For global keybinds, it takes all assigned inputs from InputManager.MappedValueInputs and adds ones that start with `input/`. It replaces all non-alphanumeric characters with spaces and uses a built-in title case method to make the names more presentable. I'm not sure if this is the best way to do it but it certainly works for the modal as it is right now. It also excludes any inputs that are bound to a mouse button by default, like god_mode_drag, which is bound to Mouse0.

Modal when no robot is loaded:
No robot loaded

Modal when a robot is loaded:
Robot loaded


KyroVibe (approved):
Looks good. Not too sure about the background of the scrollviews yet but definitely good for now.

NickBurrell (approved):

PR #788, opened 2022-08-09
Settings Panel For Window, Quality, Camera, and Preferences

victorxheng:
This pull request is the creation of the new Settings Modal that allows the user to edit and save certain preferences:
- Screen Mode: Either Full Screen or Windowed
- Quality Settings: From Very Low to Very High
- Zoom Sensitivity
- Yaw Sensitivity
- Pitch Sensitivity
- Reporting Analytics
- Metric Measurements

The significant bulk of this pull request was dealing with resolutions and whether or not to include custom resolution features. After extensive work, we will not be allowing adjustable resolutions for now but may add that back in if necessary. For now, when the user full screens, it maximizes the resolution on the specific monitor's screen size. If the user sets the screen setting back to windowed, the window becomes windowed with the same maximum resolution by default (but can be adjusted manually by the user if they would like different window sizes).

image


PepperLola:
I don't know if this is because of an issue on my end, but the camera controller doesn't seem to be set up to work properly with the different camera modes. It might be worth merging master into this branch and resolving the conflicts if you haven't already?

KyroVibe (changes requested):
Ye the camera controller is still broken, and there is an error when opening the settings modal

NickBurrell (approved):
Approved with changes made.

KyroVibe (approved):
Everything looks good. Dunno how the weird scene thing go fixed but Im fine with it

PR #789, opened 2022-08-09
Improvements to Practice Mode

PepperLola:
Changes
- Filter gamepieces by the name of their children, because those are the same for each type of gamepiece
- Store gameobject data in a way that won't break when one is deleted
- Re-add resetting of robot and make it drop the currently held gamepiece when reset


KyroVibe (changes requested):
1. The label for all the resetting buttons says something about gamepieces, could you switch it to "Reset" or something
2. Reset all does the same thing as reset field. Make it reset the bot as well

victorxheng (approved):
Tested each item, each one works well.
1. Filtering game objects by name works - duplicate items are no longer there.
2. Game objects are stored properly and doesn't break when you delete one. However, it does throw a missing reference exception when you delete all game pieces, though that likely won't happen in real use. 
3. Resetting works

KyroVibe (approved):
Fixed a weird bug on failing to reset sim. Everything works and looks good tho

PR #790, opened 2022-08-13
Linux md5

Aaron7317:
App Images are now created by running one script "package.sh" which also goes and installs appimagetools automatically. It also creates an md5 file for the generated AppImage so that the end user can verify the integrity of the file. The documentation has been updated for both automatic and manual packaging.


Aaron7317:
Tested on Ubuntu VM

KyroVibe (approved):
Everything works. @NickBurrell Could you check and make sure the README instructions that I added are all good?

NickBurrell (approved):

PR #793, opened 2022-08-15
In-Synthesis Downloader

PepperLola:
Added a modal accessible from the main HUD that is split into two sections for robots and fields. It fetches all of the robot and field files accessible from the Synthesis website using the manifest, and adds labels for each item and a button to download it. Downloading a file uses the same path as loading either of the types of files, so it should be consistent between operating systems.

With some items to download
Screen Shot 2022-08-09 at 2 54 59 PM

All are downloaded
Screen Shot 2022-08-09 at 2 51 18 PM


NickBurrell (approved):
Worked perfectly! Approved.

KyroVibe (approved):
Looks great. Just had to make the alreadyDownloaded checker ignore directory and only check names

PR #794, opened 2022-08-17
Replay Reintegration

NickBurrell:
Reintegrate Replay Mode into Synthesis, excluding several features which are currently incompatible with how we handle gamepieces and shooting.


KyroVibe (approved):
Fixed like 4 weird bugs but all looks good. Question: The toolbar is still usable from replay mode. Surprisingly, you can pause, add a robot, scroll through the replay, and resume and nothing breaks. Should we keep it or disable the toolbar during replay? @NickBurrell

MattMoradi (approved):
Works as intended using the default `tab` key configuration. In the future we'll probably want to add the collision trajectory lines and force calculations.

PR #795, opened 2022-08-18
Toastlogger Touchup

victorxheng:
For the Toastlogger, messages are now all one line and can be interacted with as buttons, opening up a modal that displays the message on a scroll view. There is also a write-to-file feature that writes the message to a file and opens this file up in explorer for the viewer to view.


NickBurrell (approved):

KyroVibe (approved):
Gucci

PR #797, opened 2022-08-22
Google Analytics Update

NickBurrell:
Update to old Analytics Manager to Google Analytics v4. Currently, more events need to be added, but functionality is there.

In order to make this work, an external server is needed. Due to Google dropping support for Universal Analytics, it is, as of July 1 2023, no longer possible to use their traditional REST API to log new analytics events. Their new system is far more streamlined for mobile and web development. 

In order to make analytics work, we have a small server, `analytics_server.mjs`, which simply needs to be run on a fixed, static IP. Then, an amendment needs to be made to this pull request to change the IP address Unity looks for. Finally, we need a new analytics ID so the data gets pushed to the correct analytics stack. In its current state, analytics go to a temporary backend.


KyroVibe (approved):
Looks good. Updated the tracking ID to the correct one and added it to the settings panel

MattMoradi (approved):

PR #799, opened 2022-08-25
Changed default export path

KyroVibe:
## Exporter
I removed the file browser and defaulted the location to where the files are exported to.

For Windows, it exports to `%AppData%\Autodesk\Synthesis\Mira[\Fields]`
For Mac, it exports to `~/Applications/Autodesk/Synthesis/Mira[/Fields]`

### Issues
I have no idea if this works for Mac, or if that's actually the correct location (Though I'm pretty sure it is).


NickBurrell (approved):
Tested on Windows. Works correctly, approved.

MattMoradi (approved):

PR #835, opened 2023-06-22
Fixed swerve turning backwards

a-crowell:
### Description
The swerve robot would either have correct translation or correct rotation at seemingly random upon spawning. This code ensures forward/right is positive for each forward, strafe, and turn to get the three to all be correct/incorrect together. To make sure they are always correct, it also resets left and right wheels upon swerve configuration.

### Objectives
- [x] Fix bug that made swerve robots drive and turn in the opposite direction

### Testing Done
- Drove (both translation and rotation) swerve robots on different runs, with multiple robots, and respawning after removal

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1503)


KyroVibe (approved):
I made a quick addition to remove a newline. Other than that, looks good

PR #837, opened 2023-06-22
Removed Aardvark and Binary Tracking

HiceS:
### Description

This change removes the Aardvark project and the Binary tracking from Github. 

The new `init` scripts will replace the tracked binaries now that installing the dependencies is easier. 

There is a new script in `Scripts/osx/resolve_deps.sh` that will use brew to install `dotnet` as well as `protobuf` with the correct versions.
 
### Objectives
- [x] Make Dep install automatic on osx
- [x] Remove the tracked binaries
- [x] Remove the entire Aardvark Project

### Testing Done
- Compiled and Opened in unity
- Loaded a robot and tested driving around

### TODO
- Make automated installation scripts for Windows
- Update README with new info on how to install and setup project


HiceS:
Don't Merge at the moment. Need to fix something with the resolve deps scripts.

BrandonPacewic:
@HiceS Would you like me to take a look at making the windows install dependences script? If so would you rather it be written in bash or powershell?

PR #840, opened 2023-06-23
Windows resolve dependencies script

BrandonPacewic:
### Description

Adds a windows `.bat` file that will install `.NET 7` and `protobuf v23.3` and link them to the system PATH. 

Because this script edits the system PATH it requires elevated permissions. 

### Objectives
- [x] Create a windows `.bat` script that installs both `.NET 7` and `protobuf v23.3`.

### Testing Done
- Used this script to setup a new windows machine and ran the project in Unity.

### Potential Bug
- When running the project on the new windows machine I set up the ability to select a field and robot were broken. The dropdown for the selection appears empty possibly indicating that those files are missing. Unclear if this is related or not.


KyroVibe (changes requested):
Please add documentation in the [root README.md](/README.md) indicating how to setup and add a README.md to [this folder](/scripts/win/) giving resolve_deps.bat a description and usage information.

HiceS:
> 

Yeah I'm thinking about doing something like a animated gif and then linking it to the README to make it easier. We should look into if we can record a gif or video of the terminal in some super nice way. Maybe even one of those programs that recreates it from a pipe out.

PR #848, opened 2023-06-24
Fake Arm Control

KyroVibe:
### Description
Implemented a faked arm control for arms with limits

### Objectives
- [x] Fake control for limited rotational joints

### Additional Tasks
- [x] Fixed no wheeled robots error
- [x] Adjusted the limits of joints driven in fusion

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1530)


BrandonPacewic (commented):
Can someone confirm that the `No wheeled robots error` is fixed? I can't remember how to reproduce it.

Markytt:
> ~Can someone confirm that the `No wheeled robots error` is fixed? I can't remember how to reproduce it.~
> 
> We need a bot with no wheels just to confirm, very small fix I'm sure it's fine.

I have tested this, and I can confirm that the 'no wheeled robot error' has been fixed.

engine/Assets/Scripts/Importer/Importer.cs
```diff
@@ -257,9 +259,16 @@ public static (GameObject MainObject, MirabufLive miraLive, SimObject Sim)
...
+                            var min = -(limits.Upper - currentPosition);
+                            while (min > 0) min -= 2 * Mathf.PI;
+                            var max = -(limits.Lower - currentPosition);
+                            while (max < 0) max += 2 * Mathf.PI;
                             revoluteA.useLimits = true;
-                            revoluteA.limits    = new JointLimits() { min = -limits.Upper * Mathf.Rad2Deg,
-                                   max                                    = -limits.Lower * Mathf.Rad2Deg };
+                            revoluteA.limits    = new JointLimits() { min = min * Mathf.Rad2Deg, max = max * Mathf.Rad2Deg };
+                            revoluteA.extendedLimits = true;
                         }
```
BrandonPacewic, line 272:
```suggestion
                        if (limits != null && limits.Lower != limits.Upper) {
                            var currentPosition = definition.Rotational.RotationalFreedom.Value;
                            currentPosition = (currentPosition + Mathf.PI) % (2 * Mathf.PI) - Mathf.PI;

                            var min = -(limits.Upper - currentPosition);
                            min = (min % (2 * Mathf.PI) + 2 * Mathf.PI) % (2 * Mathf.PI) - 2 * Mathf.PI;

                            var max = -(limits.Lower - currentPosition);
                            max = (max % (2 * Mathf.PI) + 2 * Mathf.PI) % (2 * Mathf.PI);

                            revoluteA.useLimits = true;
                            revoluteA.limits    = new JointLimits() { min = min * Mathf.Rad2Deg, max = max * Mathf.Rad2Deg };
                            revoluteA.extendedLimits = true;
                        }
```

I'm pretty sure that this will create the same output but someone should double check me on that.

engine/Assets/Scripts/Drivers/RotationalDriver.cs
```diff
@@ -84,23 +84,23 @@ public enum RotationalControlMode {
...
+                SimulationPreferences.SetRobotJointMotor((_simObject as RobotSimObject)!.MiraGUID, Name, _motor);
             }
         }
         private HingeJoint _jointA;
         public HingeJoint JointA => _jointA;
         private HingeJoint _jointB;
         public HingeJoint JointB => _jointB;
 
-        private Rigidbody _rbA = null;
+        private Rigidbody? _rbA;
```
BrandonPacewic, line 95: Is there the possibility that this gets assigned a null value later on?

BrandonPacewic (changes requested):

engine/Assets/Scripts/Importer/Importer.cs
a-crowell: Don't see how this works for when currentPos < -pi. I think min and max work, but wouldn't min = (min % (2 * Mathf.PI) - 2 * Mathf.PI) % (2 * Mathf.PI) be more efficient? Good idea avoiding loops.

BrandonPacewic: That would not put `min` in the range of `[-2PI, 0]`, you can check this by running python code against a [Text Field Calculator](https://samfundev.github.io/TextFieldCalculator/Calculator.html). Although now that I think about it is that even the goal here? I looked at `currentPosition` and assumed the range that was wanted by `max` and `min`, however I'm not sure what the goal is now that I look at it again.

BrandonPacewic: Given that `%` creates a value between `[-k, k]` instead of `[0, k]` (like how it should). You are correct and we need to change this. I just updated my suggestion. Make sure it actually compiles before you merge into `dev`. Credit goes to @a-crowell for the one-liner for bounding `currentPosition`.

engine/Assets/Scripts/Drivers/RotationalDriver.cs
KyroVibe: Its a matter of removing warnings

BrandonPacewic (approved):

Markytt (approved):

PR #851, opened 2023-06-26
Assign a robots parent object to the correct layer

LucaHaverty:
### Description
Fixed a bug where an error would occur when removing a robot then spawning another one. It was caused by the wrong physics layer being added to the dynamic layers queue when a robot was removed. I fixed it by assigning a robots parent object to the correct physics layer when it is spawned.
### Objectives

- [x] Fix issue

### Testing Done
- Tested a lot of different combinations of spawning and removing robots


BrandonPacewic (approved):
Nice and easy!

a-crowell (approved):
Fixed and doesn't seem to break anything else. Nice job!

PR #854, opened 2023-06-27
API Nullable Fixes

HiceS:
### Description

There were a bunch of warnings about nullable types since they were about half used in the original API code. At first I enabled them everywhere but the entire codebase ignores null types and there are many many errors. So for this PR given it's my last day I fixed as many as I could and reduced the warnings to 0. 

There exists a few places where nullable types are actually fairly impossible unless we redefine the types to be optional. Such as the templated default returns. 

### Objectives
- [x] Get rid of the nullable warnings
- [ ] Have the API return the real types if it can be null 

### Testing Done
- Started Synthesis
- Drove around a bit
- Loaded Field / Robot

### DIFF
--- a/api/Api/Api.csproj
+++ b/api/Api/Api.csproj
@@ -28,7 +28,9 @@ For support contact frc@autodesk.com</PackageReleaseNotes>
         <NeutralLanguage>en-US</NeutralLanguage>
         <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
         <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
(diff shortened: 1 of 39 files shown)


HiceS:
At some point in the future someone needs to enable this https://github.com/Autodesk/synthesis/blob/9f7fff09f1c40e2f4ade90f9bc9a9f631fa1c807/api/Api/Api.csproj#L31

And then ideally fix all the warnings. 
But this is a solid start.

KyroVibe (approved):
Runs fine. We'll address the rest of the API nullable issues later

BrandonPacewic (approved):

PR #855, opened 2023-06-28
Goal Zone Creation

PepperLola:
### Description
- Allows the user to create scoring zones associated with each alliance that they can score gamepieces in. They can configure the name, alliance, number of points, parent object (for moving field elements), and XYZ scale of the scoring zones.
- Includes a UI to view, add, and delete score zones.
- Adds a property to Settings to toggle rendering of scoring zones
- Also adds a scoreboard that keeps track of the score of each alliance along with the number of seconds left in a match. This will be integrated with match mode later.

~~The scoreboard currently does not appear when entering practice mode from the main menu, but I think a lot of the scoreboard integration is going to change when adding match mode so I was thinking that could be addressed then.~~ *Now it only shows once a scoring zone is created, which works when entering from the main menu.*

### Objectives
- [x] Allow the user to create, position, and rescale scoring zones and assign point values to them
- [X] Allow the user to edit and delete previously created score zones
- [X] On score events on the scoring zones and class to keep track of scores
- [X] Persistent scoreboard panel with accurate scores and time

### Testing Done
- Creating and deleting scoring zones
- Editing all properties of scoring zones
- Moving scoring zones after creation
- Cancelling config modal so that all properties reset
- Toggling visibility
- Toggling visibility while gamepieces are inside scoring zone
- Overlapping scoring zones score properly (for both teams)
- Parenting scoring zones to field elements (disallowed parenting to grounded because it disables collision trigger?)
- Parenting and unparenting multiple times to ensure transforms aren't still inherited after unparent

[JIRA Issue - Goal Zone Creation](https://jira.autodesk.com/browse/AARD-1494)
[JIRA Issue - Point System and Scoring Detection](https://jira.autodesk.com/browse/AARD-1506)

### Screenshots
Scoring Zone Config
Scoring Zone Config

Scoring Zones List
Scoring Zones List

Scoreboard
Scoreboard


engine/Assets/Scripts/Modes/MatchMode.cs
```diff
@@ -1,20 +1,52 @@
...
+    private int _redScore = 0;
+    private int _blueScore = 0;
+
     public const string PREVIOUS_SPAWN_LOCATION = "Previous Spawn Location";
     public const string PREVIOUS_SPAWN_ROTATION = "Previous Spawn Rotation";
 
     // Start is called before the first frame update
     public void Start() {
         DynamicUIManager.CreateModal<MatchModeModal>();
+        SynthesisAPI.EventBus.EventBus.NewTypeListener<OnScoreEvent>(
```
a-crowell, line 23: Looks like this isn't used/tested. Is it better to leave it commented out until we use MatchMode and can test it out?

PepperLola: Unless it's causing issues with other things, I think leave it in? I was just trying to add the same things to both practice and match mode to make it easier to integrate later, and I think I tested match mode a bit a while ago (but not with a remotely recent version I don't think).

engine/Assets/Scripts/UI/Dynamic/Panels/Information/ScoreboardPanel.cs
```diff
@@ -1,77 +1,110 @@
...
-            return u;
-        };
+        CancelButton.RootGameObject.SetActive(false);
+        AcceptButton.RootGameObject.SetActive(false);
+        Title.RootGameObject.SetActive(false);
+        PanelImage.RootGameObject.SetActive(false);
 
-        public override bool Create() {
+        Content panel = new Content(null, UnityObject, null);
+        // MainContent.SetTopStretch<Content>(Screen.width / 2 - _width / 2 - 40f, Screen.width / 2 - _width / 2 - 40f, 0);
```
BrandonPacewic, line 31: Is this still needed or can it be removed?

engine/Assets/Scripts/UI/Dynamic/Panels/Configuring/Scoring/ZoneConfigPanel.cs
```diff
@@ -0,0 +1,315 @@
...
+        _zone.GameObject.transform.localScale = new Vector3(_data.XScale, _data.YScale, _data.ZScale);
+    }
+
+    public void SetCallback(Action<ScoringZone, bool> callback) {
+        _callback = callback;
+    }
+    
+    public void SelectParentButton(Button b) {
+        if (!_selectingNode) {
+            // I don't like this; do we just want all field rigidbodies to detect collisions?
```
a-crowell, line 231: They're set to not detect collisions when freezing (FieldSimObject.cs line 46). I didn't look too much into the reason for this though, maybe it's unnecessary or just needs to be put back when you're done selecting? I did test specifically for this and the code you have now works.

a-crowell: Makes sense. What part about the current code don't you like?

PepperLola: I didn't like the whole keeping track of the initial collision-detecting state of all of the field elements because it seems inelegant, and I don't know enough about the field structure to know if there are any elements that don't detect collisions by default, or if it would be okay to set `detectCollisions` to true for all of them.

a-crowell: Gotcha. I can't say I'm sure, but if freeze affects all except grounded and currently possessed, selecting parent probably can too.

a-crowell: Ha just got your edit, sounds good.

engine/Assets/Scripts/UI/Dynamic/Panels/Configuring/Scoring/ZoneConfigPanel.cs
```diff
@@ -0,0 +1,328 @@
+using System;
+using System.Collections.Generic;
+using Synthesis.Gizmo;
+using Synthesis.UI;
+using Synthesis.UI.Dynamic;
+using UnityEditor.PackageManager;
```
KyroVibe, line 6: Either remove or add preprocessor definitions to remove this using statement when not inside the editor. Fails to compile otherwise.

KyroVibe: I don't think it's actually used so you're probably fine to just remove it

PepperLola:
@a-crowell For the timer, I took Brandon's advice and only show it when there are scoring zones created (in practice mode, it will always show up in match mode). You can remove a parent either by clicking on the button to select (while a parent is set) and then clicking the button again, or by clicking in the world but not on a field element. That is probably not very intuitive though, so maybe when there is a parent selected the button will say "Remove" instead?

Also, could you elaborate more on the double-counting bug? I wasn't able to replicate that.

KyroVibe (changes requested):

a-crowell:
@PepperLola Remove button is a good idea. I'll test it out when I get home in a few hours. I'll also send you a video of the double count

BrandonPacewic (commented):
I really like how this looks, are we going to address the double count bug in this PR?

PepperLola:
Mark found a way to reproduce the double count bug that I'm going to look at in a bit. Right now I'm addressing an issue with spawning the scoring zones modal on top of the zone config one.

engine/Assets/Scripts/UI/Dynamic/Panels/Configuring/Scoring/ZoneConfigPanel.cs
```diff
@@ -73,12 +76,23 @@ public class ZoneConfigPanel : PanelDynamic {
...
+        
+        // so that entering another panel doesn't cause the zone to be created
+        // maybe not the best way to handle this
+        // EventBus.NewTypeListener<DynamicUIManager.PanelCreatedEvent>(e => {
+        //     DynamicUIManager.PanelCreatedEvent panelCreatedEvent = (DynamicUIManager.PanelCreatedEvent) e;
+        //     if (panelCreatedEvent.Panel != this && panelCreatedEvent.Panel.GetType() != typeof(ScoreboardPanel)) {
+        //         DoCancel();
+        //         DynamicUIManager.ClosePanel<ZoneConfigPanel>();
+        //     }
+        // });
```
BrandonPacewic, line 88: Is this code block needed?

LucaHaverty:
There is a bug where if you type too high of a point value, there are overflow errors. Also I would recommend capping the length of the name so it doesn't cover up the text with the point value

LucaHaverty (approved):
Everything looks good, nice work!

BrandonPacewic (approved):

a-crowell (approved):
Awesome job! Remove button looks great

KyroVibe (approved):
Looks fantastic

PR #857, opened 2023-06-28
Fixed issue with custom wheels

Markytt:
### Description
Fixed issue where an error message would appear if at least two robots are spawned and the user exits to the main menu then re-enters synthesis. 

### Objectives

- [ ] Fix issue 

### Testing Done

- Tested the steps to repeat the bug, with two, three, four and five robots spawned. 

### DIFF
--- a/engine/Assets/Scripts/SimulationRunner.cs
+++ b/engine/Assets/Scripts/SimulationRunner.cs
@@ -173,8 +173,9 @@ namespace Synthesis.Runtime {
         /// </summary>
         public static void SimKill() {
             FieldSimObject.DeleteField();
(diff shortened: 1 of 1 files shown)


PepperLola (changes requested):
This has definitely fixed the error message! Do you think you could reset the orbit camera focus position back to (0, 0, 0) when you remove the robots? When you go back into practice mode from the main menu the camera is still focused on the location of the last possessed robot, even though it has been deleted. Aside from that, this looks very good.

PepperLola (approved):
Looks good! Works as intended.

KyroVibe (approved):
Nice job

PR #862, opened 2023-06-30
Match Mode Results

LucaHaverty:
### Description
The goal of the pr is to add a system to track any information we want to show to the player after the match. It will also include a panel to display this information during the "match results state". At the moment, not very much will be added to the panel but once match mode is in a more finished state, it should be easy to add more entries to show the user.

### Objectives
- [x] Match results tracker class
- [x] Display results on match results modal
- [x] Track points scored by each robot

### Testing Done
- Tested scoring points to see if they appear on the match results panel

### Known Issues
- Scroll menus seem to be broken at the moment for everything, not just the match results panel

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1552)


PepperLola (changes requested):
It looks good and everything works, but do you think you could make the exit button of the match results modal either go to the main menu or to the mode selector modal? Maybe just reset the whole thing for now and go to the selector and then we can update it with multiplayer later.

LucaHaverty:
The exit button on the game results page now quits to the main menu. Hunter was talking about making something to basically reset everything, so after that is done I can make an option to restart match mode.

PepperLola (approved):
Looks good!

engine/Unity.Mathematics.Editor.csproj
```diff
@@ -26,7 +26,7 @@
...
...ION_JOBS;ENABLE_MANAGED_AUDIO_JOBS;ENABLE_MANAGED_UNITYTLS;INCLUDE_DYNAMIC_GI;ENABLE_SCRIPTING_GC_WBARRIERS;PLATFORM_SUPPORTS_MONO;RENDER_SOFTWARE_CURSOR;ENABLE_VIDEO;ENABLE_ACCELERATOR_CLIENT_DEBUGGING;ENABLE_NAVIGATION_PACKAGE_DEBUG_VISUALIZATION;ENABLE_NAVIGATION_HEIGHTMESH_RUNTIME_SUPPORT;ENABLE_NAVIGATION_UI_REQUIRES_PACKAGE;PLATFORM_STANDALONE;TEXTCORE_1_0_OR_NEWER;PLATFORM_STANDALONE_WIN;UNITY_STANDALONE_WIN;UNITY_STANDALONE;ENABLE_RUNTIME_GI;ENABLE_MOVIES;ENABLE_NETWORK;ENABLE_NVIDIA;ENABLE_CRUNCH_TEXTURE_COMPRESSION;ENABLE_OUT_OF_PROCESS_CRASH_HANDLER;ENABLE_CLUSTER_SYNC;ENABLE_CLUSTERINPUT;PLATFORM_UPDATES_TIME_OUTSIDE_OF_PLAYER_LOOP;GFXDEVICE_WAITFOREVENT_MESSAGEPUMP;PLATFORM_INITIALIZES_MEMORY_MANAGER_EXPLICITLY;ENABLE_MONO;NET_4_6;NET_UNITY_4_8;ENABLE_PROFILER;DEBUG;TRACE;UNITY_ASSERTIONS;UNITY_EDITOR;UNITY_EDITOR_64;UNITY_EDITOR_WIN;ENABLE_UNITY_COLLECTIONS_CHECKS;ENABLE_BURST_AOT;UNITY_TEAM_LICENSE;ENABLE_CUSTOM_RENDER_TEXTURE;ENABLE_DIRECTOR;ENABLE_LOCALIZATION;ENABLE_SPRITES;ENABLE_TERRAIN;ENABLE_TILEMAP;ENABLE_TIMELINE;ENABLE_LEGACY_INPUT_MANAGER;TEXTCORE_FONT_ENGINE_1_5_OR_NEWER;CSHARP_7_OR_LATER;CSHARP_7_3_OR_NEWER;UNITY_EDITOR_ONLY_COMPILATION</DefineConstants>
```
KyroVibe, line 29: Please remove all these `.csproj` files  in the `engine` directory. I don't know why they are still tracking, they are ignored by the `engine/.gitignore`

KyroVibe (changes requested):
How do I play through the states to view the results panel?

engine/Unity.Mathematics.Editor.csproj
LucaHaverty: To skip through the states you press right arrow. Julian's pr adds the match timer.

engine/Assets/Scripts/Modes/MatchMode/MatchResultsTracker.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using TMPro;
+using UnityEngine;
+
+public class MatchResultsTracker {
+    public readonly Dictionary<Type, ITrackedData> MatchResultEntries = new Dictionary<Type, ITrackedData>();
```
KyroVibe, line 8: I would consider just using a List instead of a dictionary because you can get the type object by doing `GetType()`. Depends on how you use it I suppose

KyroVibe (approved):
Looks good

engine/Assets/Scripts/Modes/MatchMode/MatchResultsTracker.cs
KyroVibe: Let me know you're thoughts and then I'll merge.

BrandonPacewic (commented):
Can I ask what all the deleted files were for?

KyroVibe:
@BrandonPacewic A bunch of auto generated csproj files that were supposed to be already ignored.

BrandonPacewic:
> @BrandonPacewic A bunch of auto generated csproj files that were supposed to be already ignored.

Got it 👍

PR #874, opened 2023-07-05
Match End Options

LucaHaverty:
### Description
On the match results panel, there are now 3 options. Exit will quit to the main menu, restart will restart the match with the same settings, and configure will reset all of the robots and fields then open the match configuration menu. MiddleButton in ModalDynamic creates the unity object in a getter by calling instantiate with the accept button as a prefab.

### Objectives
- [x] Add a 3rd button to the match results panel
- [x] Add functionality to the buttons

### Testing Done
- Tested loading in multiple robots then clicking configure or restart
- Tested configure and restart with no robots spawned in
- Restarted and reconfigured the match multiple times

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1559)

PR #875, opened 2023-07-05
Configuration Mode

a-crowell:
### Description
Creates a separate configuration mode and adds motor configuration.

### Objectives
- [X] Separate configuration mode (from practice and match mode) for more accessible tools
- [X] Change motor target velocity
- [X] Have options to Cancel, Session Save, or Save motor configuration

### Testing Done
- [X] Switch between config and practice/match modes
- [X] Use each config tool
- [X] Change motor target velocity (drive/turn (on swerve)/arm)
- [X] Config mode after removing robot
- [X] Multiple different robots (tank/arcade/swerve)

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1510)
[JIRA Issue](https://jira.autodesk.com/browse/AARD-1511)
[JIRA Issue](https://jira.autodesk.com/browse/AARD-1515)
[JIRA Issue](https://jira.autodesk.com/browse/AARD-1517)


engine/Assets/Scenes/Resources.meta
a-crowell, line 1: You can ignore all Scene changes, nothing changed.

engine/Assets/Scripts/UI/Dynamic/DynamicUIComponents.cs
a-crowell, line 1: This file is from @LucaHaverty 's Match End Options PR #874. It was merged already, but seems to be gone. @KyroVibe you mentioned some merging issues this week, could this be a remnant of that?

engine/Assets/Scripts/UI/Dynamic/Modals/Configuring/ConfigMotorModal.cs
```diff
@@ -0,0 +1,261 @@
...
+                });
+            }
+            if (_robotISSwerve && _turnMotors[0].velChanged) {
+                _turnMotors.ForEach(x => {
+                    SaveToMira(x.driver as RotationalDriver);
+                });
+            }
+            _otherMotors.ForEach(x => {
+                if (x.velChanged) {
+                    if (x.driver is WheelDriver) {
```
a-crowell, line 122: There's a lot of checking and casting WheelDriver vs RotationalDriver even with the same code. Any ideas on how to simplify this? Wheel, Rotational, and Linear are all under Driver, though the motor config can only be done on Wheel and Rotational.

engine/Assets/Scripts/UI/Dynamic/Modals/Configuring/ConfigMotorModal.cs
```diff
@@ -0,0 +1,261 @@
...
+                    } else if (x.driver is RotationalDriver) {
+                        SaveToMira(x.driver as RotationalDriver);
+                    }
+
+                }
+            });
+        });
+
+        MiddleButton.SetWidth<Button>(132)
+            .SetAnchor<Button>(new Vector2(0.55f, 0), new Vector2(0.55f, 0))
```
a-crowell, line 133: Leaving it for now until UI changes, but quick clean changes are appreciated if you have ideas.

engine/Assets/Scenes/Resources.meta
KyroVibe, line 1: Did you move the Resources folder into the Scenes folder? It has to stay where it was, it's a builtin Unity thing

engine/Assets/Scripts/Modes/MatchMode/MatchMode.cs
KyroVibe, line 1: Are these changes not in dev right now?

engine/Assets/Scripts/UI/Dynamic/DynamicUIComponents.cs
KyroVibe: Yeah I'm seeing a lot of weird out-of-date changes. Make sure you fetch and pull down dev before merging it.

KyroVibe (changes requested):
The configuration looks great. I would add some hit to what you're actually changing with the slider, indicating the units that the number represents. I could tell there are a lot of missing features from dev so I'd definitely take a look at files that you don't recall changing and see how they compare to dev right now. Make sure you add support for Linear Drivers.

engine/Assets/Scenes/Resources.meta
a-crowell: Just moved it back.

engine/Assets/Scripts/Modes/MatchMode/MatchMode.cs
a-crowell: Not that I can tell. I pulled dev in yesterday and just before marking the pr ready for review. The only one I've checked code through GitHub was Match State End Options (which wasn't in dev yesterday either when I pulled dev), do you want me to go through each of the other prs this week to compare?

Which robot have you been using to test LinearDriver?

LucaHaverty:
I have a recommendation for the config motor modal. Not sure if this would make it better or worse, but maybe instead of storing separate arrays for the different ConfigMotors, they could all be in the same list and when you save the velocity's you could move the .velChanged check inside the SaveToMira function. They would reduce the need for repetitive code doing something similar on all 3 different types of motors, and make it easier to add more motor types in the future.

LucaHaverty (changes requested):
I noticed a few issues when configuring in match mode. 

- When in configuration mode while positioning the robot, clicking on the field still positions the robot which can cause some weird behavior when placing ejectors or pickup locations.
- If you go into configuration mode at any time before the match starts, the match gets stuck with the robot in the air and nothing happening even when switching back to match mode

I think both of these could be fixed by either having a separate state for just configuring the robot, or making sure configuration mode is exited when a match is started, and the robot positioning pauses when in configuration mode.

In practice mode, everything seemed to work well and was very useful. It is really nice to be able to change a robots speed. Great work!

Markytt (changes requested):
In configuration mode, when you click on pick up and then click on ejector immediately afterward (the opposite order also works), some error messages will appear saying "Object reference not set to an instance of an object". This bug doesn't seem to cause any major issues, so it isn't a big of deal, but I think it would still be a good idea to fix it. 

Screenshot 2023-07-11 095301
Screenshot 2023-07-11 095325
Screenshot 2023-07-11 095604

Markytt:
> In configuration mode, when you click on pick up and then click on ejector immediately afterward (the opposite order also works), some error messages will appear saying "Object reference not set to an instance of an object". This bug doesn't seem to cause any major issues, so it isn't a big of deal, but I think it would still be a good idea to fix it.
> 
> Screenshot 2023-07-11 095301 Screenshot 2023-07-11 095325 Screenshot 2023-07-11 095604

a-crowell:
Waiting on Fake Arm Control for LinearDriver and Match End Options for the middle button. Until then, everything else is up for testing.

engine/Assets/Scripts/UI/Dynamic/Modals/Configuring/ConfigMotorModal.cs
BrandonPacewic: Is it possible for `x.driver` to be `typeof()` something other than `WheelDriver` or `RotationalDriver`? Because if not we can let c# figure this out at runtime:

```cs
SaveToMira(x.driver as dynamic);
```

If `x.driver` can be of other types than this is probably the best solution with explicit casting.

a-crowell: Yeah it can be a LinearDriver, which I'll work on once it's fixed in Fake Arm Control. If it's similar enough, I'll look into this. Good idea, thanks.

LucaHaverty:
If you spawn a robot, click configure, click pickup, click ejector without saving pickup, then return to practice mode the camera gets messed up. Also, if you press enter while in pickup or ejector mode, you get a null reference exception. I think this is related to gizmos because I have seen similar issues with other things that use gizmos, so it may be worth addressing all gizmo related issues in another pull request.

BrandonPacewic (commented):
I clicked the configure button and my robot stopped moving.

Screenshot 2023-07-12 at 9 53 40 AM

a-crowell:
> 

Freezing physics is intentional, does movement come back when you go back into practice mode?

Markytt (commented):
The configure mode button isn't showing up for me.
Screenshot 2023-07-12 143829

BrandonPacewic:
> > 
> 
> Freezing physics is intentional, does movement come back when you go back into practice mode?

Movement does work when I click on practice mode, I did not notice the pause. Should we perhaps add something to make it clear that everything is paused while you are in that menu? I know that you were talking about a robot focus screen. Would that be for configuration mode?

a-crowell:
> > > 
> > 
> > 
> > Freezing physics is intentional, does movement come back when you go back into practice mode?
> 
> Movement does work when I click on practice mode, I did not notice the pause. Should we perhaps add something to make it clear that everything is paused while you are in that menu? I know that you were talking about a robot focus screen. Would that be for configuration mode?

Yep that's for config mode. We're waiting on doing that though, so not in this PR.

a-crowell:
> The configure mode button isn't showing up for me. Screenshot 2023-07-12 143829

It won't come up until you have a robot selected. I did just introduce and fix a bug where it won't come back after leaving config mode though. The fix is in the next commit.

PepperLola (changes requested):
I really like the session save feature! I did find a couple issues.
* I'm not sure if this is a known issue, but it seems like if you press enter to dismiss any of the gizmos in configuration mode, the camera target doesn't get set back to the position of the robot.
* Would it be possible to have it configure the robot whose button is selected when placing robots in match mode, rather than the currently possessed one? Otherwise it seems like you can only configure the last one that was spawned.
* Maybe the back button in each mode could be changed to just say back, rather than practice or match? I thought it was a bit confusing at first and that it would let me configure practice/match mode rather than return to it.

Aside from those things, it all looks good! I'm looking forward to the robot isolation bit when entering configuration mode.

engine/Assets/Scripts/UI/Dynamic/Modals/Configuring/ConfigMotorModal.cs
```diff
@@ -0,0 +1,240 @@
...
+                motor.setTargetVelocity(vel);
+        }
+    }
+
+    private class ConfigMotor {
+        public Driver driver { get; set; }
+        public float origVel { get; set; }
+        private float _force { get; set; }
+        public bool velChanged { get; set; } = false;
+        public MotorType motorType { get; set; }
```
BrandonPacewic, line 201: What's the reason for having `{ get; set; }` after each of these? I don't think it's doing anything in this case.

engine/Assets/Scripts/SimulationRunner.cs
```diff
@@ -69,6 +69,8 @@ public class SimulationRunner : MonoBehaviour {
             SetContext(RUNNING_SIM_CONTEXT);
             Synthesis.PreferenceManager.PreferenceManager.Load();
             MainHUD.Setup();
+            if (ModeManager.CurrentMode == null)
```
KyroVibe, line 72: These are readded deletions. Please remove.

engine/Assets/Scripts/UI/Dynamic/MainHUD.cs
```diff
@@ -164,4 +174,118 @@ public static class MainHUD {
...
+            if (FieldSimObject.CurrentField == null) {
+                SynthesisAPI.Utilities.Logger.Log("No field loaded!", LogLevel.Info);
+            } else {
+                if (!DynamicUIManager.PanelExists<ScoringZonesPanel>())
+                    DynamicUIManager.CreatePanel<ScoringZonesPanel>();
+            }
+        });
+
+        if (!(MatchStateMachine.Instance.CurrentState.StateName is MatchStateMachine.StateName.RobotPositioning))
+            PhysicsManager.IsFrozen = false;
```
KyroVibe, line 238: I'm running into an issue where robots are falling through the floor when I go from placement to score zone setup in match mode. This is likely the issue. PhysicsManager.IsFrozen does not work like a normal boolean and is essentially a counter. Assigning it to false like that will likely cause whatever needs to pause it to fail to actually pause.

KyroVibe (changes requested):
Looks great. A lot of the UX will change next week so get ready for that. I found couple of issues I'd like you to address real quick.

a-crowell:
Bug fixes and `LinearDriver` in the motor configuration are now in for review. Still have a little work with motor names and velocity units, but those are just changing text labels.

Markytt (approved):
Everything looks good, all the issues seem to be gone.

engine/Assets/Scripts/UI/Dynamic/Modals/Configuring/ConfigMotorModal.cs
```diff
@@ -0,0 +1,250 @@
...
+        public MotorType motorType;
+
+        public ConfigMotor(MotorType t) {
+            motorType = t;     
+        }
+
+        public void setTargetVelocity(float v) {
+            switch (driver) {
+                case (RotationalDriver):
+                    _force = (driver as WheelDriver).Motor.force;
```
KyroVibe, line 214: Causes an error when changing velocity of arms

KyroVibe (changes requested):
Only found one thing. Great work

engine/Assets/Scripts/UI/Dynamic/Modals/Configuring/ConfigMotorModal.cs
a-crowell: Fixed. Thanks for catching that. I never got the error, so let me know if it's still there.

KyroVibe (changes requested):
1. It looks like the drivers and behaviours aren't actually stopping. The physics isn't updating, but if you move an arm while in config mode, then exit config mode, it snaps to the position it was supposed to have moved to if it were moving during config mode.
2. The arm moves much slower than the set RPM, Im guessing something wrong with the unit conversions.

I found both of these with the v7 mean machine bot you put in discord

KyroVibe (approved):
Works great. We'll do UX work on this next. Great work!

BrandonPacewic (approved):
Looks like the formatting is good. Not sure why GitHub actions failed to run.

PR #881, opened 2023-07-06
Unity CEF Integration

BrandonPacewic:
### Description

Created integration for CEF ([Chromium Embedded Framework](https://en.wikipedia.org/wiki/Chromium_Embedded_Framework)) to allow us to use web development tools to create UI components.

### Objectives

- Create functionality for loading a local html file.
- Create a basic working UI with intractable elements.

### Todo

I'll update the todo list when I get a better sense of what the development path is going to look like. Currently going through so many things just to get it working in the first place. Not sure where this is going to end up currently.


BrandonPacewic:
Still not totally working but I'm making progress. Currently if you try to run the application on apple silicon the gpu will crash. Other build targets are not supported yet.

BrandonPacewic:
If you want to run the current mac build follow these steps.

1. Navigate to `synthesis/cef`.
2. Run `./build.sh`.
3. Run `./build/Release/synthesis_cef.app/Contents/MacOS/synthesis_cef`.

BrandonPacewic:
Current windows build steps.

1. Open `synthesis/cef` in Microsoft Visual Studio.
2. Run `./build.bat` from the developer powershell inside Visual Studio.
3. Run `./build/Debug/synthesis_cef.exe`

The reason for running the build script in Visual Studio is because I needed a way to use the MSVC compiler with Ninja. If you want you can configure your `PATH` to include the msvc compiler so that Ninja can access it outside of Visual Studio. Visual Studio does this for us which makes the setup much easier. 

MinGW does not appear to work with CEF because it does not know how to handle the clang specific options, Ninja on the other hand is smart enough to replace the incorrect C++ flags with the correct ones.

KyroVibe:
I am so very sorry Brandon. You can express your grievances [here](https://forum.unity.com/)

BrandonPacewic:
> I am so very sorry Brandon. You can express your grievances [here](https://forum.unity.com/)

😢

PR #886, opened 2023-07-12
Field Data Storage

KyroVibe:
### Description
Be able to store field specific information in the field Mirabuf files.

### Objectives
- [x] Refactored ScoringZoneData to support serialization
- [x] Added new field data to SimulationPreferences
- [x] Added the distinction between RobotData and FieldData

### Testing Done
- Spawned in the 2018 field, created scoring zones making sure to alter transformation and select parents.

### Known Issues
- ~~Points seem to be counting double. I have no idea where that started.~~ An existing issue.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1495)


PepperLola (changes requested):
Looks good overall! I only found a few issues
* When I didn't assign a parent to a scoring zone, when I tried to edit it again, the grounded node was automatically assigned and the whole field base was highlighted. The button state was also messed up, and clicking "Remove" led me to select another game object immediately, and the field stayed highlighted for the rest of the match.
* Also, when you edit and then accept the scoring zone config panel, all of the scoring zones are hidden if the option is enabled, but they should only be hidden when you exit the scoring zones modal.
* If you remove a parent object from a scoring zone and then click on the button to fully unassign it, trying to click accept will cause an ArgumentNullException because the Name can't be null

I also thought previously that removing a parent would just set it to null and then you would have to click the button again to select a new one, it doesn't look like something you changed? I just think that would make the most sense because you can't select the grounded node as a parent.

KyroVibe:
> Looks good overall! I only found a few issues
> 
>     * When I didn't assign a parent to a scoring zone, when I tried to edit it again, the grounded node was automatically assigned and the whole field base was highlighted. The button state was also messed up, and clicking "Remove" led me to select another game object immediately, and the field stayed highlighted for the rest of the match.
> 
>     * Also, when you edit and then accept the scoring zone config panel, all of the scoring zones are hidden if the option is enabled, but they should only be hidden when you exit the scoring zones modal.
> 
>     * If you remove a parent object from a scoring zone and then click on the button to fully unassign it, trying to click accept will cause an ArgumentNullException because the Name can't be null
> 
> 
> I also thought previously that removing a parent would just set it to null and then you would have to click the button again to select a new one, it doesn't look like something you changed? I just think that would make the most sense because you can't select the grounded node as a parent.

I was working under the assumption that it would be parented to the grounded node by default as that's how it's handled with intake and ejectors. Having objects stored under just the SimObjects root gameobject and not a node has been something I've tried to avoid

BrandonPacewic (approved):

PepperLola (approved):
Looks good!

PR #887, opened 2023-07-12
Mix & Match Robot Creation

LucaHaverty:
### Description
Create a robot in Synthesis out of smaller parts. Eventually, we can add several pre-made parts, but the user will also be able to make their own parts. The point of the part editor is to add connection points to parts that specify where they connect to other parts. In the robot editor, different parts are connected together to form full robots.

## Objectives:

- Importer changes:
  - [x] System for different object to be combined at connection points
  - [x] Change the robot importer to take in an array of mirabuf files and spawn all of them as one working sim object
  - [x] Figure out how to connect 2 multiple mirabuf robot parts together with fixed joints
  - [x] Make a way to specify the location of different parts and have the joints between them auto generated
  - [x] Combine robot parts grounded objects onto one rigidbody to remove the need for fixed joints between them

- Mix and match file management
  - [x] Serialize parts and robots to json and save those files in the appdata folder
  - [x] Load parts and robots from json files in the appdata folder
  - [x] Create a new part or robot
  - [x] Delete a part or robot and make sure other robots that depend on it don't get messed up
 
- Part and robot Editor
  - [x] Be able to save/load part and robot transform data to json
  - [x] Create a part editor to modify part json files
  - [x] Create a robot editor to modify robot json files
  - [x] Ability to parent parts to each other

### Testing Done
- Button binding of different robots with the same drivetrain
- Two of the same part on one robot
- Adding, removing, renaming, and modifying robot json files during runtime through file explorer as well as in synthesis
- Importing of regular robots and fields

### Robot editor controls
- R to rotate
- Shift + R to rotate backwards
- Tab to cycle through connection points

Testers: part mirabuf files go in the mix and match mira folder (AppData\Roaming\Autodesk\Synthesis\MixAndMatch\Mira on Windows). This folder should be automatically created when the part mira modal is opened while making a new part. Ask me or hunter if you want a couple test part files. The UI/UX of the robot builder definitely needs some more work, especially around selecting parent objects, so let me know if you have any ideas.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1566)


api/Api/Controller/ControllableState.cs
```diff
@@ -23,8 +23,8 @@ public partial class ControllableState {
 
 		public bool HasUpdates => ModifiedSignals.Count > 0;
 
-		public ControllableState(Signals signals) {
-			signals.SignalMap.ForEach(x => {
+		public ControllableState(Signals[] allSignals) {
```
KyroVibe, line 26: Switch to `params Signals[] allSignals` so it can still take single Signals

KyroVibe (changes requested):

a-crowell (changes requested):
Wow this is so cool! Everything for creating the parts and playing around with the robot worked great!


*Function Ideas:*

Is there a way to move the part editor in the scene, especially when there's a field loaded? The part can get hidden when there's something in the center of the field. On a similar note, having the physics freeze would help when a robot's in the way.

Can users choose which connection points they want to attach? When you have multiple it just seems to pick the first one.

Do we want any updating? Like when you edit the part, it'll adjust robots that use it or when you edit a robot, it'll adjust any on the field. I'm on the fence about it.

Is the elevator skewed for anyone else? Or is this not skewed and I simply misunderstood the arrows?
0811MixMatch3


*UX Ideas:*

UX looked pretty good as long as we have good tutorials for it. Maybe buttons guiding users to create if there are no robots/parts? When there are no robots in the spawn mix and match, we could replace the dropdown with a phrase "No robots built yet" and a "Build robot" button that takes users to the robot editor. Same with parts for the robot editor. It would be cool to bring the users back to the spawning modal when they're done, but I don't think it's necessary. If you do this to Mix and Match, later we should add it to regular robot/field spawning, no robots/fields downloaded leading to the Download Asset modal.

The Choose Your Part/Robot modals could be scrolling lists instead. Similar to Scoring Zones with the edit and delete buttons on each part and the add button at the bottom.

It would be nice to automatically select a part after users add it. Same thing with connection points in the part editor.


*Errors:*

In Robot Editor, the second time you press the parent button in the same session for one part.
0811MixMatch

While in Robot Editor (the panel selecting parts), opened mix and match modal and went through naming before canceling back out to robot editor. Haven't been able to recreate though.
0811MixMatch2


Overall, awesome job! I really like the catches you have in there for stuff like naming bots the same name and deleting a part being used. Preventing a lot of breaks, but also lets the user know not to do that.

KyroVibe (commented):
image
Very awesome work. I think there a handful of little tweaks we should come up with to make it a little more user friendly surrounding the attachment nodes and what not, but that can wait until later. I'm going to continue testing it to make sure normal bots still work but its looking really good.

BrandonPacewic (commented):
Are we moving this to sprint 4?

Markytt (changes requested):
In the part editor, if you exit before saving the part, that will cause error messages to appear. 

Screenshot 2023-08-17 094934

Screenshot 2023-08-17 094958

Other than that issue, everything else works great.

a-crowell (approved):
Looks great! Love the editor isolation.

List of bugs for later:

In robot editor, put an elevator on a swerve. With elevator still selected, add an arm and move your mouse over the bot. Then you get an elevator helicoptering away.
0817MixMatchElevatorFly


Move an object another is connected/parented to. Less of a bug, and more of a cool feature to add, making it follow.
0817MixMatchMoveParent

The more buggy part of the previous "bug." This was the same bot after then moving the arm onto the elevator
0817MixMatchMoveParent2

If you don't have your swerve on the bottom, likely due the swerve assuming it has the lowest wheels.
0817MixMatchSwerve
0816MixMatch

This is just a list for later of mostly user failure stuff--no crashes even! Great work!

KyroVibe (approved):
Can't wait for our users to cause absolute mayhem

PR #883, opened 2023-07-11
Multiplayer robot selection

BrandonPacewic:
~~Blocked by #869.~~

### Description

Implement client robot selection during connection to multiplayer lobby. Also added a basic outline for what the server and client modes are going to look like.

Not my best code and I would be in favor of re-writing it later but it's a good outline for what multiplayer should look like.

### Objectives
- [x] Create test mode for starting an empty server.
- [x] Create test mode for connecting to an empty server.
- [x] Implement connected client downloading server side robots.
- [x] Create menu for robot selection.
- [x] Move away from testing modes.

### To Test

- Build the project and run it.
- Open up a server in either the unity editor or the built application.
- Then connect to it in either the unity editor or the built application.
- Send some robot data and pick a robot.
- Check out the list of connected clients.

[JIRA Issue - Client Upload Robot Data](https://jira.autodesk.com/browse/AARD-1561)
[JIRA Issue - Multiplayer Robot Selection](https://jira.autodesk.com/browse/AARD-1516)

PR #888, opened 2023-07-13
Client Side Ghost Bot

PepperLola:
### Description
Serves to demonstrate client-host communication through the server for one user. The "ghost" bot is controlled by the user and has only behaviours that take input and generate signals. These signals are sent through the server to the host who does the physics simulation and who applies these signals to their own copy of that user's robot (the "host" bot in this PR). After applying the signals, the host responds with the transform data of each node of the robot. The client ghost bot receives these transforms and applies them to the corresponding nodes on its own instance. The ghost bot still lags behind the host bot, and I think it might be worth seeing if we can optimize the server-client communication.

There is no management of host vs. client and the host/ghost bots are hardcoded into the server test mode. Once the other multiplayer PRs that are in progress are merged, we can go through and get everything working together so that hosts and clients are designated properly when on different machines.

### Objectives
- [X] Client sends signal data and receives transform data to apply

### Testing Done
- Tested with Dozer, 2471, 6443

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1499)


api/Api/Aether/Lobby/LobbyClient.cs
```diff
@@ -45,8 +47,8 @@ private class Inner : IDisposable {
             private readonly LobbyClientHandler _handler;
             public LobbyClientHandler Handler => _handler;
 
-            private ReaderWriterLockSlim _transformDataLock;
-            private Dictionary<ulong, ServerTransforms> _transformData;
+            public ReaderWriterLockSlim _transformDataLock;
+            public Dictionary<ulong, ServerTransforms> _transformData;
```
BrandonPacewic, line 51: If we want to make these public we need to update the variable names.

engine/Assets/Scripts/Modes/ServerTestMode.cs
BrandonPacewic, line 1: What's our plan for leaving this file in on master?

engine/Assets/Scripts/SimObjects/RobotSimObject.cs
```diff
@@ -281,6 +304,8 @@ public class RobotSimObject : SimObject, IPhysicsOverridable, IGizmo {
     }
 
     public override void Destroy() {
+        Client.Dispose();
+        Client = null;
```
BrandonPacewic, line 308: Do we need to check if the client is `null` here?

engine/Assets/Scripts/Modes/ServerTestMode.cs
KyroVibe: @BrandonPacewic We'll wait for your branch with the new modes then work from there

BrandonPacewic: 👍

BrandonPacewic (approved):

BrandonPacewic:
Just need to fix the conflicts now that we merged #883. Other than that I'm looking forward to putting this all together.

BrandonPacewic:
I pushed one small change, I realized I left in some `using`s that were not needed inside of `LobbyServer.cs`.

PepperLola:
Something from dev introduced a lot of lag again. All of the communication still works, but the host bot only started moving ~17s after I pressed a key and the ghost bot started receiving position updates after about a minute.

api/Api/Aether/Lobby/LobbyClient.cs
```diff
@@ -367,27 +366,26 @@ internal class LobbyClientHandler : IDisposable {
                     var startRead = DateTime.UtcNow;
                     while (bytesRead != msgSize && (DateTime.UtcNow - startRead).TotalMilliseconds < READ_TIMEOUT_MS) {
                         bytesRead += stream.Read(msgBuf, bytesRead, msgSize - bytesRead);
-                        Thread.Sleep(10);
```
PepperLola, line 370: Is reading from a network stream particularly expensive? I don't see why we needed a `Thread.Sleep` and removing it completely removed the lag.

BrandonPacewic: Well if removing it improves the lag then its the sleep that seems to be more expensive.

engine/Assets/Scripts/Modes/Aether/ServerTestMode.cs
```diff
@@ -39,7 +62,37 @@ public class ServerTestMode : IMode {
...
     }
 
-    public void Update() {}
+    public void Update() {
+        var transformsList = new List<ServerTransforms>();
+        if (_ghost is null || _host is null)
+            return;
+        if (_ghost.Client is null || _host.Client is null)
+            return;
+
```
LucaHaverty, line 71: There are quite a few nullable reference type warnings around here that should probably be addressed when we integrate multiplayer into match mode.

LucaHaverty (approved):
The ghost bot seems to work well and not lag too far behind! I found a bug to do with exiting to the main menu then going back into server test mode, but because the point of this pr is just to get the ghost bot working, you probably don't have to worry about that now.

Nice work, I can't wait for match multiplayer!

KyroVibe:
Looks like Multiplayer is in the integration stages now

PR #891, opened 2023-07-18
Toast System

KyroVibe:
### Description
Updated the toast system to use new UI. Also added a custom tweening library because the current one keeps breaking for no reason.

image

Toasts now reside in the lower-right corner and populate upwards as new toasts are made. Current it is configured to only show 5 at a time. Toasts remaining past 5 are "swiped" aside and out of view. The toasts currently disappear when modals are opened and remain when panels are open. Currently there can be a maximum of 20 toasts, each one only existing for 30 seconds. All of these parameters can be configured in the `engine/Assets/Scripts/UI/Dynamic/Toaster.cs` class.

[image](https://www.youtube.com/watch?v=QUverfe66FU)

Heres a demo of how the toasts move with their tweens.

image

I'm currently unsure about how big these should be. I feel like at the moment they take up a ton of room. That being said, most toasts we do send should be addressed immediately.

### Objectives
- [x] Update Toast UI
- [x] Update Toast System
- [x] Create new Tween system
- [x] Create new SubScreenSpace object to hold background UI elements

### Testing Done
- Currently has a button in practice mode (and doesn't show up in executeables) that opens a random toast
- Tried exiting out to the main menu and coming back in multiple times and all it does is rest the toasts, which is fine.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1554)


a-crowell (commented):
Looks good, I like the animation and shadow. I agree they could be a little smaller, but they aren't too bad right now. Is there a way to expand toasts other than errors?
image

KyroVibe:
> Looks good, I like the animation and shadow. I agree they could be a little smaller, but they aren't too bad right now. Is there a way to expand toasts other than errors?

Yes, but I only included errors. The idea being that if we want like maybe a help popup or something, you can specify a callback and it will show the button again

Markytt (approved):
It appears to work fine, I wasn't able to find any issues.

a-crowell (approved):
Great work

BrandonPacewic (approved):
I think it looks good, if you want to avoid fading then I'm not sure how I would change it.

PepperLola (approved):
This all looks good! The downward-facing arrow to expand the error toasts made it seem like the toasts would expand downward rather than opening a modal, and maybe having the whole toast be clickable to open the modal might be more intuitive? Then maybe the down arrow could point to the right instead to indicate that you can expand the toast. That's just my opinion and if you disagree you can definitely keep it how it is.

KyroVibe:
> This all looks good! The downward-facing arrow to expand the error toasts made it seem like the toasts would expand downward rather than opening a modal, and maybe having the whole toast be clickable to open the modal might be more intuitive? Then maybe the down arrow could point to the right instead to indicate that you can expand the toast. That's just my opinion and if you disagree you can definitely keep it how it is.

Ye, I just tossed some random icon in for now because Annie didn't have the exit and open button mocked up yet. I'll look into making it clickable

LucaHaverty (commented):
Seems to work great!

It would be nice if clicking anywhere on the modal expanded it instead of just on the small expand button. Also, I wonder if the toast could expand instead of a separate modal popping up. Something like this:

image (1)

KyroVibe:
> This all looks good! The downward-facing arrow to expand the error toasts made it seem like the toasts would expand downward rather than opening a modal, and maybe having the whole toast be clickable to open the modal might be more intuitive? Then maybe the down arrow could point to the right instead to indicate that you can expand the toast. That's just my opinion and if you disagree you can definitely keep it how it is.

I'll wait for Luca's new monobehaviour for receiving mouse and click events

PR #897, opened 2023-07-20
Main Menu UI Updates

a-crowell:
### Description
Update the Main Menu with new UI.

### Objectives
- [X] Update background
- [X] Move social media to the top right and change icons
- [X] Move logo to the top left
- [X] Add graphic design on the left (once decided on by Sara)
- [X] Create launch button
- [X] Create a second page with singleplayer and multiplayer buttons

### Testing Done
- [X] Click on each upper corner link
- [X] Click and observe slide after launching
- [X] Enter each different mode
- [X] Different aspect ratios

Note: Contact still goes to Discord, still working with Annie on that.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1518)


KyroVibe (approved):
Added some last minute tweaks, great work!

LucaHaverty (changes requested):
Looks really good! Couple things I noticed:
- When choosing a multiplayer mode, the "connect to a server" text has a typo

- I don't remember if it was like this before, but the modal for selecting a mode probably should not have a cancel button on it
- The "socials" text does not scale up when you hover over it
- The rest of the links are not scaling uniformly when you hover over them because increasing the font size causes the text to move down. If you set the vertical alignment of the text to middle, this should be fixed
image

a-crowell:
> 

Thanks! I removed the cancel button, but later let's look into bringing users back to the single/multiplayer screen.

a-crowell:
@KyroVibe Would the tweaks you made cause this when the screen is more square? I like both versions of the movement, but it looks like having the objects show outside the screen space is more versatile aspect ratio-wise.

Also note the screen space is just a mimic of the blur. I can look into making it more accurate by either having screen size adjust more similarly to the image or using a shader to make the blur instead.

image

KyroVibe:
> 

Yes, that was me, but tbh as long as its fine at 16:9, I don't care for the beta

LucaHaverty (commented):
Great job! Everything looks good

LucaHaverty (approved):

PR #904, opened 2023-07-27
`clang-format` action fix

BrandonPacewic:
### Description

Fixed the action. LLVM updated the `llvm.sh` install script we started using and it was causing unresolved dependencies unrelated to `clang-format`. Instead of installing the whole llvm project we are now only installing `clang-format`.

This also improved the runtime of the action ten fold.


a-crowell:
Code looks good and didn't bring up any errors when run. Are there other things we should test/look out for?

KyroVibe (approved):

PR #906, opened 2023-07-28
Quick Fixes

KyroVibe:
### Objectives
- [x] MainHUD always opens when closing a modal
- [x] Fields don't break when they are first saving stuff


a-crowell (approved):
Looks good! Do want to slip cleaning up the log with the close/open panels, etc in this pr?

PR #920, opened 2023-08-04
Concitus Quattuor

KyroVibe:
# July 31st - August 11th


a-crowell (changes requested):
Saving with the new Mix and Match doesn't work yet, at least on my laptop. With the limited testing I've done, for regular robots, configuring pickup/ejector and all drivers and for mix and match robots, just the LinearDriver. Controls don't save half the time either. The issue for drivers on regular robots at least is that they aren't seeing their motorRef key in Mira, making them use the default motor and preventing saving. Still testing to find the full scope and exact replication steps. Are others having this issue?

a-crowell (approved):
Saving is all good now.

PR #921, opened 2023-08-06
Selection Race Case Bug

KyroVibe:
### Description
The selection inputs for the exporter had a race case that just happened to work for a while... until a recent update.

### DIFF
--- a/exporter/SynthesisFusionAddin/src/UI/ConfigCommand.py
+++ b/exporter/SynthesisFusionAddin/src/UI/ConfigCommand.py
@@ -1652,6 +1652,8 @@ class MySelectHandler(adsk.core.SelectionEventHandler):
             self.selectedOcc = adsk.fusion.Occurrence.cast(args.selection.entity)
             self.selectedJoint = args.selection.entity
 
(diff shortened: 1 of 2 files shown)


exporter/SynthesisFusionAddin/src/UI/ConfigCommand.py
```diff
@@ -2109,22 +2148,17 @@ def notify(self, args):
             elif cmdInput.id == "wheel_select":
                 self.reset()
 
-                wheelSelect.isEnabled = False
-                wheelSelect.isVisible = False
+                # wheelSelect.isVisible = False
```
BrandonPacewic, line 2151:
```suggestion
```

BrandonPacewic (commented):
Looks like the `privacy/beare` workflow is failing for some reason.

BrandonPacewic (approved):
I think everything is working fine. I'm going to need to spend more time with the exporter at some point to get more comfortable with it.

a-crowell:
It exported a CAD model into Synthesis successfully. Are there more steps to reproduce the bug?

PR #922, opened 2023-08-08
Fixed COM

KyroVibe:
### Description
Removed Physical Properties of Part definitions that have no bodies.
### Test
Verified on TestArm for Mix and Match that Luca and I discovered had the issue with the center of mass being off.

Fusion COM:
image

Broken Unity COM:
image

Fixed Unity COM:
image


LucaHaverty (approved):
Looks good, nice work!

BrandonPacewic (approved):

PR #923, opened 2023-08-10
Small beta changes

BrandonPacewic:
Here are some small changes from the beta feedback. Taking a break from my cef binary problem 😢 

Updates:

- Collapsing the main HUD while in config mode will now exit config mode.
- Increased the max pickup size.
- Increased the max ejector velocity.
- While in match mode robot placement state main HUD is now hidden.

Features Implemented:

- Added configuration option for pickup holding multiple gamepieces at a time.

### DIFF
--- a/engine/Assets/Scripts/UI/Dynamic/MainHUD.cs
+++ b/engine/Assets/Scripts/UI/Dynamic/MainHUD.cs
@@ -1,5 +1,3 @@
-using System;
-using System.Collections.Generic;
 using DigitalRuby.Tween;
 using Modes.MatchMode;
 using Synthesis.Gizmo;
@@ -8,11 +6,14 @@ using Synthesis.Runtime;
 using Synthesis.UI.Dynamic;
 using SynthesisAPI.EventBus;
 using SynthesisAPI.Utilities;
+using System;
+using System.Collections.Generic;
 using UI;
 using UnityEngine;
 using UnityEngine.SceneManagement;
 using UnityEngine.UI;
 using Utilities.ColorManager;
+
 using Button  = Synthesis.UI.Dynamic.Button;
 using Image   = Synthesis.UI.Dynamic.Image;
 using Logger  = SynthesisAPI.Utilities.Logger;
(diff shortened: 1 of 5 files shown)


LucaHaverty (approved):
Looks good, I didn't find any issues. Nice work!

Markytt (changes requested):
Everything is working well. Could you make it so that the main HUD also disappears right after placement state when the spawn score zone panel appears.

a-crowell (changes requested):
Love the multiple gamepiece pickup. By calling `LeaveConfig()` upon `MainHUD` collapse though, you also leave config when you open any modals in config mode.  You can call `LeaveConfig()` when the close button is clicked to avoid this.

engine/Assets/Scripts/UI/Dynamic/MainHUD.cs
```diff
@@ -172,19 +176,14 @@ public enum DrawerPosition {
 
         _closeButton.OnClicked += (b) => Collapsed = true;
```
a-crowell, line 177: Here's what I mean.

BrandonPacewic:
Do we want the main hud to start collapsed after match mode setup?

a-crowell:
> Do we want the main hud to start collapsed after match mode setup?

Could go either way, but collapsed is a good idea. On a similar note, do we want the robot tooltips when they start the match?

KyroVibe (approved):
Everything worked as expected. Tbh I designed the gamepiece manipulation to have multiple eventually but I didn't actually think it worked. Good work

Markytt (approved):
Everything is working, good job.

PepperLola (approved):
Is the ejector trajectory visualizer cube thing supposed to scale with the ejection speed? I thought I remembered that happening before but it doesn't right now. I'm not sure if that's an issue with this PR though so I'll approve it anyways.

BrandonPacewic:
> > Do we want the main hud to start collapsed after match mode setup?
> 
> Could go either way, but collapsed is a good idea. On a similar note, do we want the robot tooltips when they start the match?

Just going to leave it as is for this PR. Something we can have a discussion about in the future.

BrandonPacewic:
> Is the ejector trajectory visualizer cube thing supposed to scale with the ejection speed? I thought I remembered that happening before but it doesn't right now. I'm not sure if that's an issue with this PR though so I'll approve it anyways. Everything else looks good.

I thought it used to as well but it currently doesn't in dev so I'm not sure.

a-crowell (approved):
Good work!

PR #924, opened 2023-08-12
08.12 Bug Fixes

a-crowell:
### Description
Some bug fixes.

### Objectives
- [X] Made the end match results consider non-persistent scoring zones by having it remove points upon object exit instead of adding more.

### Testing Done
- [X] The different arrangements of non/persistent zones and leaving objects vs taking them out vs putting back in with both alliances.

Note: I added a button to skip to match results that I will remove before merging in.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1585)

### DIFF
--- a/engine/Assets/Scripts/UI/Dynamic/MainHUD.cs
+++ b/engine/Assets/Scripts/UI/Dynamic/MainHUD.cs
@@ -26,7 +26,7 @@ public static class MainHUD {
     private const string EXPAND_TWEEN   = "expand";
 
     private static Action<SynthesisTween.SynthesisTweenStatus> collapseTweenProgress = s => {
-        _tabDrawerContent?.SetAnchoredPosition<Content>(
+        _tabDrawerContent.CheckIfNull<Content>()?.SetAnchoredPosition<Content>(
             new Vector2(s.CurrentValue<int>(), _tabDrawerContent.RootRectTransform.anchoredPosition.y));
     };
 
@@ -41,18 +41,23 @@ public static class MainHUD {
             if (!_isSetup)
                 return;
 
-            if (_enabled != value) {
-                _enabled = value;
-                if (_enabled) {
+            if (value) {
+                if (_enabled != value) {
                     Collapsed = false;
-                    _tabDrawerContent.RootGameObject.SetActive(true);
-                    // _accordionButton.RootGameObject.SetActive(true);
-                } else {
-                    Collapsed = true;
-                    _tabDrawerContent.RootGameObject.SetActive(false);
-                    _accordionButton.RootGameObject.SetActive(false);
                 }
+
+                _tabDrawerContent.RootGameObject.SetActive(true);
+                // _accordionButton.RootGameObject.SetActive(true);
+            } else {
+                if (_enabled != value) {
+                    Collapsed = true;
+                }
+
+                _tabDrawerContent.RootGameObject.SetActive(false);
+                _accordionButton.RootGameObject.SetActive(false);
             }
+
+            _enabled = value;
         }
     }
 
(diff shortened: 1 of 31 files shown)


KyroVibe (changes requested):
Short and sweet. Nice work, however, could you set the Skip to End button to only go to 10 seconds if the timer is greater than 10 seconds.

PepperLola:
There's already a HandleScoreEvent method that does essentially the same thing but with different variables. Would it be worth moving the tracking into that method so it could all happen in one place? You'd have to figure out how to pass the tracker in but I don't think that would be too difficult.

engine/Assets/Scripts/UI/Dynamic/MainHUD.cs
```diff
@@ -445,6 +445,11 @@ public enum DrawerPosition {
             }
         }, drawerPosition: DrawerPosition.Bottom);
 
+#if UNITY_EDITOR
+        AddItemToDrawer("Skip to End", b => MatchMode.MatchTime = MatchMode.MatchTime > 10 ? 10 : MatchMode.MatchTime,
+            drawerPosition: DrawerPosition.Bottom);
+#endif
+
```
BrandonPacewic, line 452: Is it not possible to do this outside of unity? Or is it just for us during testing?

a-crowell: Tis just for us. Users can go to practice mode if they don't want the timer.

Markytt (commented):
You can get around persistent scoring by intaking the object that was scored, and then shooting the object outside of the score zone. 

Screenshot 2023-08-16 141325

Screenshot 2023-08-16 141348

Markytt (approved):
The code is working well, and outside of the small exploit I found, I didn't find any bugs.

a-crowell:
> You can get around persistent scoring by intaking the object that was scored, and then shooting the object outside of the score zone.
> 
> Screenshot 2023-08-16 141325
> 
> Screenshot 2023-08-16 141348

I've only gotten this to happen when the scoring zone is configured to be nonpersistent. If you are seeing this with persistent scoring zones, could you show me next time in office?

KyroVibe (approved):
> You can get around persistent scoring by intaking the object that was scored, and then shooting the object outside of the score zone.

I'm calling that a feature. Good work

BrandonPacewic (approved):
> > You can get around persistent scoring by intaking the object that was scored, and then shooting the object outside of the score zone.
> 
> I'm calling that a feature. Good work

Feature it is.

KyroVibe:
> > > You can get around persistent scoring by intaking the object that was scored, and then shooting the object outside of the score zone.
> > 
> > 
> > I'm calling that a feature. Good work
> 
> Feature it is.

Can't remember why but you have to disable the collision detection of the of he gamepieces when they are collected. If you don't you get some weird affects with how they are temporarily stored.

BrandonPacewic:
> Can't remember why but you have to disable the collision detection of the of he gamepieces when they are collected. If you don't you get some weird affects with how they are temporarily stored.

I'll make a note of that.

a-crowell:
> > You can get around persistent scoring by intaking the object that was scored, and then shooting the object outside of the score zone.
> 
> I'm calling that a feature. Good work

Aight, I'll make a Jira ticket

PepperLola:
> > > > You can get around persistent scoring by intaking the object that was scored, and then shooting the object outside of the score zone.
> > > 
> > > 
> > > I'm calling that a feature. Good work
> > 
> > 
> > Feature it is.
> 
> Can't remember why but you have to disable the collision detection of the of he gamepieces when they are collected. If you don't you get some weird affects with how they are temporarily stored.

Was that because the collisions would all re-trigger when enabling and disabling physics? I had to disable collisions before freezing everything or the game pieces would re-score every time a modal was opened and closed or the game was paused. It also caused an issue with temporary scoring that you mentioned but I don’t remember what that was either; something about the exit event being triggered right after enter so we couldn’t keep track of a list of scores game pieces manually?

PR #926, opened 2023-08-14
More Beta Changes

LucaHaverty:
### Description
Some more changes that were requested by users during beta testing day.

### Changes
- [x] Center cam when a robot is removed
- [x] Remove 'alpha' text in controls panel
- [x] 'Esc' to close the active modal
- [x] Remove .mira file extension in add object modal
- [x] Make tooltips last longer
- [x] UI gradient updates - made some gradients backwards like on Jira:
image
- [x] Theme editor updates - make it clear that the default theme cannot be edited, button locations for deleting are more intuitive, color values round to the nearest int
- [x]  Make robots not appear under the floor during robot placement
- [x] Fix weird scaling of the spawn location panel with different aspect ratios


a-crowell (approved):
It all works well, nice work!

engine/Assets/Scripts/UI/Dynamic/Modals/Spawning/AddRobotModal.cs
```diff
@@ -40,10 +40,12 @@ public class AddRobotModal : ModalDynamic {
...
             });
 
-            var chooseRobotDropdown = MainContent.CreateDropdown()
-                                          .SetOptions(_files.Select(x => Path.GetFileName(x)).ToArray())
-                                          .AddOnValueChangedEvent((d, i, data) => _selectedIndex = i)
-                                          .SetTopStretch<Dropdown>();
+            var chooseRobotDropdown =
+                MainContent.CreateDropdown()
+                    .SetOptions(
+                        _files.Select(x => Path.GetFileName(x).Substring(0, Path.GetFileName(x).Length - 5)).ToArray())
```
BrandonPacewic, line 46: I feel like we could get some unintended consequences from doing this. However all the files should be `.mira` anyways so I think it should be fine.

KyroVibe: I'd use this instead: https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getfilenamewithoutextension?view=net-7.0

That being said, I changed where this code is stored in my multiplayer branch it prolly won't matter.

KyroVibe (approved):

KyroVibe:
Will the removal of "alpha" work for the tooltips as well?

Markytt (approved):

LucaHaverty:
> Will the removal of "alpha" work for the tooltips as well?

It should, assuming it uses the same code to generate the names. That feature was added in Ayla's controls branch I believe, so we can check when that is merged.

BrandonPacewic (approved):
Nice and easy. Will make a note to check the keycode names in the tooltips later.

KyroVibe (changes requested):
image
There isn't a way to continue to the next step in match mode

LucaHaverty:
> Should the theme editor panel tween out and back every time you change the selected theme save?

Good point, that has been changed now

KyroVibe (changes requested):
While the accept button now shows up, the accept button on the robot placement panel doesn't work.

Also could you just disable the entire MainHUD while setting up robots and scoring zones in match mode?

KyroVibe (approved):
Button works. Good work

PR #931, opened 2023-08-21
Driver Acceleration

a-crowell:
### Description
Adding acceleration to joint motion and allowing users to adjust it through stall torque.

### Objectives
- [X] Fix WheelDriver acceleration
- [X] Implement Linear/RotationalDriver acceleration
- [X] Allow users to customize acceleration
- [X] Smooth UX and real world accuracy/applicability

Note: Still waiting on saving to get merged in to really test the config joints saving and adjusting units to make values easier to handle, but we can get started testing that acceleration is in there and makes relative sense.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1583)


LucaHaverty (changes requested):
The acceleration seems to work well, but I noticed that if you spawn in a dozer, it drives exactly how robots did before acceleration was implemented, but if you change the acceleration at all, even just turn it down by 1 rmp/s, it accelerates way slower. It could be that the default is way more than 350, but the joint config won't show values over 350. Also you might want to consider increasing the max acceleration a lot in general, because the current max feels quite slow.

Also, would we want to consider having the drivetrain motors in m/s/s that the actual robot would move if at all possible? I know that a lot of teams use that unit when working with drivetrains.

Overall, nice work! This will be really nice to have and make robots feel more realistic.

a-crowell:
> The acceleration seems to work well, but I noticed that if you spawn in a dozer, it drives exactly how robots did before acceleration was implemented, but if you change the acceleration at all, even just turn it down by 1 rmp/s, it accelerates way slower. It could be that the default is way more than 350, but the joint config won't show values over 350. Also you might want to consider increasing the max acceleration a lot in general, because the current max feels quite slow.
> 
> Also, would we want to consider having the drivetrain motors in m/s/s that the actual robot would move if at all possible? I know that a lot of teams use that unit when working with drivetrains.
> 
> Overall, nice work! This will be really nice to have and make robots feel more realistic.

You're right in that the force is way above 350 right now. I'll talk to Hunter about getting more realistic values in the mira files we send out. Also good idea on the m/s for drive! I'll work on that and adjusting the slider limits.

KyroVibe (changes requested):
So the max acceleration slider doesn't seem to cover enough of a range. Also, when using mean machine, while when the arm hits the top limit, the velocity doesn't carry, when it hits the bottom is does and takes a bit to snap back and start moving in the other direction.

KyroVibe:
You might want to consider using a text box for the acceleration and speed rather than a slider

BrandonPacewic (commented):
Something about this doesn't seem quite right. Are we using wheel friction to calculate deceleration or something else?

a-crowell:
> Something about this doesn't seem quite right. Are we using wheel friction to calculate deceleration or something else?

The wheel velocity is determined based on the values they entered, then friction with the floor determines robot velocity.

BrandonPacewic:
> > 
> 
> Looking into allowing users to change in M/S/S for drive, I don't have the time right now to go through all of the friction. I could do a basic one not including all friction though. Do we think it would be useful? Do your teams account for friction anyway? My team starts from the basic rpm to m/s then tunes a conversion factor based on what we actually measure.

I was mainly talking about it in terms of feel. I don't believe that any teams would want real word friction simulation.

engine/Assets/Scripts/Drivers/RotationalDriver.cs
```diff
@@ -281,6 +293,8 @@ public enum RotationalControlMode {
                         }
 
                         _fakedTheta = Mathf.Clamp(_fakedTheta, -180, 179);
+                        if (_fakedTheta == 179 || _fakedTheta == -180)
```
KyroVibe, line 296: Depending on how the limits are setup, the limit can be anything in between -180 and 179. Check for no change in _fakedTheta instead.

KyroVibe (changes requested):
I think we should remove doing any kind of acceleration for wheels temporarily. There is a lot of weirdness with how they should accelerate that is going to be extremely difficult to capture.

KyroVibe (approved):
K I added a couple more tweaks and also removed bounciness with the wheels to hopefully help the random collider issues. I'll start fixing some of the currently exported bots to have better accelerations

LucaHaverty (approved):
Everything looks good and seems to function well.

Nice work!

PR #934, opened 2023-08-22
Bug fix: Buttons colors not changing

Markytt:
### Description
Home button, Spawn button, and Back button wouldn't change colors when color changes are made in color manager

### Objectives
- [x] Make button colors change 
- [x] Make color of button text and score board text change 

### Testing Done
- Changed Background HUD and Main Text multiple times to make sure the buttons are changing to the correct colors 

### DIFF
--- a/engine/Assets/Scripts/UI/Dynamic/Panels/Information/ScoreboardPanel.cs
+++ b/engine/Assets/Scripts/UI/Dynamic/Panels/Information/ScoreboardPanel.cs
@@ -51,11 +51,11 @@ public class ScoreboardPanel : PanelDynamic {
             _bottomContent.SetHeight<Content>(bottomHeight);
 
             _time = _topContent.CreateLabel(topHeight)
-                       .SetStretch<Label>()
-                       .SetText(MatchMode.MatchTime.ToString())
-                       .SetHorizontalAlignment(HorizontalAlignmentOptions.Center)
-                       .SetVerticalAlignment(VerticalAlignmentOptions.Middle)
-                       .SetFontSize(40);
+                        .SetStretch<Label>()
+                        .SetText(MatchMode.MatchTime.ToString())
+                        .SetHorizontalAlignment(HorizontalAlignmentOptions.Center)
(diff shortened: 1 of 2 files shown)


engine/Assets/Scripts/UI/Dynamic/Panels/Information/ScoreboardPanel.cs
```diff
@@ -18,8 +19,8 @@ public class ScoreboardPanel : PanelDynamic {
 
     private readonly bool _showTimer;
 
-    private Label time, redScore, blueScore;
-    private Content topContent, bottomContent;
+    private static Label _time, _redScore, _blueScore, _redLabel, _blueLabel;
```
LucaHaverty, line 22: Should these be static? If AssignColors is made not static, then I don't think these would have to be

engine/Assets/Scripts/UI/Dynamic/Panels/Information/ScoreboardPanel.cs
```diff
@@ -114,13 +117,22 @@ public ScoreboardPanel(bool showTimer = true)
...
+        _redScore.SetText(Scoring.redScore.ToString());
+        _blueScore.SetText(Scoring.blueScore.ToString());
     }
 
-    public override void Delete() {}
+    public override void Delete() {
+        EventBus.RemoveTypeListener<ColorManager.OnThemeChanged>(AssignColors);
+    }
+
+     public static void AssignColors(IEvent e) {
```
LucaHaverty, line 132: This too?

engine/Assets/Scripts/UI/Dynamic/Panels/Information/ScoreboardPanel.cs
Markytt: Maybe AssignColors could be made not static, but I don't think I'm knowledgeable enough to confidently make that change, so for now, the variables need to be static so there won't be errors.

a-crowell (approved):
Looks good, great work Mark!

BrandonPacewic (approved):
Excellent!

PR #939, opened 2023-08-24
Scoring zone editor tool

LucaHaverty:
### Description
A custom editor tool for creating and editing scoring zones quickly inside of the unity editor. This is only meant to be used by us, and the scene it's in should not included in any builds. To use, open the "EditorTools" scene, then find the "ScoringZoneEditor" object in the scene hierarchy. Here, fields can be loaded, and scoring zones can be added and saved. After loaded a field, it's scoring zones will appear as children of it's child object, and each have a "ScoringZoneConfigListener" MonoBehavior with additional settings besides name and transform. 

Also, unity editor tools can be used in the editor while not in play mode, so going into play mode isn't necessary.

Tips: If you shift/ctrl click to select multiple zones at once, settings like points can all be edited at the same time for all of them, and they can all be copied and pasted, then moved together. To rename zones, rename the unity gameobject, and to parent, drag them under a different field node.

### Objectives
- [x] Custom unity editor
- [x] Save and load zones
- [x] Edit zone settings

### Testing Done
- Loaded fields in the main sim that have been modified in the new zone config editor

image

Where to find scoring zones:
image

image


HunterBarclay (changes requested):
Could you add some kind of list of registered scoring zones in the editor tool? I'm gonna move this into a different milestone so we can get the release wrapped up

KyroVibe:
Very sorry Luca. You can express your grievances [here](https://forum.unity.com/)

PR #944, opened 2023-10-05
Release v6.1 (Prod)

KyroVibe:
### Production Update
Updating `master` to include v6.1 changes


BrandonPacewic (commented):
Does the installer still need to be updated? What does that process look like?

PR #950, opened 2023-11-07
Fixed MacOS Dependencies for Exporter

### LINKED ISSUE #948: Add-in not working
**Describe the bug**
Installed Synthesis V6 from website but the Add-in does not appear in Inventor Professional 2024 and in Fusion 360 is not opening at all.

from Synthesis log file:

```
2023-10-26 11:16:46,369 - synthesis.import_manager - ERROR - Failed
Traceback (most recent call last):
  File "C:\Users/inse9/AppData/Roaming/Autodesk/Autodesk Fusion 360/API/AddIns/Synthesis\src\general_imports.py", line 36, in <module>
    from proto import deps
  File "C:\Users\inse9\AppData\Roaming\Autodesk\Autodesk Fusion 360\API\AddIns\Synthesis\proto\deps.py", line 3, in <module>
    from os import __file__, path
ImportError: cannot import name '__file__' from 'os' (unknown location)
```

KyroVibe:
### Description
Removed the use of `os.__file__` due to new issues with it being frozen in python v3.11 (see #948). In addition, modified the path concatenation.

### Testing Done
- [x] Tested MacOS can install dependencies
- [x] Tested MacOS can install dependencies AGAIN after update
- [x] Tested Windows can install dependencies AGAIN after update

### Notes
This should hopefully resolve #948 but will be difficult to confirm do to the illusive nature of the bug.

### DIFF
--- a/exporter/SynthesisFusionAddin/proto/deps.py
+++ b/exporter/SynthesisFusionAddin/proto/deps.py
@@ -1,10 +1,8 @@
+import os
 import platform, subprocess
 from pathlib import Path
-# from os import __file__, path, system
-import os
-from src.general_imports import INTERNAL_ID;
 
-help("modules")
+from src.general_imports import INTERNAL_ID;
 
 import adsk.core, adsk.fusion
 
@@ -25,16 +23,13 @@ def getPythonFolder() -> str:
     import importlib.machinery
     osPath = importlib.machinery.PathFinder.find_spec('os', sys.path).origin
 
+    # The location of the python executable is found relative to the location of the os module in each operating system
     if system == "Windows":
-        pythonFolder = Path(osPath).parents[
-            1
-        ]  # Assumes the location of the fusion python executable is two folders up from the os lib location
+        pythonFolder = Path(osPath).parents[1]
     elif system == "Darwin":
         pythonFolder = f'{Path(osPath).parents[2]}/bin'
     else:
-        raise ImportError(
-            f"Unsupported platform! This add-in only supports windows and macos"
-        )
+        raise ImportError('Unsupported platform! This add-in only supports windows and macos')
     
     logging.getLogger(f'{INTERNAL_ID}').debug(f'Python Folder -> {pythonFolder}')
     return pythonFolder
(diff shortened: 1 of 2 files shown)


exporter/SynthesisFusionAddin/proto/deps.py
```diff
@@ -1,18 +1,80 @@
 import platform, subprocess
 from pathlib import Path
-from os import __file__, path
+# from os import __file__, path, system
```
BrandonPacewic, line 3:
```suggestion
```

Remove?

exporter/SynthesisFusionAddin/proto/deps.py
```diff
@@ -1,18 +1,80 @@
...
+import os
+from src.general_imports import INTERNAL_ID;
+
+help("modules")
```
BrandonPacewic, line 7: Still needed?

exporter/SynthesisFusionAddin/proto/deps.py
```diff
@@ -1,18 +1,80 @@
...
+            bufsize=1,
+            creationflags=subprocess.CREATE_NO_WINDOW,
+            shell=False
+        )
+    else:
+        # executionResult = subprocess.call(
+        #     command,
+        #     bufsize=1,
+        #     shell=False
+        # )
```
BrandonPacewic, line 63: Leaving this here for now?

exporter/SynthesisFusionAddin/proto/deps.py
```diff
@@ -1,18 +1,80 @@
...
+        joinedCommand = str.join(' ', command)
+        executionResult = os.system(joinedCommand)
```
BrandonPacewic, line 65: Are we planning on ignoring the security workflow?

BrandonPacewic (commented):
If it works it works.

exporter/SynthesisFusionAddin/proto/deps.py
KyroVibe: Yeah, there's no good way around it. The complaint is that there is possible command injection into the shell that python will use to run our pip commands. However, it doesn't take into account that we don't give the user an opportunity to inject anything, and if they really really wanted to, you can modify it because it's Python. At the end of the day we control the release and as long as we don't allow users to specify custom dependencies, then we're fine.

BrandonPacewic (approved):
I did not re-install fusion but reinstalling the exporter fixed my windows computer. Looks good.

PR #952, opened 2023-12-10
MacOS Installer

KyroVibe:
### Description
Created a DMG based MacOS installer with (hopefully) exporter installation support.

Screenshot 2023-12-19 at 5 55 35 PM

Tbh, it doesn't look great but the coordinates are so inconsistent, its impossible to line everything up correctly.

Instead of building another script that will run into permission issues, I've included a PDF file compiled from a markdown file to include in the installer. These instructions are pretty basic and should be easy for users to follow.


BrandonPacewic (changes requested):
On my computer it shows the addins folder as a file and it won't let me move the exporter into it.

image

BrandonPacewic (approved):

PR #954, opened 2024-02-20
Fusion Selection Bug

KyroVibe:
### Bug
Users were unable to select joints and occurrences when choosing wheel joints and game pieces.

### Cause
API changed the order in which event handlers were assessed and executed, resulting in the `InputChangedEventHandler` being executed before the `SelectionEventHandler`. This resulted in the selections being clear via `self.reset()`, and therefore caused there to be no selection for the `SelectionEventHandler`.

### Fix
Removed the `self.reset()` calls when a selection is made and instead cleared selections when the add buttons are pressed.

### Notes
Also fixed a bug where the preselection cursor didn't work for joints, as well as fixed the preselection cursor so it actually sits on top of where your cursor really is.


KyroVibe:
This PR has been user tested and confirmed to work (both the installer and the bug fix).

PR #968, opened 2024-03-19
World System and Basic Physics

KyroVibe:
### Description
This is the early stages of a core component to fission that I'm calling the World. It will be the glue between all other systems and will facilitate the main loop of the simulation. This will essentially replace the old SimulationRunner MonoBehaviour from our Unity code.

image
*Mean machine spawned in. Note that collision is disabled between dynamic bodies and no jointing has been worked on yet.*

### Tasks
- [x] Create a mock up of the World and integrate the SceneRenderer and new PhysicsSystem into it.
- [x] Create the PhysicsSystem and begin adding support for Mirabuf provided data.
- [x] Successfully load Dynamic and Static Mirabuf assemblies.

### Side Tasks
- [x] Added enum for changing core material used with mirabuf meshes. Options include Regular: standard phong shading material, Normals: show the normals of the surface via color, and Toon: a basic toon shader for a more cartoony effect.
- [x] Added an optional boolean to render Center of Mass vectors and wireframes of meshes.
- [x] Updated workflows and added unit testing and packaging workflow for fission

### Unit Tests
- [x] Parser Jolt bodies from Dozer.
- [x] Parser Jolt bodies from Mean Machine.
- [x] Check that world isn't alive by default (ran into window global issues).


PepperLola (approved):
I was able to import 2471, Dozer, and the 2018 field and all behaved as expected.

PR #969, opened 2024-03-19
Non-BRep Mesh Support

KyroVibe:
### Description
Added support for non BRep meshes inside of the Fusion exporter.

image
*This is a bird model that we've had in the share folder for YEARS now...*

### Testing Done
- Exported Bird model that is a custom mesh and imported it into v6.1


KyroVibe:
Investigating an issue with meshes that have issues

KyroVibe:
Confirmed partial meshes still export as expected.

image
*Pagani in the new Fission renderer with physics disabled*

PR #977, opened 2024-06-18
Input Manager

LucaHaverty:
### Description
An input management system that can control Synthesis using both keyboard and controller with customizable. Some controls are global and some are robot specific. The input manager works with new behavior scripts to move robots.

### Objectives
- [x] Input management system
- [x] Ability to change inputs from the controls modal
- [x] Arcade drive behavior ([JIRA Issue](https://jira.autodesk.com/browse/AARD-1663))
- [x] Generic arm behavior ([JIRA Issue](https://jira.autodesk.com/browse/AARD-1664))
- [x] Generic elevator behavior ([JIRA Issue](https://jira.autodesk.com/browse/AARD-1674))

**(the wacky slider joint behavior is fixed in [Elevator Bug](https://github.com/Autodesk/synthesis/pull/983))**

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1658)


HunterBarclay (changes requested):
image

The joint inputs are appearing under the global section.

PepperLola (dismissed):
It works pretty well (except for the slider I think). I think something about the keybind detection needs to change to work on Mac though. When doing <kbd>Shift + 1</kbd> it instead registers as <kbd>Shift + !</kbd>, and with alt it becomes <kbd>Alt + ¡</kbd> rather than detecting the base key (1) and the modifier (Shift). I can rebind the controls to <kbd>Alt + ¡</kbd> and it'll work properly, although it seems that joint 2 on the default TMM 2018 robot doesn't do anything and joint 3 controls both stages of the elevator.

It also looks like either the joint inputs aren't updating (maybe since they're being categorized as global?) or they're being added to robots that shouldn't have them, because when only Dozer is loaded I still see the arm and slider joints from the TMM 2018 bot.

a-crowell (approved):
Other than the multibot issues (controlling all robots simultaneously and controls from other robots showing up) which sounds like it will be solved later--looks great!

Someday:
* The stages of the 2471 elevator could have less impact on each other (they can move each other almost min to max), but the elevator fix commit was still a huge improvement and seems good enough for now.
* Shift/alt/ctrl combos don't do anything. Easily fixed by not using them, so again, future problem.



Looking over this was super impressive. Great job!

HunterBarclay (changes requested):
Modifier keys still don't seem to work. `Shift + 1` originally worked but after rebinding, no modifier keys responded (with or without)

LucaHaverty:
> Modifier keys still don't seem to work. `Shift + 1` originally worked but after rebinding, no modifier keys responded (with or without)

Could you give me exact steps to recreate? I haven't been able to get modifier keys to not work, but Ayla seems to have found some issues with them too.

fission/src/systems/simulation/stimulus/WheelStimulus.ts
```diff
@@ -34,9 +34,9 @@ class WheelRotationStimulus extends EncoderStimulus {
         this._wheel = wheel;
     }
 
-    public Update(deltaT: number): void {
+    public Update(_: number): void {
         if (this._accum) {
-            this._wheelRotationAccum += this._wheel.GetAngularVelocity() * deltaT;
+            this._wheelRotationAccum += 0;
```
HunterBarclay: Why'd you remove this?

LucaHaverty: I wasn't sure what it did and removed it while debugging wheel behavior, but I just now added it back in my most recent commit.

HunterBarclay: @LucaHaverty K, looks like when you added it back, you accidentally pasted in the entire line again. Fixed it just now.

HunterBarclay (approved):
Looks great. Found a lot of little bugs with this one, so that's always good too.

PR #982, opened 2024-06-19
Refresh APS Token

PepperLola:
Changes:
- Refresh the APS auth data when requested after it's expired
- Move code challenge generation and verifier to server

Testing:
- Sign in with APS and make sure your name and pfp show up in the main HUD
- Refresh token and see if name/pfp disappear and then reappear
- Expire token (might have to click multiple times?) and see if same thing happens as refresh token

> [!IMPORTANT]
> The server has to be running when testing, which can be found in the website repo.

> [!IMPORTANT]
> - [x] ~~Waiting on server update PR to merge~~

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1699)


fission/src/aps/APS.ts
```diff
@@ -116,14 +125,38 @@ class APS {
         }
     }
 
-    static async convertAuthToken(code: string, codeVerifier: string) {
+    static async refreshAuthToken(refresh_token: string) {
+        let now
```
HunterBarclay: This was added by me because I'm a *fantastic* developer. Could you remove the time check and instead replace with an async mutex?

fission/src/aps/APS.ts
```diff
@@ -116,14 +125,38 @@ class APS {
...
+                headers: {
+                    'Content-Type': 'application/x-www-form-urlencoded'
+                },
+                body: new URLSearchParams({
+                    'client_id': CLIENT_ID,
+                    'grant_type': 'refresh_token',
+                    'refresh_token': refresh_token,
+                    'scope': 'data:read',
+                })
+            }).then(res => res.json());
```
HunterBarclay: Please check for error on refresh and handle accordingly. You should redirect the user to login again. In which case, you can also provide the email automatically as an autofill using [this](https://aps.autodesk.com/en/docs/oauth/v2/developers_guide/authoptions/). I did look into the refresh token, and the only thing I've found is a [stackoverflow post from 2017](https://stackoverflow.com/questions/43680794/does-the-refresh-token-expire/43681411#43681411) indicating they are "extremely long lived." Other than that, we just have to assume it can eventually expire.

fission/src/aps/APS.ts
```diff
@@ -34,7 +34,7 @@ class APS {
     static get auth(): APSAuth | undefined {
```
HunterBarclay: Remove the original getter. Lets force it to check for expiration every time.

HunterBarclay (changes requested):
Couple of things to tweak. Consult with [this documentation](https://aps.autodesk.com/en/docs/oauth/v2/developers_guide/error_handling/) for addressing errors.

fission/src/aps/APS.ts
HunterBarclay: Actually just make this private

HunterBarclay:
image

I'm getting a number of errors. It's trying to use APS.auth but that is private.

BrandonPacewic (commented):
I managed to get the token to expire in such a way that resulted in a failure to login. This then followed a toast and this error:
image

Upon which the APS Login button seemed to be broken and would not let me login again.

Also there appears to be lots of error messages in the developer console while running this branch, things related to message ports closing before a expected response? Is this expected?

Edit - Steps to Reproduce:
- Spam the expire token button.

Not sure that this issue would ever be reproducible in production, just want to make sure that this is not a symptom of a different problem.

BrandonPacewic (approved):
See previous review comment, lets not let this get in the way anymore. It has been 3 weeks...

fission/src/Window.d.ts
BrandonPacewic, line 1: Almost did not catch this. Potential merge conflict from #991?

BrandonPacewic (changes requested):
Sorry to backtrack on previous review.

fission/src/Window.d.ts
PepperLola: Is `window.getTheme` called anywhere in #991? I think that was a leftover change from before he switched to using the hook from ThemeContext and deleting that file shouldn't affect anything.

Dhruv-0-Arora: You can delete Window.d.ts
Implementation of getting theme was changed so the window.getTheme function wasn't used

BrandonPacewic (approved):

HunterBarclay:
@PepperLola I'm having issues with the code challenge endpoint. For some reason it's giving a 500 error

HunterBarclay (approved):
Looking good. Can't wait to see the further integration.

PR #983, opened 2024-06-20
Haverty/elevator bug

LucaHaverty:
### Description
Fixed strange elevator behavior by adding velocity control. Elevator stages can now be controlled independently with much greater accuracy. Also reduces elevator stages moving each other and arms bouncing when reaching a position limit.


HunterBarclay (commented):
Looks good. We'll wait for #977

a-crowell (commented):
Waiting on review? The last commit was my approval for both input manager and this so probably good to go, just merge first.

HunterBarclay (changes requested):
It seems to move pretty well. The major thing for me is how much the elevator pieces get tossed around. When moving the carriage, it shifts the second stage of the elevator a ton. Same with the arm movement. Is there something we can do to essentially add movement braking?

HunterBarclay (approved):
Perfect. Great work

a-crowell (approved):
Woah this is so much better! All from weakening it a bit. Nice work.

HunterBarclay:
@LucaHaverty Could you resolve the merge conflicts, then we can merge this in

PR #986, opened 2024-06-21
Advanced Inputs

LucaHaverty:
### Description
More advanced input functionality such as controller support, axes instead of just buttons, input saving, and different controls for different robots. Each additional robot that's spawned will be given a unique control scheme, allowing them to all be controlled at once by different people. Inputs can be bound to either controller or keyboard through the change inputs modal.

### Limitations
- The UI panel is very rough at the moment. Although it doesn't look great, it should function correctly and any bugs should be reported
- At the moment, there is only support for one controller at a time
- Input saving has been pushed back to **sprint 2** when the preference manager is developed. Inputs are not supposed to save in the pr

### Objectives
- [x] Controller support
- [x] UI for assigning axes (both with joystick and two keyboard buttons)
- [x] Store controller and keyboard controls separately
- [x] Different controls for multiple robots spawned at once
- [x] Five unique preset control schemes

~~Robot specific input saving~~

## Features

When robots are spawning in, they will be automatically bound to a new control scheme. For example, if three robots are spawned, the first will be controlled by WASD, the second by arrow keys, and the third by a controller. These control schemes can be modified, reset, or bound to different robots via the controls modal.

To reset to a default control scheme, select the scheme then press the 'set' button as shows below

image

Toggling "Use Controller" for a robot will switch the modal to controller mode, allowing users to bind an input axis to either a joystick axis (like "left x") or two gamepad buttons. Joystick axes can be inverted in case a joint or robot moves backwards.

image

### Testing Done
- Spawning more that five robots (after five, they spawn with  unbound controls)
- Removing and respawning robots
- Rebinding keys then switching another robot to the same scheme
- Controlling many robots at once

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1658)


BrandonPacewic (changes requested):
This is a lot, and very impressive.

Couple of suggestions that could be changed that I would like feedback on:

- I assume we don't support binding a mouse click event because we use that for other things in the simulator that we can't currently rebind?
  - Perhaps we should revisit this once the preference manager gets implemented?
- I think the default presets could be set up a bit better from a UX standpoint. I was a little confused at first how it worked.
  - Perhaps the text should be `"Default Presets"` rather than `"Reset to Defaults"`.
  - We could then also add an additional `"Reset to Defaults"` button that would just restore default presets across all robots.
- Currently the `Cancel` and `Accept` buttons don't seem to do anything other than close the panel. I expected that I would need to apply my changes and the cancel button would avoid making any changes.
  - Perhaps this is something that we would change later once we have proper control saving?
- When selecting the `Arrow Keys` preset it assigns joint(s) `(-)` to be on `Alt` & another key. 
  - Mac should have this bound to `Command` or `Option` & another key if possible as the standard Mac keyboard does not have an `Alt` key.

Just some things that I was thinking about. Feel free to tell me whether or not these are within the scope of this PR or not.


I did not find any bugs, everything worked flawlessly.

> [!note]
> I did not test anything with a controller.

LucaHaverty:
> This is a lot, and very impressive.
> 
> Couple of suggestions that could be changed that I would like feedback on:
> 
> * I assume we don't support binding a mouse click event because we use that for other things in the simulator that we can't currently rebind?
>   
>   * Perhaps we should revisit this once the preference manager gets implemented?
> * I think the default presets could be set up a bit better from a UX standpoint. I was a little confused at first how it worked.
>   
>   * Perhaps the text should be `"Default Presets"` rather than `"Reset to Defaults"`.
>   * We could then also add an additional `"Reset to Defaults"` button that would just restore default presets across all robots.
> * Currently the `Cancel` and `Accept` buttons don't seem to do anything other than close the panel. I expected that I would need to apply my changes and the cancel button would avoid making any changes.
>   
>   * Perhaps this is something that we would change later once we have proper control saving?
> * When selecting the `Arrow Keys` preset it assigns joint(s) `(-)` to be on `Alt` & another key.
>   
>   * Mac should have this bound to `Command` or `Option` & another key if possible as the standard Mac keyboard does not have an `Alt` key.
> 
> Just some things that I was thinking about. Feel free to tell me whether or not these are within the scope of this PR or not.
> 
> I did not find any bugs, everything worked flawlessly.
> 
> Note
> 
> I did not test anything with a controller.

@BrandonPacewic 

I agree that revisiting mouse controls for robots is something to consider after preferences. Also, the save/cancel buttons do the same thing for now. They will be set up to work properly once robot inputs are integrated into the preference manager during sprint 2.

I'm almost done adding your suggestions about UX and presets.

BrandonPacewic (commented):
Is it too much to also ask for a confirm reset to default popup? 🙃

BrandonPacewic (approved):
I think we are having some problems with the linter in fission right now, should be investigated.

Fantastic work, super impressed.

> [!note]
> I did not do any testing with a controller.

Edit:
- I pushed some changes that were overriding #1000 in dev.

a-crowell (commented):
This is so cool wow!

The only thing I want to check in about is handling different controller inputs when they are hardcoded. I'm testing with a generic USB controller and finding various differences: Only the left joystick works when selected as an axis, but the right still controls a joint in a particular direction. The sticks and back/start are switched as well, but that is a nonissue since it's just naming. 

I don't think it would be reasonable it would be to handle cases from different manufacturers though. I don't use a controller much--do you have any ideas? Is there an input redirector we can suggest to users or is there a way to tell them what kind of controller will work?

LucaHaverty:
> This is so cool wow!
> 
> The only thing I want to check in about is handling different controller inputs when they are hardcoded. I'm testing with a generic USB controller and finding various differences: Only the left joystick works when selected as an axis, but the right still controls a joint in a particular direction. The sticks and back/start are switched as well, but that is a nonissue since it's just naming.
> 
> I don't think it would be reasonable it would be to handle cases from different manufacturers though. I don't use a controller much--do you have any ideas? Is there an input redirector we can suggest to users or is there a way to tell them what kind of controller will work?

@a-crowell For the MVP it may be worth just having xbox controller support and prioritizing other tasks then adding support for different controls in later weeks. @HunterBarclay what do you think?

a-crowell (approved):
Awesome. While that's getting planned, current code looks great!

HunterBarclay:
@LucaHaverty The XBox layout/scheme is the most common from what I've seen among usb controllers. I believe we're fine for now. Great work!

PR #989, opened 2024-06-24
Adaptive Physics Time Step

HunterBarclay:
### Description
Modulate the physics time step to no longer be fixed, and instead slightly adjust it towards the actual DeltaT. During the original Spike of JoltPhysics, we discovered the significant impact that adjusting the physics delta can have. Ideally we'd have a fixed, consistent time step; however, this is proving difficult as we would like to accommodate the varying performance levels.

> [!WARNING]
> This is a bandage. We may want to look into better ways of smoothing out the physics step.
> We should also investigate allowing advance options to manual set/configure physics time steps.

### Adjustment Method
We initially set a physics step rate of 120FPS. We then calculate the difference between the last used deltaT and the current frames recorded deltaT. This change is capped out to a constant and the targeted deltaT is then adjusted by this capped difference. We then calculate number of sub steps via the inverse of this deltaT. This way we can try to make up for lack of frames per second with more physical calculation.

> [!IMPORTANT]
> All of these values are limited with constants that can be found in [`PhysicsSystem.ts`](/fission/src/systems/physics/PhysicsSystem.ts)

### Downsides
Sub steps can't seem to be calculated to a good value that doesn't flip between two adjacent numbers, at higher frame rates. In addition, this inconsistent deltaT generally decays the fidelity of the simulation.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1716)


LucaHaverty (approved):
Worked well on my machine at various frame rates. Nice work!

HunterBarclay:
@LucaHaverty @PepperLola Fixed prettier and lint validation and fixing. Updated workflow to check them separately.

LucaHaverty (commented):
Seems to work at around 60fps, but when it drops to 30 I noticed joints flying apart and robots falling through the ground

image

HunterBarclay:
@LucaHaverty @PepperLola It looks like the number of substeps is decreasing as the frame rate drops which isn't right. I'll take a look.

HunterBarclay:
K, I was actually shocked at the fidelity with this last change. I couldn't see any significant joint drift
@LucaHaverty @PepperLola

PepperLola (approved):
Performed a lot better after the substep changes.

PR #990, opened 2024-06-24
God Mode: Physics

azaleacolburn:
### Description
Created a function to create and attach a "ghostBody" to some other body using a SwingTwist (Ball and Socket) constraint. The ghostBody's purpose is to control the original body, changing the position of the ghostBody will "swing" the original body along with it. The intention of this function is to be called when a user in "god mode" clicks on an object.

### Objectives
- [x] Create a ghostBody function
-  [x] Add a button to the main HUD labeled "Test God Mode," to test this ghostBody function on an arbitrary robot (at the moment, the first robot in the sceneObjects list). When clicked, the button creates and then attaches a ghost body to the robot, then repositions the it in order to move the robot.

### Testing Done
- [x] Very basic unit testing to ensure the ghostBody and constraint are actually created 
- [x] Used the aforementioned testing button to check behavior.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1675)


fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -719,6 +721,45 @@ class PhysicsSystem extends WorldSystem {
...
+        const bodyPosition = body.GetPosition()
+        const ghostPosition = new THREE.Vector3(bodyPosition.GetX(), bodyPosition.GetY() + 0.05, bodyPosition.GetZ())
+        const ghostBody = this.CreateBox(new THREE.Vector3(0.05, 0.05, 0.05), undefined, ghostPosition, undefined)
+
+        const ghostBodyId = ghostBody.GetID()
+        this._joltBodyInterface.SetObjectLayer(ghostBodyId, LAYER_GHOST)
+        this._joltBodyInterface.AddBody(ghostBodyId, JOLT.EActivation_Activate)
+        this._bodies.push(ghostBodyId)
+
+        const constraintSettings = new JOLT.SwingTwistConstraintSettings()
```
HunterBarclay: I did some more in depth reading on the SwingTwist constraint, and I'm not sure it is the correct joint we're looking for. It technically removes 3 DOF and I believe limits 1 or 2 more. Luckily, I found the Point constraint which actually only removes the 3 translational DOF

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -719,6 +721,45 @@ class PhysicsSystem extends WorldSystem {
...
+        const constraint = constraintSettings.Create(ghostBody, body)
```
HunterBarclay, line 752: You don't specify the anchor points for the constraint and it appears to snap to an arbitrary position. We want to make sure that this position is both in compliance with their current configuration (aka. The robot shouldn't be snapping to anywhere, it should just restrict further movement) and easily configurable for when god mode is implemented. It doesn't appear to be determined by the position of the ghost body at all, as I removed it and set the position to zero and it changed very little.

HunterBarclay (changes requested):
It seems like the joint is creating some odd behavior, especially as the joints kind of fail to get into compliance with any urgency. See specifics below

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -719,6 +721,45 @@ class PhysicsSystem extends WorldSystem {
...
+    /**
+     * Creates a ghost object and a distance constraint that connects it to the given body
+     * The ghost body is part of the LAYER_GHOST which doesn't interact with any other layer
+     * The caller is responsible for cleaning up the ghost body and the constraint
+     *
+     * @param id The id of the body to be attatched to and moved
+     * @returns The ghost body and the constraint
+     */
+
+    public CreateGodModeBody(id: Jolt.BodyID): [Jolt.Body, Jolt.Constraint] {
```
HunterBarclay:
```suggestion
    public CreateGodModeBody(id: Jolt.BodyID, anchorPoint: Jolt.Vec3): [Jolt.Body, Jolt.Constraint] {
```

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -719,6 +721,45 @@ class PhysicsSystem extends WorldSystem {
...
+        const body = this.GetBody(id)
+        const bodyPosition = body.GetPosition()
+        const ghostPosition = new THREE.Vector3(bodyPosition.GetX(), bodyPosition.GetY(), bodyPosition.GetZ())
+        const ghostBody = this.CreateBox(new THREE.Vector3(0.05, 0.05, 0.05), undefined, ghostPosition, undefined)
```
HunterBarclay:
```suggestion
        const ghostBody = this.CreateBox(new THREE.Vector3(0.05, 0.05, 0.05), undefined, JoltVec3_ThreeVector3(anchor), undefined)
```
Or something like that

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -719,6 +721,45 @@ class PhysicsSystem extends WorldSystem {
...
+
+        const ghostBodyId = ghostBody.GetID()
+        this._joltBodyInterface.SetObjectLayer(ghostBodyId, LAYER_GHOST)
+        this._joltBodyInterface.AddBody(ghostBodyId, JOLT.EActivation_Activate)
+        this._bodies.push(ghostBodyId)
+
+        const constraintSettings = new JOLT.PointConstraintSettings()
+        const constraint = constraintSettings.Create(ghostBody, body)
```
HunterBarclay, line 752:
```suggestion
        constraintSettings.set_mPoint1(anchorPoint)
        constraintSettings.set_mPoint2(anchorPoint)
        const constraint = constraintSettings.Create(ghostBody, body)
```
You may want to explicitly state Constraint Space for readability

fission/src/ui/components/MainHUD.tsx
```diff
@@ -159,4 +163,29 @@ const MainHUD: React.FC = () => {
...
+            return y
+        })
+        .map(x => x[1])[0] as MirabufSceneObject
+    const rootNodeId = robot.GetRootNodeId()
+    if (rootNodeId == undefined) {
+        console.error("Robot root node not found for god mode")
+        return
+    }
+    const robotPosition = World.PhysicsSystem.GetBody(rootNodeId).GetPosition()
+    const [ghostBody, _ghostConstraint] = World.PhysicsSystem.CreateGodModeBody(rootNodeId)
```
HunterBarclay:
```suggestion
    const [ghostBody, _ghostConstraint] = World.PhysicsSystem.CreateGodModeBody(rootNodeId, new JOLT.Vec3(0, 2, 0))
```
Adding the placeholder anchor point

HunterBarclay (changes requested):
Suggestions for adding controllable anchor point along with ghost body

fission/src/ui/components/MainHUD.tsx
azaleacolburn: Shouldn't the anchor point be placed right on the node?

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: The anchor point is shared because it's a point constraint, right?

HunterBarclay:
@Ewie21 Could you make sure you specify the mPoint1 and mPoint2 of the point constraint to be the anchor point. When driving the robot off the platform and testing the godmode, the anchor point is still defaulted to the origin of the world since the points are not set.

fission/src/systems/physics/PhysicsSystem.ts
HunterBarclay: @Ewie21 These are the mPoint1 and mPoint2 changes

HunterBarclay:
@Ewie21  It looks like one of the unit tests is failing; however, if  I run it separately, it works. I think we may be hitting the memory limit. I'm going to try and see if I can find some more information

a-crowell (commented):
So cool!

From Autodesk/synthesis : colbura/1675 now? It looks a commit ahead.

@HunterBarclay what are you looking for? Ghost body in the same place as the robot's initial position and worrying about gravity later?

HunterBarclay:
@a-crowell We'll worry about it later. My raycasting code will supply the initial contact point and the ghost body will act as the maneuver node for moving the connected assembly around.

HunterBarclay (approved):
Great work! We can ignore the issues with unit testing for now. They are being addressed in #996

HunterBarclay:
@Ewie21 Could you fix the merge conflict and update your fork branch to the `Autodesk/synthesis` branch?

a-crowell (approved):
Only a quick double check after the merge then. Awesome job!

PR #992, opened 2024-06-25
Mirabuf Caching

a-crowell:
### Description
Utilizing browser storage for mirabuf caching: local storage for a map of keys and metadata, and OPFS for storing mirabuf data

### Objectives
- [X] Cache from fetchLocation or ArrayBuffer (if not already in)
- [X] Get assembly from ID
- [X] Remove specific entry from local storage and OPFS
- [X] Clear storage
- [X] SpawningModal separates mirabuf files in cache vs remote as well as uses the names of the files.

### Testing
Two new temp buttons on the MainHUD:
- Print Mira Maps -> prints the maps of robot/field fetch locations and ids in console
- Clear Mira -> clears both these maps stored in local storage and any mirabuf files stored in OPFS

Other than making sure robots and fields import correctly (no user difference from before), you can also check:
- Pulling from cache when it can and remote when it needs to (via console "Mira in cache" vs "Caching new mira")
- Saving correctly (via "Print Mira Maps" for the local storage maps and an OPFS explorer for the mirabuf files)
>                     

> [!Warning]
> Clear browser cache completely before testing. Storage structure is different now, so past data is breaking.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1667)


fission/src/Synthesis.tsx
```diff
@@ -95,6 +95,11 @@ function Synthesis() {
                 .catch(_ => LoadMirabufRemote(DEFAULT_MIRA_PATH))
                 .catch(console.error)
 
+			const miraAssembly = await LoadMirabufRemote(mira_path, MiraType.ROBOT)
+				.catch(
+					_ => LoadMirabufRemote(DEFAULT_MIRA_PATH, MiraType.ROBOT)
+				).catch(console.error);
```
BrandonPacewic: There seems to be a left over merge conflict here. Duplicate definition of `miraAssembly`.

fission/src/ui/modals/spawning/SpawningModals.tsx
```diff
@@ -7,7 +7,8 @@ import { useModalControlContext } from "@/ui/ModalContext"
 import Label, { LabelSize } from "@/components/Label"
 import { CreateMirabufFromUrl } from "@/mirabuf/MirabufSceneObject"
 import World from "@/systems/World"
-import { useTooltipControlContext } from "@/ui/TooltipContext"
+import { useTooltipControlContext } from "@/TooltipContext"
```
BrandonPacewic:
```suggestion
import { useTooltipControlContext } from "@/ui/TooltipContext"
```

Sneaky merge conflict.

BrandonPacewic (commented):
Couple of fixes after the merge `dev` commit that I had to make to run the branch.

HunterBarclay:
@BrandonPacewic Could you commit those fixes?

BrandonPacewic:
> @BrandonPacewic Could you commit those fixes?

@HunterBarclay, committed.

PepperLola (changes requested):
This looks good and seems like it works well. The only nitpicks I have are to add some documentation to the UnzipMira function (what do the magic numbers mean?) and then to use `undefined` instead of `null` per the [official TypeScript style guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined).

a-crowell (commented):
@PepperLola Done on the undefined. I didn't write the UnzipMira though. @HunterBarclay do you know how it works?

BrandonPacewic (commented):
Is this the correct behaviour, I have a robot and field in the scene. Robot data looks correct to me. For some reason the field data includes the blob of the local host? Is this correct or am I misstesting?

image

a-crowell:
@BrandonPacewic I have never seen that before. 1. What were the steps you took before then and 2. When was the last time you cleared browser history?

BrandonPacewic:
> @BrandonPacewic I have never seen that before. 1. What were the steps you took before then and 2. When was the last time you cleared browser history?

@a-crowell 

Steps:
- Loaded synthesis normally
- Printed the mira maps (normal behaviour)
- Spawned the field
- Printed the maps again (detailed behaviour)

I don't think I have ever cleared my browser history on this computer, what affects are we expecting / accounting on from that?

a-crowell:
> Steps:
> 
> * Loaded synthesis normally
> * Printed the mira maps (normal behaviour)
> * Spawned the field
> * Printed the maps again (detailed behaviour)
> 
> I don't think I have ever cleared my browser history on this computer, what affects are we expecting / accounting on from that?

@BrandonPacewic 

If the cache hadn't been cleared, the blob could have been leftover from an old review. But the maps printed empty just a few seconds earlier?

a-crowell:
Out of ideas for the localhost blob in local storage and unable to recreate it on any laptops (mine, Brandon's, Dhruv's) even checking for first time running this code, so leaving be for now.

HunterBarclay:
@BrandonPacewic I used a `URL.createObjectURL()` for the `ImportLocalMirabufModal` to convert the blob I get from the file input to a downloadable URL. We should probably customize the key for those, maybe a hash of the blob?

HunterBarclay:
@a-crowell ^^^

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -1,26 +1,125 @@
 import { mirabuf } from "../proto/mirabuf"
 import Pako from "pako"
-// import * as fs from "fs"
+
+const robots = "Robots"
+const fields = "Fields"
```
BrandonPacewic: I would change these variable names to indicate what they are doing. Like `robotsDir` or something like that. If there not doing anything special and it would just be `robotText` then we could probably just leave them as string literals.

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -1,26 +1,125 @@
...
+    }
+    for await (const key of fieldFolderHandle.keys()) {
+        fieldFolderHandle.removeEntry(key)
+    }
+
+    window.localStorage.removeItem(robots)
+    window.localStorage.removeItem(fields)
+}
+
+export function GetMap(type: MiraType): any {
```
BrandonPacewic: Formatter getting upset at `any` here. Is this something we can change?

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -1,26 +1,125 @@
...
+
+        // Local cache map
+        let map: { [k: string]: string } = GetMap(type) ?? {}
+        map[fetchLocation] = backupID
+        window.localStorage.setItem(type == MiraType.ROBOT ? robots : fields, JSON.stringify(map))
+
+        console.log(assembly)
+        return assembly
+    } catch (e) {
+        console.error("Failed to load and cache mira")
```
BrandonPacewic: I would definitely suggest also console logging the error and returning `undefined`

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -1,26 +1,125 @@
...
+
+        // Get assembly from file
+        if (fileHandle != undefined) {
+            const buff = await fileHandle.getFile().then(x => x.arrayBuffer())
+            const assembly = mirabuf.Assembly.decode(UnzipMira(new Uint8Array(buff)))
+            console.log(assembly)
+            return assembly
+        }
+    } catch (e) {
+        console.error("Failed to find file from OPFS")
```
BrandonPacewic: Same as previous except this time just log the error.

fission/src/test/MirabufParser.test.ts
```diff
@@ -2,12 +2,11 @@ import { describe, test, expect } from "vitest"
 
 import { mirabuf } from "../proto/mirabuf"
 import MirabufParser, { RigidNodeReadOnly } from "../mirabuf/MirabufParser"
-import { LoadMirabufRemote } from "@/mirabuf/MirabufLoader"
-// import { LoadMirabufLocal } from "../mirabuf/MirabufLoader"
+import { LoadMirabufRemote, MiraType } from "../mirabuf/MirabufLoader"
```
BrandonPacewic: Reason for this not being imported like it was previously with `@`? Not sure why these are all over the place right now but this is not the only case so perhaps there is something that I am missing?

fission/src/systems/simulation/behavior/GenericArmBehavior.ts
```diff
@@ -11,6 +11,8 @@ class GenericArmBehavior extends Behavior {
     
     private _rotationalSpeed = 6;
 
+    private _rotationalSpeed = 30
+
```
BrandonPacewic: Not sure where this change came from, perhaps another merge error?

BrandonPacewic (changes requested):
I'm impressed you were able to figure this out, very weird thing to get working correctly.

fission/src/test/MirabufParser.test.ts
PepperLola: I prefer using the @ path aliases because it means if we were to refactor and move mirabuf stuff somewhere else (or something like that) we would only have to change where the path alias points to rather than changing all the imports.

fission/src/mirabuf/MirabufLoader.ts
PepperLola: I would create a type for the map. Something like `type MiraCache = { [path: string]: string }`, that way `GetMap` could return `MiraCache | undefined`. I also think `GetMap` could have a more descriptive name like `GetCache` because it's not immediately obvious what it does without reading the code.

PepperLola:
What's the purpose of hashing the mira data?

a-crowell:
@PepperLola The caching usually compares the url/fetchLocation to determine whether or not the file is in the cache already. The ImportLocalMiraModal uses createObjectURL to pass in a fetchLocation though, giving keys of "blob: ... localhost ...", so instead, we now use the hash as a key.

PepperLola:
> @PepperLola The caching usually compares the url/fetchLocation to determine whether or not the file is in the cache already. The ImportLocalMiraModal uses createObjectURL to pass in a fetchLocation though, giving keys of "blob: ... localhost ...", so instead, we now use the hash as a key.

If you have to load, unzip, and decode the mira to hash it before checking if it's already in the cache, wouldn't that defeat the point of caching it in the first place? Or is there something I'm missing?

a-crowell:
> If you have to load, unzip, and decode the mira to hash it before checking if it's already in the cache, wouldn't that defeat the point of caching it in the first place? Or is there something I'm missing?

You're right what is the point of this.

In case Hunter meant hash of the url, I tried that too, but the whole thing is that it's different every time so it doesn't help either.

The two options I see are to either use the file name or just accept a new file from that modal and not cache it. @HunterBarclay You were against cutting down the fetchLocation to the name for other files--do your reasons still apply here?

Is there another way to check for same files I'm not thinking of?

BrandonPacewic (commented):
Thanks for making all my suggestions. It sounds like there is still some internal discussion that needs to be had on this feature so I'll wait to properly review it till after everything has been finalized.

PepperLola (changes requested):
Instead of caching the file and then getting it and decoding there in the `CacheAndGetLocal`/`Remote` functions, why not unzip and decode there or create a new function to do it that is called everywhere else instead? If the end result of caching and then getting the cache is just getting the mira assembly, and if they're both using the same file buffer, this would skip however much time it takes to (unnecessarily?) read the file back from OPFS. Obviously let me know if I'm missing something, but to me it looks like a redundant step that could be skipped to save time.

a-crowell:
@PepperLola Thanks for pointing that out! I removed the remote version since it was never used anyway and adjusted `CacheAndGetLocal` to avoid grabbing from OPFS while taking advantage of having the assembly already to cache the name.

PepperLola (approved):
Looks good

BrandonPacewic (commented):
I'm not 100% up to date on this code and I just want to make sure that @HunterBarclay 's comment was address / is no longer relevant.

> I used a `URL.createObjectURL()` for the `ImportLocalMirabufModal` to convert the blob I get from the file input to a downloadable URL. We should probably customize the key for those, maybe a hash of the blob?

a-crowell:
@BrandonPacewic Yep, all covered! That modal now uses a hash of the file as key instead of url, so you won't find another localhost in the map.

BrandonPacewic (approved):
I tried to push formatting fixes but this branch is still not passing validation for some reason. I mentioned this in another PR review.

From what I can tell looks good, will be easier to test everything when we have end to end functionality.

HunterBarclay:
Fixed the unit tests and added a few clean ups. Didn't format because it still doesn't work on my desktop.

HunterBarclay:
Merging for now. This Prettier issue is going to be a continuous problem

PR #994, opened 2024-06-25
Physics & Suspension

Dhruv-0-Arora:
### Description
Fixed frictional constants for the floor and wheels in addition to fixing suspension to prevent robot from rocking sideways while stationary. Changes to the WheelDriver.ts file include the constants used to determine the friction between the wheels and other surfaces and setting the friction. Changes to the PhysicsSystem.ts file includes the removal of the current friction implementation and adding friction to the floor. In addition, this file includes changes to the values of *mSuspensionMinLength* and *mSuspensionMaxLength* to make them a function of the wheel radius (which is not constant).

> [!WARNING]
> **Potential Issue:** change in suspension has increased the height of the robot from the floor.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1720)

### DIFF
--- a/fission/src/systems/physics/PhysicsSystem.ts
+++ b/fission/src/systems/physics/PhysicsSystem.ts
@@ -68,8 +68,10 @@ class PhysicsSystem extends WorldSystem {
             new THREE.Vector3(0.0, -2.0, 0.0),
             undefined
         )
+        ground.SetFriction(0.7);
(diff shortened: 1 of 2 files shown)


BrandonPacewic (approved):
Significantly better!

azaleacolburn (approved):
I agree with Brandon, much better!

PR #985, opened 2024-06-21
AARD-1710: Exporter auto formatter updates

BrandonPacewic:
Unfortunately not many options to choose from.

> Black is a PEP 8 compliant opinionated formatter. Black reformats entire files in place. Style configuration options are deliberately limited and rarely added.

- Increased line length to `120` for all `.py` files.
- Now sorting imports with `isort` before running `black`.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1710)

PR #996, opened 2024-06-26
Vitest Playwright

HunterBarclay:
### Description
We've been bouncing around the issue of vitest running node, when we have a web app. This PR adds playwright, to run both the development server and a headless browser for running unit tests to avoid issues of node environment compilation.

### Motivation
I need to update Jolt Physics and something in the recent releases changed, such that it uses a `new URL()` call for getting the path to a file. This causes issues because vitest in a jsdom environment transforms `new URL()` calls to using `http://` paths, and if we use the node environment, we lose access to things like the `window`.

### Physics Memory
This *should* have a new version of the Jolt Physics package that enables memory growth and maybe increases the default size.


HunterBarclay:
Waiting on #990

a-crowell (commented):
Was this meant to fix #990 vitest memory? I'm still getting the error running Body Loading 2471 that doesn't occur individually.

HunterBarclay:
> Was this meant to fix #990 vitest memory? I'm still getting the error running Body Loading 2471 that doesn't occur individually.

@a-crowell Yes, partly. We can ignore it in #990 for right now.

HunterBarclay:
@a-crowell @PepperLola @Dhruv-0-Arora Can you verify that you can run unit tests with no errors? Note an extra setup command that I'm about to add to the readme and package.json:
```
npx playwright install
```
This installs the playwright browsers.

a-crowell (approved):
Cleared!

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
```diff
@@ -402,7 +402,8 @@ def _jointOrigin(
     geometryOrOrigin = (
         (
             fusionJoint.geometryOrOriginOne
-            if fusionJoint.geometryOrOriginOne.objectType == "adsk::fusion::JointGeometry"
+            if fusionJoint.geometryOrOriginOne.objectType
+            == "adsk::fusion::JointGeometry"
```
PepperLola: I don't really like this formatting. Has this been the config for a while and we just haven't run the formatter? It seems to really want to split things into multiple lines that end up making it more verbose and harder to read. Later on it split an if with a single variable flag into 3 lines with parentheses because there was a comment. Is the line width too low?

PepperLola: Actually I just checked #985 where the formatting was fixed so this isn't an issue.

PepperLola (approved):
Looks good

PR #997, opened 2024-06-26
Format Bun Fallback

azaleacolburn:
### Description
Previously, trying to run `bun run format`, used sub commands that ran with npm and npx. People who couldn't use npm or npmx weren't able to successfully run these scripts with bun. Now, if the npm and npmx sub commands fail, they will fall back on running them with bun and bun x respectively.

### Objectives
- [x] Change scripts to fallback on bun

### Testing Done
- [x] Verified that it doesn't break anything for users who can use both npm and bun, by formatting the first commit
- [x] Run for someone who can use bun but can't use npm, to verify that it works

No JIRA Issue


HunterBarclay:
@a-crowell I think you're the only one with only bun atm. Could you try these functions out and see how it behaves on your machine?

a-crowell (approved):
Haha works great! Thanks for this, I appreciate it.

fission/prettier.config.js
```diff
@@ -9,7 +9,15 @@ const config = {
     bracketSpacing: true,
     bracketSameLine: false,
     arrowParens: "avoid",
-    printWidth: 120
+    printWidth: 120,
+    overrides: [
+        {
+            files: "*.json",
```
PepperLola: Why the override? Not complaining, I just don't understand. Is 2 spaces standard for JSON?

azaleacolburn: I was told my @BrandonPacewic that is was. I believe it's because there's so much deep nesting / indentation in json, two spaces makes it more readable.

exporter/SynthesisFusionAddin/src/Parser/ExporterOptions.py
```diff
@@ -77,7 +77,13 @@ class ModelHierarchy(Enum):
...
 @dataclass
 class ExporterOptions:
-    fileLocation: str = field(default=(os.getenv("HOME") if platform.system() == "Windows" else os.path.expanduser("~")))
+    fileLocation: str = field(
+        default=(
+            os.getenv("HOME")
+            if platform.system() == "Windows"
+            else os.path.expanduser("~")
+        )
+    )
```
BrandonPacewic: > [!warning]
> - This change will have conflicts with #985.

azaleacolburn: Should we wait for that to be merged, then rerun the formatter on this branch?

fission/prettier.config.js
BrandonPacewic: All of our `.json` files up to this point have been 2 spaces. Don't think that there is a real "standard" but it's what we have been using.

exporter/SynthesisFusionAddin/src/Parser/ExporterOptions.py
BrandonPacewic: 👍

PepperLola (changes requested):
Since Bun seems to be considerably faster than npm/npx, would it make more sense for it to be attempted before npm? That way it will default to Bun for those who have both installed (like myself).

azaleacolburn:
> Since Bun seems to be considerably faster than npm/npx, would it make more sense for it to be attempted before npm? That way it will default to Bun for those who have both installed (like myself).

Sure, but bun is less common than npm, and I would bet that the cost of failing bun then succeeding at npm is greater than the cost of using npm. I'll make the change though, but I think it'll be slower for more people.

PepperLola (approved):
Looks good.

PR #998, opened 2024-06-27
Addable Transform Gizmos

Dhruv-0-Arora:
### Description
Added function to add Transform Gizmos to any THREE.Mesh. Gizmos will stay the same size relative to the size of the object it is attached to. If Gizmos are stacked on the same object, translations will be prioritized and uniform scaling can be achieved by holding shift as the user drags on a single axis of the gizmo. To use with Jolt bodies, we would need to create a dummy mesh that will translate changes to all the components.

### Usage Example
```typescript
const box = new THREE.Mesh(new THREE.BoxGeometry(1.0, 1.0, 1.0), this.CreateToonMaterial());
box.position.set(0.0, 1, 0.0);
this._scene.add(box);
this.AddTransformGizmo(box, 'translate', 3.0);
this.AddTransformGizmo(box, 'scale', 5.0);
this.AddTransformGizmo(box, "rotate", 7.0)
```

### Items to note:
- Orbit controls were moved to SceneRenderer from Scene.tsx to prevent orbiting while using transformGizmo.
- size must be defined when gizmo is created

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1722)


BrandonPacewic:
### Tips for the code block

```
```ts
const box = new THREE.Mesh(new THREE.BoxGeometry(1.0, 1.0, 1.0), this.CreateToonMaterial());
        box.position.set(0.0, 1, 0.0);
        this._scene.add(box);
        this.AddTransformGizmo(box, 'translate', 3.0);
        this.AddTransformGizmo(box, 'scale', 5.0);
        this.AddTransformGizmo(box, "rotate", 7.0)
`‎``
```

Will appear like this:
```ts
const box = new THREE.Mesh(new THREE.BoxGeometry(1.0, 1.0, 1.0), this.CreateToonMaterial());
        box.position.set(0.0, 1, 0.0);
        this._scene.add(box);
        this.AddTransformGizmo(box, 'translate', 3.0);
        this.AddTransformGizmo(box, 'scale', 5.0);
        this.AddTransformGizmo(box, "rotate", 7.0)
```

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -137,6 +165,35 @@ class SceneRenderer extends WorldSystem {
             gradientMap: gradientMap,
         })
     }
+
+    public AddTransformGizmo(obj: THREE.Object3D, mode: "translate" | "rotate" | "scale" = 'translate', size: number = 5.0) {
```
azaleacolburn: It would be nice if you wrote some documentation, just simple stuff like what it returns, why it might early return any edge-cases where it might fail (maybe a not about the exponential scaling we talked about), etc.

HunterBarclay (changes requested):
1. I would add a button to the MainHUD that'll drop in your sample code. We can just keep it there for testing until it is in use elsewhere.
2. When rotating the gizmo, only the scaling gizmo matches the new rotation. The translation and rotation gizmos don't update.
3. When trying to click on the different gizmos, it'll highlight multiple of them depending on the situation. When clicking on the position gizmo with multiple highlighted, it only translates. However, when selecting the scale gizmo, it can click through and enable multiple.

I would look into making it so you can add one transform gizmo and it'll return a transform gizmo class. This class can be used to switch between the different translate, rotate, scale gizmos, as well as remove them after the fact. This could avoid the overlap issues and maybe simplify the presentation because having all 3 at once gets a bit crowded.

Dhruv-0-Arora:
@HunterBarclay does this fit what you wanted?

HunterBarclay (changes requested):
Pretty close, but the rotate transform gizmo doesn't actually rotate the other axes when you use it.

HunterBarclay (approved):
Approving to get the ball rolling for #1009

PR #1002, opened 2024-06-28
Additional `CODEOWNERS` subteams

BrandonPacewic:
Current behaviour for new codeowner file:

- @HunterBarclay Now auto requested for review on every change.
- Changes in `/exporter/`, `/fission/src/aps/`, `/fission/src/mirabuf/`, `/fission/src/proto/`, and `/installer/` auto request to @Autodesk/fusion.
- Changes in `fission/src/ui/components/` and `fission/**/*.css` (all `.css` files in `/fission/` auto request to @PepperLola @LucaHaverty (current "ui" framework people).
- All other changes that are not for files specified above will reach the fallback pattern of @autodesk/synthesis-devs (@HunterBarclay will still be requested).

This is not final and I would appreciate any and all name suggestions for the teams.

Currently they are:
- `fission` currently corresponds to general feature development.
- `fusion` currently corresponds to the exporter and other cross platform "integration" things.
- Currently no UI team, currently specifying @PepperLola and @LucaHaverty as the reviewers on just the reusable UI components.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,16 @@
 # This file is used to auto request reviews for a pull request
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
-* @autodesk/synthesis-devs
+/exporter/ @autodesk/fusion hunter.barclay@autodesk.com
+
+/fission/src/aps/ @autodesk/fusion hunter.barclay@autodesk.com
+/fission/src/mirabuf/ @autodesk/fusion hunter.barclay@autodesk.com
+/fission/src/proto/ @autodesk/fusion hunter.barclay@autodesk.com
+
+/fission/src/ui/components/ julian.wright@autodesk.com luca.haverty@autodesk.com hunter.barclay@autodesk.com
+/fission/**/*.css julian.wright@autodesk.com luca.haverty@autodesk.com hunter.barclay@autodesk.com
+
+/fission/ @autodesk/fission hunter.barclay@autodesk.com
+/installer/ @autodesk/fusion hunter.barclay@autodesk.com
+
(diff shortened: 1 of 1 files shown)


PepperLola:
Is our `fusion` team blocking anyone else at Autodesk from using that team name, and could that be an issue going forward?

BrandonPacewic:
> Is our `fusion` team blocking anyone else at Autodesk from using that team name, and could that be an issue going forward?

Well this would only affect the open source Autodesk organization. They won't use this team name internally because they have a internal closed-source organization called `fusion` already.

PepperLola (approved):
Looks good, now for the cool codenames...

.github/CODEOWNERS
```diff
@@ -1,4 +1,16 @@
 # This file is used to auto request reviews for a pull request
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
-* @autodesk/synthesis-devs
+/exporter/ @autodesk/fusion hunter.barclay@autodesk.com
+
+/fission/src/aps/ @autodesk/fusion hunter.barclay@autodesk.com
+/fission/src/mirabuf/ @autodesk/fusion hunter.barclay@autodesk.com
+/fission/src/proto/ @autodesk/fusion hunter.barclay@autodesk.com
+
```
HunterBarclay, line 9:
```suggestion

/fission/src/ui/ hunter.barclay@autodesk.com julian.wright@autodesk.com
```

Thoughts?

BrandonPacewic: I initially avoided doing this because 99% of all features are going to be editing some type of modal, this would make it so that all of those changes would request @PepperLola in addition to the normal pattern for system changes. 

If we want we can do that but it would mean that @PepperLola would need to review almost every PR. Or at least would be requested for it.

HunterBarclay (approved):

PR #1006, opened 2024-07-01
Packaging update

HunterBarclay:
### Description
Updating packaging to match unit testing

### DIFF
--- a/.github/workflows/FissionPackage.yml
+++ b/.github/workflows/FissionPackage.yml
@@ -22,21 +22,6 @@ jobs:
       run: |
         echo "timestamp=$(date +'%Y-%m-%dT%H-%M-%S')" >> $GITHUB_OUTPUT
 
-    - name: Cache downloaded assets
-      id: cache-assets
-      uses: actions/cache@v2
-      with:
-        path: fission/public/
-        key: ${{ runner.os }}-assets-v1
-        restore-keys: |
-          ${{ runner.os }}-assets-
-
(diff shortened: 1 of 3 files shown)


.github/workflows/FissionPackage.yml
```diff
@@ -22,9 +22,25 @@ jobs:
...
+        path: fission/public/
+        key: ${{ runner.os }}-assets-v1
+        restore-keys: |
+          ${{ runner.os }}-assets-
+
+    - name: Download assets if not cached
+      if: steps.cache-assets.outputs.cache-hit != 'true'
+      run: |
+        cd fission
+        curl -o public/assetpack.zip https://synthesis.autodesk.com/Downloadables/assetpack.zip && unzip -o public/assetpack.zip -d public/
```
azaleacolburn: I think it would be best if we used the same tools for accomplishing the same tasks across workflows, even if the results are the same (tar vs unzip). This is the tiniest nitpick though.

```suggestion
        npm run assetpack
```

azaleacolburn (approved):
I have one tiny nitpick that can be ignored if you want. But otherwise LGTG.

.github/workflows/FissionPackage.yml
HunterBarclay: I actually removed the assetpack as that will be provided by the server.

a-crowell (approved):
Looks good

PR #1007, opened 2024-07-01
Preferences System

LucaHaverty:
### Description
A system to save and load player preferences including settings from the settings modal, robot controls, robot game piece manipulation configuration, field scoring zone data, and color themes. There are default preferences that will be used if a preference is not yet configured by the user.

### Objectives
- [x] System for getting/settings preferences, including default preferences
- [x] Ability to save and load preferences to local storage
- [x] Integrate with the settings modal
- [x] Integrate with the robot controls modal
- [x] Saving and loading of robot specific preferences
- [x] Saving and loading of field specific preferences

> [!note] 
The goal of this PR is to build a system to load, modify, and save preferences, not to make the preferences themselves have an effect on gameplay. While you should be able to change like zoom sensitivity for example, it won't yet affect the actual camera movement. The only preferences that should fully work are robot controls and color themes.

### Preferences List (all checked off have been implemented AND TESTED)
- Global
   - [x] Screen mode
   - [x] Quality
   - [x] Zoom sensitivity
   - [x] Pitch sensitivity
   - [x] Yaw sensitivity
   - [x] Report analytics
   - [x] Use metric
   - [x] Render scoring zones

- Robot Specific
   - [x] Control scheme
   - [x] Intake transform
   - [x] Ejector transform and velocity 

- Field Specific
   - [x] Scoring zones (location, alliance, points, destroy GP, persistent, parent obj)
   - [x] Default robot spawn location

### Testing Done
- All controls that are checked off have been tested

### How to add a global preference

1. Add it to type GlobalPreference in PreferencesTypes.ts

```
export type GlobalPreference =
    | "ScreenMode"
    ...
    | "MyPreference"
```

2. Assign it a default value in PreferenceTypes.ts

```
 export const DefaultGlobalPreferences: { [key: string]: Object } = {
    "ScreenMode": "Windowed",
    ...
    "MyPreference":  0.5,
}
```

3. From anywhere, set the value by calling

`PreferencesSystem.setGlobalPreference<number>("MyPreference", 2)`

4. From anywhere, get the value by calling. This is guaranteed to return a value, and will return the default value if it has not yet been assigned

`PreferencesSystem.getGlobalPreference<number>("MyPreference")`

> [!Warning]
> Merge after #986

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1692)


fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -0,0 +1,125 @@
+import { DefaultFieldPreferences, DefaultGlobalPreferences, DefaultRobotPreferences, FieldPreferences, FieldPreferencesKey, GlobalPreference, RobotPreferences, RobotPreferencesKey } from "./PreferenceTypes"
+
+class PreferencesSystem {
+    private static _preferences: { [key: string]: Object }
```
HunterBarclay: The linter really doesn't like using `Object`. I'd use `object`.

fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -0,0 +1,125 @@
...
+    private static _localStorageKey = "Preferences"
+
+    /** Sets a global preference to be a value of a specific type */
+    public static setGlobalPreference<T extends Object>(key: GlobalPreference, preference: T) {
+        if (this._preferences == undefined) this.loadPreferences()
+
+        this._preferences[key] = preference
+    }
```
HunterBarclay: Could you setup some custom events for when preferences change. I would add custom event type to wrap some relevant data in, then create functions inside of preference manager to handle the boiler plate of subscribing to an event. Use the `window.addEventListener` and `window.dispatchEvent` for events. *Those may not be the correct ones, but nearly. See my poker branch for how the UI knows when the mouse is clicked*

HunterBarclay (changes requested):

LucaHaverty:
Access events from elsewhere by calling
`PreferencesSystem.addEventListener((e) => {  /* functionality here */  })`

HunterBarclay (approved):
This will definitely be shaped as it gets used more and more, but good for now

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
```diff
@@ -42,13 +50,24 @@ class SynthesisBrain extends Brain {
...
-            this.configureInputs();
-
-            SynthesisBrain._currentRobotIndex++;
+            if (SynthesisBrain.numberRobotsSpawned[assemblyName] == undefined)
+               SynthesisBrain.numberRobotsSpawned[assemblyName] = 0 
+            else
+                SynthesisBrain.numberRobotsSpawned[assemblyName]++
+
+            this._assemblyIndex = SynthesisBrain.numberRobotsSpawned[assemblyName]
+            SynthesisBrain.robotsSpawned.push(`[${this._assemblyIndex}] ${assemblyName}`)
```
PepperLola: It would probably be best if this could go in one place (like a function to create the name or something) that's used here and later on line 84 so that it's only specified in one place.

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
```diff
@@ -1,39 +1,47 @@
...
-    private static _currentRobotIndex = 0;
-
-    // Tracks how many robots are spawned for control identification
     private _assemblyName: string
+    private _assemblyIndex: number = 0
 
-    public constructor(mechanism: Mechanism, assemblyName: string) {
-        super(mechanism);
+    // Tracks the number of each specific mira file spawned
+    public static numberRobotsSpawned: { [key: string]: number } = {}
```
PepperLola, line 32: Could this and `robotsSpawned` be combined? Maybe a map from key to list of spawned robots? That might be a bit slower if we're iterating over `robotsSpawned` a lot but it would also mean the information would only be stored in one place and we don't have to maintain two separate places.

LucaHaverty: An issue I see with this is if you spawn two robots for example 'mean machine [0]' and 'mean machine [1]', then remove only 'mean machine [0]', it needs to know that the robot still around is index 1 not 0. I can't think of an easy way around this, but let me know if you do.

BrandonPacewic (approved):
Love it. Nice work!

PR #1009, opened 2024-07-02
Transform Gizmos on Mirabuf Assemblies

Dhruv-0-Arora:
### Description
This branch adds transform gizmos to the mirabuf assemblies as they spawn in. These transform gizmos currently are always on the Mirabuf assemblies as they spawn. The physics system is disabled when the user presses the transform gizmo (will implement so physics is disabled for only the one assembly and also it stays disable as long as the gizmo is attached).

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1733)


Dhruv-0-Arora:
Right now it does not set Rotation or Scale. Only position

Dhruv-0-Arora:
Completely ready for review

HunterBarclay (changes requested):
I added a quick change to your PR that also sets the bodies to be sensors when disabled, meaning they won't act on generated collision contacts.

When spawning in a field, we should restrict it to translation movement only. Not rotation. I also noticed that after finishing with the gizmo, if I try to move the camera by clicking near the center of the screen, it doesn't work, I have to click on the edge, and then it essentially snaps out of it and I can orbit the camera like normal.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -77,11 +99,36 @@ class MirabufSceneObject extends SceneObject {
...
+                    World.PhysicsSystem.EnablePhysicsForBody(this._mechanism.GetBodyByNodeId(this._mechanism.rootBody)!)
+                    return
+                } else if (InputSystem.getInput("escape")) {
+                    this.transformGizmos.removeGizmos()
+                    World.SceneRenderer.RemoveSceneObject(this.id)
+                    return
+                }
+
+                // disable physics for the body
+                World.PhysicsSystem.DisablePhysicsForBody(this._mechanism.GetBodyByNodeId(rn.id)!)
```
HunterBarclay: Having this run every frame probably isn't the best idea.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -77,11 +99,36 @@ class MirabufSceneObject extends SceneObject {
...
             })
 
+            /**
+             * transform gizmo block to update the position and rotation of the body to match the gizmos's position
+             */
+            if (this.transformGizmos.isActive) {
+                // commands to either cancel gizmo creation or confirm position
+                if (InputSystem.getInput("enter")) {
+                    this.transformGizmos.removeGizmos()
+                    World.PhysicsSystem.EnablePhysicsForBody(this._mechanism.GetBodyByNodeId(this._mechanism.rootBody)!)
```
HunterBarclay:
```suggestion
                    World.PhysicsSystem.EnablePhysicsForBody(this._mechanism.GetBodyByNodeId(rn.id)!)
```

HunterBarclay (changes requested):
Nearly there, just a few tweaks. After positioning everything, the robot doesn't interact with the gampieces and the arm of mean machine doesn't interact with anything. Fix is added.

fission/src/mirabuf/MirabufSceneObject.ts
Dhruv-0-Arora: Oh what I forgot to change that

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -77,11 +106,35 @@ class MirabufSceneObject extends SceneObject {
...
+                    World.SceneRenderer.RemoveSceneObject(this.id)
+                    return
+                }
+
+                // if the gizmo is being dragged, copy the mesh position and rotation to the Mirabuf body
+                if (this.transformGizmos.isBeingDragged()) {
+                    this.transformGizmos.UpdateMirabufPositioning(this, rn)
+                    World.PhysicsSystem.DisablePhysicsForBody(this._mechanism.GetBodyByNodeId(rn.id)!)
+                }
+            }
```
HunterBarclay: You should move this to only be done once per update, not for each node. You're running into an issue where enter and escape are only being detected in one of these loops before being discarded so running enable on all the rigid nodes doesn't work.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -37,6 +49,22 @@ class MirabufSceneObject extends SceneObject {
...
+            new THREE.Mesh(
+                new THREE.SphereGeometry(3.0),
+                new THREE.MeshBasicMaterial({ color: 0x000000, transparent: true, opacity: 0 })
+            )
+        )
+        this.transformGizmos.AddMeshToScene()
+        this.transformGizmos.CreateGizmo("translate")
+
+        // disabling physics initially for all rigid node bodies
+        this._mirabufInstance.parser.rigidNodes.forEach(rn => {
```
HunterBarclay: Please turn adding and removing the gizmos and essentially "putting the mirabuf object in move mode" into a separate function for reusability

HunterBarclay (changes requested):

PepperLola (commented):
Are the negative axis arrows supposed to be partially invisible? Also, I noticed the squares for moving along a plane disappear when the angle between them and the camera is too shallow. Not sure if there's much you can do about that though but I could be wrong.

Dhruv-0-Arora:
> Are the negative axis arrows supposed to be partially invisible? Also, I noticed the squares for moving along a plane disappear when the angle between them and the camera is too shallow. Not sure if there's much you can do about that though but I could be wrong.

Well I can't really do much about either since it is the built in transform gizmo from three.js. The negative axis arrows being partially invisible is intentional on their end

fission/src/Window.d.ts
HunterBarclay, line 1: This file was removed in dev, what's going on with this file?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -27,6 +29,17 @@ class MirabufSceneObject extends SceneObject {
     private _debugBodies: Map<string, RnDebugMeshes> | null
     private _physicsLayerReserve: LayerReserve | undefined = undefined
 
+    private _transformGizmos: TransformGizmos | undefined = undefined
+    private _deleteGizmoOnEscape: boolean = true
```
HunterBarclay, line 33: I'll let this be for now, but the entirety of transform gizmos needs a major refactor to be more adaptable. It should honestly be worked into some kind of UI component. [Jira ticket for refactoring](https://jira.autodesk.com/browse/AARD-1744)

fission/src/Window.d.ts
HunterBarclay: @PepperLola You know what's up with this file?

fission/src/ui/components/TransformGizmos.tsx
HunterBarclay, line 1: This isn't a UI component as it stands atm.

HunterBarclay (approved):
I'm approving this for now because we want it in for the MVP and it's currently blocking two critical tasks. That being said I created a new Jira ticket for Sprint 4 to address a lot of the issues I have with Transform Gizmos.

BrandonPacewic (approved):
Let's get this MVP done!

PR #1012, opened 2024-07-04
The Poker

HunterBarclay:
Add a new "Poker" panel that can be used to demo the physics raycasting. It can place markers in the scene, as well as punch other dynamic objects.

image

image

Note: This was originally an idea for a generic "mouse interaction system" that actually just abstracted itself away as I was making.


HunterBarclay:
@BrandonPacewic @PepperLola Could I get some help with this? On my desktop, Prettier is trying to reformat every single file.

Dhruv-0-Arora (approved):

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -893,4 +923,10 @@ function tryGetPerpendicular(vec: Jolt.Vec3, toCheck: Jolt.Vec3): Jolt.Vec3 | un
     ).Normalized()
 }
 
+export type RayCastHit = {
```
Dhruv-0-Arora, line 998: would it be better to move this to the top of the file?

fission/src/Synthesis.tsx
```diff
@@ -49,12 +49,13 @@ import ScoringZonesPanel from "@/panels/configuring/scoring/ScoringZonesPanel"
...
-import ImportMirabufModal from '@/modals/mirabuf/ImportMirabufModal.tsx';
-import Skybox from './ui/components/Skybox.tsx';
-import ImportLocalMirabufModal from '@/modals/mirabuf/ImportLocalMirabufModal.tsx';
+import ManageAssembliesModal from "@/modals/spawning/ManageAssembliesModal.tsx"
+import World from "@/systems/World.ts"
+import { AddRobotsModal, AddFieldsModal, SpawningModal } from "@/modals/spawning/SpawningModals.tsx"
+import ImportMirabufModal from "@/modals/mirabuf/ImportMirabufModal.tsx"
+import Skybox from "./ui/components/Skybox.tsx"
+import ImportLocalMirabufModal from "@/modals/mirabuf/ImportLocalMirabufModal.tsx"
+import PokerPanel from "./panels/PokerPanel.tsx"
```
Dhruv-0-Arora: This is currently in src/panels
All the other panels are in @/panels ----- src/ui/panels

BrandonPacewic (commented):
> Could I get some help with this? On my desktop, Prettier is trying to reformat every single file.

I have also been having problems with running prettier on my computer as well.

fission/src/Synthesis.tsx
HunterBarclay: Fun, this predates that move. Thanks for catching that

HunterBarclay:
@PepperLola All the panels are having the unique key prop error.

image

I tried for like a solid hour to remove it and couldn't figure out where the issue was.

azaleacolburn (commented):
What is the purpose of having the Accept and Cancel buttons? The punch and sphere are applied and placed with the mouse-click, and both the Cancel and Accept buttons close the panel while leaving the sphere. It might make sense to me that the cancel button would delete the sphere, or (in a different world) that the accept button would place the sphere instead of it being at the time of the mouse-click. But right now, the Cancel and Accept buttons seem to have the same effect. It also seems nearly like every modal / panel has these two buttons, so I assume it's something inherited from some parent component. Are we able to easily remove these buttons when they're redundant?

 Please correct me if I'm mistaken.

HunterBarclay:
@Ewie21 The cancel and accept buttons are enabled by default, I just kinda forgot about them.

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -658,6 +661,91 @@ class PhysicsSystem extends WorldSystem {
...
+     */
+    public RayCast(from: Jolt.Vec3, dir: Jolt.Vec3): RayCastHit | undefined {
+        const ray = new JOLT.RayCast(from, dir)
+
+        const raySettings = new JOLT.RayCastSettings()
+        const collector = new JOLT.CastRayClosestHitCollisionCollector()
+        const bp_filter = new JOLT.BroadPhaseLayerFilter()
+        const object_filter = new JOLT.ObjectLayerFilter()
+        const body_filter = new JOLT.BodyFilter() // We don't want to filter out any bodies
+        const shape_filter = new JOLT.ShapeFilter() // We don't want to filter out any shapes
```
BrandonPacewic, line 750: If you don't mind me asking, why do we not want to filter these things?

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -1,22 +1,21 @@
-import * as THREE from 'three';
-import SceneObject from './SceneObject';
-import WorldSystem from '../WorldSystem';
+import * as THREE from "three"
+import SceneObject from "./SceneObject"
+import WorldSystem from "../WorldSystem"
```
BrandonPacewic: Should this be changed to the `@` syntax?

fission/src/ui/components/Skybox.tsx
```diff
@@ -1,12 +1,12 @@
-import World from "@/systems/World";
-import { useTheme } from "../ThemeContext";
+import World from "@/systems/World"
+import { useTheme } from "../ThemeContext"
```
BrandonPacewic: Same as previous: `@` syntax?

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -658,6 +661,91 @@ class PhysicsSystem extends WorldSystem {
...
+                    this.UpdateMinMaxBounds(new JOLT.Vec3(vert), min, max)
+                    JOLT.destroy(vert)
+                }
+                const indexArr = body.triangleMesh.mesh.indices
+                for (let i = 0; i < body.triangleMesh.mesh.indices.length; i += 3) {
+                    settings.mIndexedTriangles.push_back(
+                        new JOLT.IndexedTriangle(indexArr.at(i)!, indexArr.at(i + 1)!, indexArr.at(i + 2)!, 0)
+                    )
+                }
+            }
```
azaleacolburn:
```suggestion
            const vertArr = body.triangleMesh?.mesh?.verts
            const indexArr = body.triangleMesh?.mesh?.indices
            if (!vertsArr || !indexArr) return
            for (let i = 0; i < vertArr.length; i += 3) {
                const vert = MirabufFloatArr_JoltFloat3(vertArr, i)
                settings.mTriangleVertices.push_back(vert)
                this.UpdateMinMaxBounds(new JOLT.Vec3(vert), min, max)
                JOLT.destroy(vert)
            }
            for (let i = 0; i < indexArr.length; i += 3) {
                settings.mIndexedTriangles.push_back(
                    new JOLT.IndexedTriangle(indexArr.at(i)!, indexArr.at(i + 1)!, indexArr.at(i + 2)!, 0)
                )
            }
```

fission/src/util/TypeConversions.ts
```diff
@@ -72,10 +72,18 @@ export function MirabufVector3_JoltVec3(v: mirabuf.Vector3): Jolt.Vec3 {
...
+}
+
 export function MirabufFloatArr_JoltVec3(v: number[], offsetIndex: number): Jolt.Vec3 {
     return new JOLT.Vec3(v[offsetIndex] / 100.0, v[offsetIndex + 1] / 100.0, v[offsetIndex + 2] / 100.0)
 }
 
+export function MirabufFloatArr_JoltFloat3(v: number[], offsetIndex: number): Jolt.Float3 {
+    return new JOLT.Float3(v[offsetIndex] / 100.0, v[offsetIndex + 1] / 100.0, v[offsetIndex + 2] / 100.0)
+}
+
```
BrandonPacewic, line 86: Not that this is necessarily specific to this PR but what was the reasoning for this naming convention?

azaleacolburn (changes requested):
Left some suggestions for cleaning up the function

fission/src/systems/physics/PhysicsSystem.ts
HunterBarclay: This suggestion doesn't free the allocated objects at the end.

azaleacolburn: Is this a joke about my use of the phrase "cleaning up?"

azaleacolburn: And because the two arrays don't go out of scope within this suggestions, like they previously did?

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -658,6 +661,91 @@ class PhysicsSystem extends WorldSystem {
...
+        settings.mMaxTrianglesPerLeaf = 8
+
+        settings.mTriangleVertices = new JOLT.VertexList()
+        settings.mIndexedTriangles = new JOLT.IndexedTriangleList()
+        settings.mMaterials = new JOLT.PhysicsMaterialList()
+
+        settings.mMaterials.push_back(new JOLT.PhysicsMaterial())
+
+        const min = new JOLT.Vec3(1000000.0, 1000000.0, 1000000.0)
+        const max = new JOLT.Vec3(-1000000.0, -1000000.0, -1000000.0)
```
BrandonPacewic:
```suggestion
        const min = new JOLT.Vec3(Number.POSITIVE_INFINITY, Number.POSITIVE_INFINITY, Number.POSITIVE_INFINITY)
        const max = new JOLT.Vec3(Number.NEGATIVE_INFINITY, Number.NEGATIVE_INFINITY, Number.NEGATIVE_INFINITY)

```

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Or is there an actual issue with the suggestions. If the arrays are undefined that's fine

BrandonPacewic (changes requested):
From my testing the poker marker persists even after clicking off of the poker. Is this intended behaviour? I was envisioning that it would clean up all the visuals once the modal is closed.

fission/src/ui/panels/PokerPanel.tsx
```diff
@@ -0,0 +1,99 @@
...
+            ballMesh.position.set(hitPoint.GetX(), hitPoint.GetY(), hitPoint.GetZ())
+        }
+
+        if (punch) {
+            World.PhysicsSystem.GetBody(res.data.mBodyID).AddImpulse(
+                ThreeVector3_JoltVec3(dir.normalize().multiplyScalar(punchForce)),
+                res.point
+            )
+        }
+    }
```
azaleacolburn:
```suggestion
    if (!res) return
    if (mark) {
        const ballMesh = World.SceneRenderer.CreateSphere(
            markRadius,
            World.SceneRenderer.CreateToonMaterial(0xd6564d)
        )
        World.SceneRenderer.scene.add(ballMesh)
        const hitPoint = res.point
        ballMesh.position.set(hitPoint.GetX(), hitPoint.GetY(), hitPoint.GetZ())
    }

    if (punch) {
        World.PhysicsSystem.GetBody(res.data.mBodyID).AddImpulse(
            ThreeVector3_JoltVec3(dir.normalize().multiplyScalar(punchForce)),
            res.point
        )
    }
```

fission/src/systems/physics/PhysicsSystem.ts
BrandonPacewic: On second look it appears that these are the bounds, for some reason I thought that this was for finding the max and min. Not sure why I thought that. If the purpose is to set the bound restriction to these "reasonable" numbers and not necessary infinity then feel free to dismiss this suggestion.

HunterBarclay: Infinity is fine

fission/src/util/TypeConversions.ts
HunterBarclay: The underscore just made it clearer in my mind

fission/src/systems/physics/PhysicsSystem.ts
HunterBarclay: Because we want to cast the ray on everything and see what it hits

HunterBarclay:
> From my testing the poker marker persists even after clicking off of the poker. Is this intended behaviour? I was envisioning that it would clean up all the visuals once the modal is closed.

That was the original intended behavior, but you're right, we can get rid of that

fission/src/systems/scene/SceneRenderer.ts
HunterBarclay: Yeah, whatever I have that does formatting in vscode really doesn't like it. I'll fix it

HunterBarclay:
image

azaleacolburn (approved):
LGTM

BrandonPacewic (approved):

PR #1013, opened 2024-07-04
Convert to Absolute Imports `[AARD-1737]`

BrandonPacewic:
- Moved all relative imports to be absolute to the whole project.
- Removed `general_imports.py` and all code that was running on some form of `import *`.
- Found and fixed some hanging state that existed in silent bugs.
  - Previously `cmdDefs` in `MarkingMenu.py` was not being reset, moving into an `__init__` structure made this obvious.
- Added an `__init__` file to contain the global and saved state. This state will persist through restarts of the exporter.
- And, of course, removed more lines of code than I added 🙃 

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1737)


BrandonPacewic (commented):
@HunterBarclay Looks like the formatter for the exporter broke. My best guess is that the version of `isort` that runs on linux is behaving differently than the one for mac and windows. Might just need to disable it for now.

All files in this PR are formatting correctly.

Edit:

Check #1081 for the fix.

azaleacolburn (approved):
LGTM

Dhruv-0-Arora (approved):

PR #1016, opened 2024-07-08
APS Logout

azaleacolburn:
### Description
Allows logout from your APS account in fission.

### TODO:
- [x] Write functions to logout

### Testing Done
- [x] Test logout

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1709)


HunterBarclay:
So, quick note. This ticket was actually meant for Fission, not Fusion. Regardless, both should be done so could you extend this PR into also adding proper revoking capabilities to Fission as well?

BrandonPacewic (commented):
@Ewie21 Ready for review? Looks like there are still some `TODO`s unchecked.

fission/src/aps/APS.ts
```diff
@@ -122,12 +123,43 @@ class APS {
     }
 
     /**
-     * Logs the user out by setting their auth data to undefined.
+     * Logs the user out by setting their auth data to undefined and revoking their auth token.
+     * Does nothing if the auth token couldn't be revoked
      */
     static async logout() {
+        if (!(await this.revoke_token_public())) return
```
BrandonPacewic:
```suggestion
        if (!(await this.revokeTokenPublic())) return
```

fission/src/aps/APS.ts
```diff
@@ -122,12 +123,43 @@ class APS {
...
     }
 
+    /*
+     * Revokes the users token
+     *
+     * The client should be public since we're an spa
+     * Endpoint documentation:
+     * https://aps.autodesk.com/en/docs/oauth/v2/reference/http/revoke-POST/
+     */
+    static async revoke_token_public(): Promise<boolean> {
```
BrandonPacewic:
```suggestion
    static async revokeTokenPublic(): Promise<boolean> {
```

BrandonPacewic (changes requested):
Looks good! One minor request is that we stick to camel case for function names.

fission/bun.lockb
HunterBarclay, line 1: Can we remove this. I believe it is in the git ignore in dev.

fission/src/aps/APS.ts
HunterBarclay: If the token isn't revoked, lets just drop it anyways.

HunterBarclay (changes requested):
Revoked great, couple changes.

fission/src/aps/APS.ts
```diff
@@ -122,12 +123,43 @@ class APS {
     }
 
     /**
-     * Logs the user out by setting their auth data to undefined.
+     * Logs the user out by setting their auth data to undefined and revoking their auth token.
+     * Does nothing if the auth token couldn't be revoked
```
BrandonPacewic: Should this comment line be removed now that we always set the auth token to undefined even if we can't revoke the token?

```suggestion
```

BrandonPacewic (approved):
Revoking my stale review...

PepperLola (approved):
Looks like it works well. For some reason I'm still able to get my own user info after revoking the token but nothing else works (which is good).

HunterBarclay (approved):

PR #1018, opened 2024-07-09
Codeowner Subteams Quick Fix

BrandonPacewic:
Previous attempt to add proper subteams was unsuccessful. Currently it is just assigning @HunterBarclay and @Autodesk/synthesis-devs for everything. Hopefully the pattern order change will fix this but I'm not 100% sure.


PepperLola (approved):
Only one way to find out I suppose

HunterBarclay (approved):
Lets find out

PR #1019, opened 2024-07-09
Prettier Fix

PepperLola:
### Description
Fix prettier format validation workflow.


HunterBarclay (approved):
Worked on both my macbook and my desktop

.github/workflows/FissionESLintFormat.yml
```diff
@@ -24,30 +24,14 @@ jobs:
         npm install
     - name: Linter
       id: linter-validation
+      if: ${{ always() }}
```
BrandonPacewic, line 27: Is the `if always` really necessary? We don't use it in our other workflows, would like to keep them the same if possible.

.github/workflows/FissionESLintFormat.yml
```diff
@@ -24,30 +24,14 @@ jobs:
...
-        if [ ${{ steps.linter-validation.outcome }} == "success" ]; then
-          echo "EsLint Validation Passed"
-        else
-          echo "EsLint Validation Failed"
-          exit 1
-        fi
+        npm run lint && echo "ESLint Validation Passed" || (echo "ESLint Validation Failed" && exit 1)
     - name: Prettier
       id: prettier-validation
+      if: ${{ always() }}
```
BrandonPacewic, line 33: Is the `if always` really necessary? We don't use it in our other workflows, would like to keep them the same if possible.

.github/workflows/FissionESLintFormat.yml
PepperLola: It's necessary for it to continue on error but still report errors properly. I think we want to get both ESLint and prettier errors, but when we use continue-on-error it always says the step passed. With `if always` it will still run both but the steps will fail if ESLint and prettier fail.

BrandonPacewic: Understood, as long as there is a specific reason for it.

BrandonPacewic (approved):

PR #1020, opened 2024-07-09
Quick Fix in dev

Dhruv-0-Arora:
### Description
Fixed syntax error coming from first transform gizmo branch in SceneRenderer that was causing the error with SceneObject casting in Scene.tsx
Missing function error from MainHUD in the God Mode function. Fixed in MirabufSceneObject.ts


azaleacolburn (approved):
LGTM

BrandonPacewic (approved):
I think we may be getting a little too carless with our fixing of merge conflicts. I assume that is where these errors are coming from.

Good work.

PR #1026, opened 2024-07-15
Loading Notification

HunterBarclay:
### Description
Adds a loading notification to the bottom of the screen. This can be used via the `ProgressHandle` class which communicates to the UI via events. Progress handles can then be used to update status of tasks, and show completion or error when necessary.

### Tasks
- [x] Generalized `ProgressHandle` for UI interfacing outside of a react node.
- [x] Animate cards and progress bars.
- [x] Integrate with Mirabuf assembly spawning.

### Async Loading
So I've looked into how to get the loading to stop holding the main thread, and I haven't found a good option. Web workers have horrid restrictions on shared data and I haven't been able to find anything on deprioritizing the promise that loads the mirabuf assembly.

### Style
image
image


fission/src/mirabuf/MirabufParser.ts
```diff
@@ -72,11 +73,13 @@ class MirabufParser {
...
     }
 
-    public constructor(assembly: mirabuf.Assembly) {
+    public constructor(assembly: mirabuf.Assembly, progressHandle?: ProgressHandle) {
         this._assembly = assembly
         this._errors = new Array<ParseError>()
         this._globalTransforms = new Map()
 
+        progressHandle?.Update("Parsing assembly...", 0.3)
+
```
Dhruv-0-Arora, line 82: Would it be better to move this to MirabufSceneObject instead?

fission/src/ui/components/ProgressNotification.tsx
```diff
@@ -0,0 +1,153 @@
...
+function Interp(elapse: number, progressData: ProgressData) {
+    const [value, setValue] = useState<number>(0)
+
+    useEffect(() => {
+        const update = () => {
+            const n = Math.min(1.0, Math.max(0.0, (Date.now() - progressData.lastUpdate) / elapse))
+            const v = progressData.lastValue + (progressData.currentValue - progressData.lastValue) * easeOutQuad(n)
+
+            setValue(v)
+        }
```
Dhruv-0-Arora: Could you add documentation as to what is happening here

fission/src/util/EasingFunctions.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Source: https://easings.net/#easeOutQuad
+ *
+ * @param x
+ * @returns
```
Dhruv-0-Arora: Could you update this documentation as well

Dhruv-0-Arora (approved):
Overall looks good; it would be helpful to add documentation

fission/src/mirabuf/MirabufParser.ts
HunterBarclay: The point of that update is to indicate that the assembly is being parsed, so no.

HunterBarclay: Also, the MirabufSceneObject doesn't exist at that point

LucaHaverty (approved):
Great work! Gives people something good to look at while a robot is loading instead of thinking Synthesis is frozen 👍👍👍

PR #1030, opened 2024-07-16
Scene Tags

Dhruv-0-Arora:
### Description
Creating 2D tags that will overlay on top of the scene (as they are rendered independently) and are currently used to follow around the robots on screen but they can be used for anything. Since the tags are 2D they always are facing the user.

Screenshot 2024-07-16 at 7 01 03 PM

Screenshot 2024-07-16 at 7 01 56 PM

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1763)


HunterBarclay:
Looking pretty good. Adjust the name tags to look like this:

image

fission/src/ui/components/SceneOverlayEvents.ts
```diff
@@ -0,0 +1,61 @@
...
+        this._id = nextTagId++
+
+        this.text = text
+        this.position = position ?? [0, 0]
+    }
+
+    /** Update the tag's text */
+    public Update() {
+        new SceneOverlayTagEvent(this)
+    }
```
a-crowell, line 46: This updates position too, yeah? With this being called every frame with a robot in it for every tag, and there being only one listener, could we remove the event middleman and call the update in SceneOverlay directly?

a-crowell (dismissed):
Looking good! Just be sure to remove the tag when the robot is removed.

LucaHaverty (changes requested):
Nice work! This is looking good, just a couple comments:

For me, the tags are appearing at the bottom of robots instead of above them
image

Do you think you can add a toggle in the settings modal to disable nametag rendering? To do so: 

1. Name  the preference under "GlobalPreference" in PreferenceTypes.ts and assign it a default value
2. Add it to the settings modal (you can pretty much copy any other toggleable setting) 
3. Access it elsewhere by calling:

```ts
PreferencesSystem.getGlobalPreference<boolean>("RenderSceneTags")
```

fission/src/ui/components/SceneOverlayEvents.ts
Dhruv-0-Arora: The update function allows the SceneOverlay to get the new updated tag so I can't call it from SceneOverlay directly

a-crowell: All of the tags get updated though, and this update does the entire map for each tag. If MirabufSceneObject adjusted a public map instead of sending the update event, SceneOverlay could just use the map and do one update for all of them at once.

fission/src/ui/components/SceneOverlayEvents.ts
Dhruv-0-Arora, line 1: @HunterBarclay is it possible to combine events because I have 3 events in this file that are only named different and have different keys?

HunterBarclay: Make an enum describing the purpose of the event, then go from there.

Dhruv-0-Arora: Ok, I fixed it 👍

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -151,6 +155,10 @@ class SceneRenderer extends WorldSystem {
             )
         })
 
+        // Update the tags each frame if they are enabled in preferences
+        if (PreferencesSystem.getGlobalPreference<boolean>("RenderSceneTags"))
```
HunterBarclay, line 159: So, in my mind, this shouldn't be in SceneRenderer. You should have a useEffect inside the SceneOverlay component that adds an event listener to the PreferenceSystem, and will update isDisabled accordingly.

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -219,6 +227,22 @@ class SceneRenderer extends WorldSystem {
...
+     * @returns Pixel space coordinates
+     */
+    public WorldToPixelSpace(world: THREE.Vector3): PixelSpaceCoord {
+        this._mainCamera.updateMatrixWorld()
+        const screenSpace = world.project(this._mainCamera)
+        const widthHalf = window.innerWidth / 2
+        const heightHalf = window.innerHeight / 2
+        const x = screenSpace.x * widthHalf + widthHalf
+        const y = -(screenSpace.y * heightHalf) + heightHalf
+        return [x, y]
```
HunterBarclay:
```suggestion
        return [ window.innerWidth * (screenSpace.x + 1.0) / 2.0, window.innerHeight * (1.0 - screenSpace.y) / 2.0 ]
```

Just to reduce assignments and condense it quite a bit. This isn't something that will be constantly debugged, if ever.

fission/src/ui/components/SceneOverlay.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    backgroundColor: "rgba(0, 0, 0, 0.5)",
+                    borderRadius: "8px",
+                    padding: "8px",
+                    whiteSpace: "nowrap",
+                    transform: "translate(-50%, -100%)",
+                }}
+            >
+                <h1 className="text-2xl text-white font-sans" style={{ fontWeight: "bold" }}>
+                    {x.text}
+                </h1>
```
HunterBarclay: See Label component and use that instead for consistency.

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -219,6 +227,22 @@ class SceneRenderer extends WorldSystem {
         return screenSpace.unproject(this.mainCamera)
     }
 
+    /**
+     * Convert world space coordinates to screen space coordinates
+     *
+     * @param world World space coordinates
+     * @returns Pixel space coordinates
+     */
+    public WorldToPixelSpace(world: THREE.Vector3): PixelSpaceCoord {
```
HunterBarclay, line 236: Do you want to support depth yet or leave it for later?

HunterBarclay (changes requested):
Couple changes, all around looking good tho

fission/src/systems/scene/SceneRenderer.ts
Dhruv-0-Arora: I think leave depth for later

HunterBarclay (approved):

LucaHaverty (approved):
Works well and looks great, nice work!

PR #1014, opened 2024-07-07
Friction Data Export

azaleacolburn:
### Description
Serializes the friction override for the wheels of a robot if the checkbox is clicked. Additionally this PR creates a friction lookup table for when friction isn't overrided.

The friction override is for the `dynamic_friction` and `static_friction` fields in the each [ProtoMaterial](https://www.mirabuf.dev/#mirabuf.material.PhysicalMaterial) in the exported mirabuf.

### Todo:
- [x] Export Friction Override
- [x] ~~Parse Friction Override in fission~~ (Beyond the scope of the PR)

### Testing Done
- [x] Friction correctly set in exporterOptions
- [x] ~~Friction correctly serialized and deserialized (Not sure exactly how to test this without fission allowing a friction override in fission)~~ (See above)

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1723)

PR #995, opened 2024-06-26
AARD-1685: Joint Config Panel

BrandonPacewic:
- [x] Created a new exporter tab containing all joint configuration options.
- [x] Integrated wheel options within the joint tab for `Revolute` joints.
- [x] Moved all event listeners' 'handle' dealing with updates of the joint panel into a seperate file.
- [x] Joint settings now correctly load and save from `ExporterOptions` as implemented in #981.
- Added several quality of life things for dealing with the joint panel:
  - [x] Adding a joint that is already added will now prompt for removal of selected joint before actually removing it.
  - [x] Added a cancel button to escape out of joint selection mode.
- And, of course, small refactors here and there related to these changes as a part of ongoing exporter refactors ([AARD-1719](https://jira.autodesk.com/browse/AARD-1719)).
  - I say small refactors but this one was rather large.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1685)

PR #1032, opened 2024-07-18
Core Meta Panel & Gamepiece Configuration

BrandonPacewic:
- Moved all remaining general and advanced options to be in one tab.
- Everything is now structured like #995 for easy (er) refactoring of `ConfigCommand.py`.
- Integrated #1014.
- Made some small improvements to the way auto calculating robot weight works.
  - Now a checkbox instead of a button.
  - Choosing a different unit will now update he value input in real time.
- Also refactored gamepiece config options to be structured like #995.
- Switching between `Static` and `Dynamic` export mode will toggle the visibility of the joint config tab, the gampiece config tab, and the dynamic export specific options.

I am still not 100% sure how I feel about structuring the UI in this way. Seems better than what we were doing before but I still think it can be improved upon.

I combined both these JIRA tickets to avoid any downtime in the gamepiece configuration functionality as the meta panel task would have overridden / deprecated some of the field configuration options.

### To Test

This is a rather large PR here are somethings to make sure still work:
1. Exporting a robot normally.
2. Exporting a field normally.
3. Selection of game pieces working as expected.

Most of these changes were structural so everything should look and feel like it always has.

> [!important]
> Blocked by:
> - [x] #1014.
>   - Need to move friction data into the general tab for dynamic exports.
> - [x] #1045.
>   - Types file for unit typing and conversions.

[JIRA Issue - Meta Panel](https://jira.autodesk.com/browse/AARD-1683)
[JIRA Issue - Gamepiece Config](https://jira.autodesk.com/browse/AARD-1769)


PepperLola (approved):
Everything looks good and works well. Automatically showing/hiding tabs based on static vs dynamic makes the process a lot simpler.

Maybe the auto calculate robot weight checkbox should go above the weight though? It's something the user would need to think about before the weight because changes to it impact whether or not they're able to manually input a weight.

BrandonPacewic:
> Maybe the auto calculate robot weight checkbox should go above the weight though? It's something the user would need to think about before the weight because changes to it impact whether or not they're able to manually input a weight.

I would be open to suggestions, I made it this way because I felt it was weird to separate the other checkboxes but I could be persuaded otherwise.

PepperLola:
> I would be open to suggestions, I made it this way because I felt it was weird to separate the other checkboxes but I could be persuaded otherwise.

Yeah I was considering mentioning that in my reply. I think either way works when you've gone through the process already. The only reason I mention that is because my first instinct after hearing that it was a checkbox was to look above. It didn't take long to find it so I don't think it's super important.

BrandonPacewic:
> Yeah I was considering mentioning that in my reply. I think either way works when you've gone through the process already. The only reason I mention that is because my first instinct after hearing that it was a checkbox was to look above. It didn't take long to find it so I don't think it's super important.

I would have liked to make the checkbox part of the row with the rest of the weight inputs but I had problems making a labeled checkbox. Something I can revisit later on.

exporter/SynthesisFusionAddin/src/UI/GeneralConfigTab.py
```diff
@@ -0,0 +1,334 @@
...
+    def frictionOverrideCoeff(self) -> float:
+        frictionSlider: adsk.core.FloatSliderCommandInput = self.generalOptionsTab.children.itemById(
+            "frictionCoefficient"
+        )
+        return frictionSlider.valueOne
+
+    @logFailure
+    def handleInputChanged(self, args: adsk.core.InputChangedEventArgs) -> None:
+        commandInput = args.input
+        if commandInput.id == "exportModeDropdown":
```
azaleacolburn, line 238: Since you're matching to exact values, this could be a switch statement. If the performance cost doesn't matter, disregard this comment.

azaleacolburn (approved):
The code quality here is substantially higher than some of the older code in `ConfigCommand.py`

exporter/SynthesisFusionAddin/src/UI/GeneralConfigTab.py
BrandonPacewic: Reason for me avoiding that here was that the line lengths would be an additional 4 characters longer, and this function already has some really long lines in it. But yes, you are absolutely correct.

PR #1033, opened 2024-07-18
Configuring Mirabuf Brains

Dhruv-0-Arora:
### Description
Functionality to allow users to configure SynthesisBrain and WIPLIBBrain settings for different robots

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1751)


LucaHaverty:
@Dhruv-0-Arora @HunterBarclay 
Should we make this part of the new combined configuration panel?

Dhruv-0-Arora:
> @Dhruv-0-Arora @HunterBarclay Should we make this part of the new combined configuration panel?

yes

BrandonPacewic (changes requested):
We should wait to put this in until #1068 is in since it contains parts of this PR. After that we can figure out what was left out and integrate everything else.

> [!note]
> - Waiting until #1068.

PR #1034, opened 2024-07-18
Named robot controls

LucaHaverty:
### Description
Adds names and name tags to robots to help the user distinguish between robots when configuring controls and other settings like intake/ejector. You can create completely custom input schemes, or edit the default ones. Deleting a customized version of a default input scheme will reset it, not completely remove it.

### Objectives
- [x] Refactor how inputs are stored and accessed for different robots to support named input schemes instead of only mira specific inputs
- [x] Add names to default input schemes and a list of other names to auto assign custom ones
- [x] Panel to select a name when a robot is spawned
- [x] Update intake and ejector config to work with the new system
- [x] Integrate with name tags

> [!note]
> When testing, make sure to mess around with configuring intakes and ejectors because the way robot control info is stored changed, which could potentially create issues.

image

image

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1755)

PR #1035, opened 2024-07-19
Intake/Scoring Collisions

a-crowell:
### Description
Using collision detection for intaking and scoring.

### Objectives
- [X] Intakes on contact and Q (easily changed once #1034 is merged in)
- [X] Scoring zone configuration applies to sensor and mesh: alliance, points, position/scale/rotation, and parentNode
- [X] Settings > Render Scoring Zones
- [X] Score keeping
- [X] Persistent points
- [X] Basic Score UI

### Note
- Destroying gamepieces will be looked back into once we can bring more gamepieces onto the field, so not this PR.

### Watch For
- Double counted points (not persistent zone). I have seen this every once in a while, but haven't been able to reproduce it yet.
- Changing the render scoring zone preference and refreshing page/adding zones/etc
- Zones with changed scale not rotating properly

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1731)


LucaHaverty (changes requested):
Amazing work, this is coming along well! Just a couple things I noticed:

- When you move a scoring zone then click 'accept', the rendered zones position does not update until you close the 'scoring zones' panel. This isn't a big issue, but can make it seem like moving the zone didn't work. Also, do you think you could make the rendered scoring zone disappear right away when you delete a zone?
- The persistent point option doesn't quite work for me. What happens is the points stay counted even after the gamepiece exits the zone, but when any new gamepiece is picked up, points are then subtracted. With persistent points toggled off, everything works as expected.
- I'm pretty sure this is a mistake from my PR, but the button to select a zone parent node says "select pickup node" because it was copied from the intake panel.
image

How hard would it be to add a toggle in settings to not show the score UI? It might be nice to have it out of the way for screenshots and such. This could definitely wait until after the MVP though.

Another things that would be nice is adding the scoring zone colors to the zone config gizmo, but again, this can wait.

Overall great work, I'm really impressed!

fission/src/mirabuf/ScoringZoneSceneObject.ts
```diff
@@ -0,0 +1,276 @@
...
+        shininess: 0.0,
+        opacity: 0.7,
+        transparent: true,
+    })
+    static blueMaterial = new THREE.MeshPhongMaterial({
+        color: 0x0066B3,
+        shininess: 0.0,
+        opacity: 0.7,
+        transparent: true,
+    })  //0x0000ff
```
LucaHaverty: I'm not exactly sure how we'd do this, but in the future we may want to integrate these colors into the theme editor. I'm not sure why, but someone may want to change the scoring zone colors to their team color or something.

fission/src/ui/panels/information/ScoreboardPanel.tsx
```diff
@@ -20,6 +21,13 @@ const ScoreboardPanel: React.FC<PanelPropsImpl> = ({ panelId, openLocation, side
         [setInitialTime, setTime, setStartTime]
     )
```
LucaHaverty: This is causing a build error because it's unused

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -12,11 +12,12 @@ import Mechanism from "@/systems/physics/Mechanism"
 import SynthesisBrain from "@/systems/simulation/synthesis_brain/SynthesisBrain"
 import InputSystem from "@/systems/input/InputSystem"
 import TransformGizmos from "@/ui/components/TransformGizmos"
-import { EjectorPreferences, FieldPreferences, IntakePreferences } from "@/systems/preferences/PreferenceTypes"
+import { EjectorPreferences, FieldPreferences, IntakePreferences, ScoringZonePreferences } from "@/systems/preferences/PreferenceTypes"
```
LucaHaverty: Importing ScoringZonePreferences is causing a build error because it isn't used

fission/src/mirabuf/ScoringZoneSceneObject.ts
```diff
@@ -0,0 +1,276 @@
...
+                
+                // Position/rotate/scale sensor to settings
+                this._deltaTransformation = Array_ThreeMatrix4(this._prefs.deltaTransformation)
+                const fieldTransformation = JoltMat44_ThreeMatrix4(World.PhysicsSystem.GetBody(this._parentBodyId).GetWorldTransform())
+                const zoneTransformation = this._deltaTransformation.clone().premultiply(fieldTransformation)
+
+                const translation = new THREE.Vector3(0, 0, 0)
+                const rotation = new THREE.Quaternion(0, 0, 0, 1)
+                const scale = new THREE.Vector3(1, 1, 1)
+                zoneTransformation.decompose(translation, rotation, scale)
```
LucaHaverty: I believe this same math is used two or three times, so maybe we could put it in a utils function. Not super important but would be nice if it changes down the line.

fission/src/systems/simulation/SimulationSystem.ts
```diff
@@ -49,6 +52,11 @@ class SimulationSystem extends WorldSystem {
             return false
         }
     }
+
+    public ResetScores(): void {
+        SimulationSystem.redScore = 0
+        SimulationSystem.blueScore = 0
+    }
```
LucaHaverty: Should this be static?

fission/src/mirabuf/ScoringZoneSceneObject.ts
a-crowell: We could make it work. When we have more time, let's add them.

HunterBarclay:
@a-crowell Can you update this branch to the new Input schemes?

LucaHaverty (approved):
Amazing work!!! This is very impressive 😄

LucaHaverty (changes requested):
Sorry.......just one quick build error to fix :(((

``` ts
src/ui/panels/configuring/scoring/ZoneConfigPanel.tsx:116:21 - error TS6133: 'setDestroy' is declared but its value is never read.

116     const [destroy, setDestroy] = useState<boolean>(SelectedZone.zone.destroyGamepiece)
```

Also run `npm run build` and `npm run format` after you fix this to make sure there are no more sneaky errors

LucaHaverty (approved):

fission/src/ui/panels/configuring/scoring/ScoringZonesPanel.tsx
Dhruv-0-Arora, line 1: On lines 164-166 can you make it so that when a zone is deleted then the zone itself is also actually deleted?

Dhruv-0-Arora (changes requested):

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1005,6 +1022,96 @@ class PhysicsSystem extends WorldSystem {
...
+            }
+
+            this._physicsEventQueue.push(new OnContactPersistedEvent(message))
+        }
+
+        contactListener.OnContactRemoved = subShapePairPtr => {
+            const shapePair = JOLT.wrapPointer(subShapePairPtr, JOLT.SubShapeIDPair) as Jolt.SubShapeIDPair
+
+            const event = new OnContactRemovedEvent(shapePair)
+            event.Dispatch()
```
Dhruv-0-Arora: Could you put event.dispatch inside the constructor of OnContactRemovedEvent since its not in the event queue

HunterBarclay: @a-crowell Was having this event be dispatched in the queue causing an issue? If not, it should probably be added to the queue as well.

a-crowell: Yeah having it in the queue causes issues. It seems there is just enough delay to allow removed events to go unnoticed/in the wrong order. I can put the dispatch inside the constructor though.

fission/src/ui/panels/configuring/scoring/ScoringZonesPanel.tsx
a-crowell: Zones are all deleted and repopulated after updating field preferences, so unless you are seeing a bug with it staying on the field that I am not seeing, this does delete the zone.

Dhruv-0-Arora (approved):
immaculate

PR #1042, opened 2024-07-24
Refactor: `ExporterOptions.py`

BrandonPacewic:
Not much to say here, wrote some bad code, made it better. Functionality is unchanged, aside from the fact that `readFromDesign` will no longer throw exceptions when we update `ExporterOptions` and will now default to the individual field default.

> [!important]
> - ~~Blocked by #1045.~~


azaleacolburn (approved):
Love this!

PepperLola (approved):
Looks good

PR #1010, opened 2024-07-03
AARD-1735: Logging module updates

BrandonPacewic:
- Added logging file history:
  - The past 10 log files will now be saved.
- Removed old analytics and timing system.
- Improved / added timing and performance based logging:
  - Example: Added `@timed` decorator to functions added in #981 to track performance concerns.
  - Also added timing to the overall export to show end to end runtime.
- Created a wrapper to avoid global `try` and `except`s that were all leading into `logger.error` and `ui.messageBox` error messages.

> [!important]
> - ~~Blocked by #995.~~

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1735)

PR #1048, opened 2024-07-25
APS Upload Pipeline Polish

BrandonPacewic:
- Now no longer throws a message box error if the user is not signed in.
  - This would previously prevent the event handlers from being added properly and would cause some undefined behaviour.
- Will now prompt for login on export with upload and missing APS login data instead of throwing an error message.
  - Previously would parse the whole robot before telling you that your destination was invalid.
- Removed the log that was logging the user auth token.
- Now calling the logging module from #1010.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1774)

### DIFF
--- a/exporter/SynthesisFusionAddin/src/APS/APS.py
+++ b/exporter/SynthesisFusionAddin/src/APS/APS.py
@@ -1,5 +1,4 @@
 import json
-import logging
 import os
 import pathlib
 import pickle
@@ -12,6 +11,9 @@ from typing import Any
 import requests
 
 from ..general_imports import INTERNAL_ID, gm, my_addin_path
+from ..Logging import getLogger
+
+logger = getLogger()
 
 CLIENT_ID = "GCxaewcLjsYlK8ud7Ka9AKf9dPwMR3e4GlybyfhAK2zvl3tU"
 auth_path = os.path.abspath(os.path.join(my_addin_path, "..", ".aps_auth"))
@@ -70,7 +72,6 @@ def getAuth() -> APSAuth | None:
     if os.path.exists(auth_path):
         with open(auth_path, "rb") as f:
             p: APSAuth = pickle.load(f)
-            logging.getLogger(f"{INTERNAL_ID}").info(msg=f"{json.dumps(p.__dict__)}")
             APS_AUTH = p
     else:
         return None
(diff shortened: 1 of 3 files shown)


exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Parser.py
```diff
@@ -178,9 +182,9 @@ def export(self) -> None:
             folder_id = project.rootFolder.id
             file_name = f"{self.exporterOptions.fileLocation}.mira"
             if upload_mirabuf(project_id, folder_id, file_name, assembly_out.SerializeToString()) is None:
-                gm.ui.messageBox("FAILED TO UPLOAD FILE TO APS", "ERROR")  # add throw later
-        # Download Mirabuf File
+                raise RuntimeError("Could not upload to APS")
         else:
+            assert self.exporterOptions.exportLocation == ExportLocation.DOWNLOAD
```
azaleacolburn, line 187: I forgot that enums could be undefined 🤦‍♀️

exporter/SynthesisFusionAddin/src/APS/APS.py
```diff
@@ -65,20 +65,23 @@ def getAuth() -> APSAuth | None:
     global APS_AUTH
     if APS_AUTH is not None:
         return APS_AUTH
-    try:
-        curr_time = time.time()
+
+    currTime = time.time()
+    if os.path.exists(auth_path):
```
azaleacolburn, line 72: Should it display a messagebox or smth if the path is invalid?

azaleacolburn (approved):
LGTM

exporter/SynthesisFusionAddin/src/APS/APS.py
BrandonPacewic: I'm thinking not. Just means that the user is not currently logged in / has never logged in before. But this could be up for debate.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Parser.py
BrandonPacewic: I'm more so just checking to make sure we don't add different export locations in the future. Sort of a "this was the entire behaviour of `ExportLocation` when I wrote this code so it may need to be updated" type of thing. But yea I suppose it will also catch if it is `None` as well.

exporter/SynthesisFusionAddin/src/APS/APS.py
azaleacolburn: omg I'm dumb, I was thinking of the download path oml, totally out of the scope of this PR + smth that won't ever be undefined. 🤦‍♀️

PepperLola (approved):
Looks good

PR #1049, opened 2024-07-25
Analytics

HunterBarclay:
### Description
Created a system to interact with Google Analytics, and provide avenues for tracking custom events and exceptions. A pop up will display in the bottom-right corner asking for consent to use cookies. This will show every time the application loads if the user hasn't consented. I would say this is a tactic to encourage the user consenting to allowing the use of cookies, but I just don't want to track if they say no or not.

The google analytics NPM package I am using has built-in support for ensuring consent compliance within the EU.

### Testing
You can view analytic data using the Synthesis google account on the Google Analytics site (I will actually refuse to give you creds until you setup Keeper). Debug view is enabled only on development servers.

image

image

### Current Analytic Data Points
Caching Data Points (file size information and Mirabuf assembly type):
- Remote Download
- APS Download
- Cache Store
- Cache Get

APS Data Points:
- Successful Login
- Successful Challenge Code
- Failed Login
- Failed Refresh
- \# of APS calls per minute (by endpoint)
- Associated User ID

Performance:
- Average Frame time (both total and broken up by System)

Beta:
- Associated Beta code with user

User:
- Associated user property for if they are on a mobile device or not

### Data Collection Information
I've added an information page about our data collection methods and reasonings to the website. This is like compliance x1000 so we should be in the clear to collect anything we need.

### DIFF
--- a/fission/index.html
+++ b/fission/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-6XNCRD7QNC"></script>
     <link rel="icon" type="image/svg+xml" href="/synthesis-logo.svg" />
     <link rel='stylesheet' href='https://static-dc.autodesk.net/etc/designs/v201808022224/templates-general/structure/fonts/artifakt/clientlibs/artifakt.css'/>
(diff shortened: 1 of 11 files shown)


a-crowell:
Pop up works and I don't see any problems with the code. Can we talk more about testing during standup tomorrow? Or are you planning to merge without testing?

HunterBarclay:
@PepperLola I also added a bug for APS that only presented itself once it was on production.

PR #1051, opened 2024-07-26
UI Touchups v2

LucaHaverty:
## Description
Made a new dropdown component that's similar to the input component and moved repeated style definitions into one place. These can be found in [`StyledComponents.tsx`](https://github.com/Autodesk/synthesis/blob/haverty/1775/ui-code-cleanup/fission/src/ui/components/StyledComponents.tsx).

The new dropdown component now supports a default option, which can be used as follows:

```tsx
<Dropdown
    defaultValue={"Option1"}
    options={["Option1", "Option2", "Option3", "Option4"]}
    onSelect={(val) => { }}
    label={"Dropdown Label"}
/>
```

### Tasks
- [x] Move repeated component definitions to [`StyledComponents.tsx`](https://github.com/Autodesk/synthesis/blob/haverty/1775/ui-code-cleanup/fission/src/ui/components/StyledComponents.tsx).
- [x] move all icons to [`StyledComponents.tsx`](https://github.com/Autodesk/synthesis/blob/haverty/1775/ui-code-cleanup/fission/src/ui/components/StyledComponents.tsx).
- [x] Remake the dropdown component

~~Merge brain config with manage assemblies, sequential joint config, intake config, ejector config, and possibly inputs (although inputs are cluttered enough as is). This might be big enough to be a separate PR~~
(**Moved to a separate PR**)

### Opened Dropdown
image

Feel free to request any other small UI changed you'd like on this PR!

> [!note] 
> The bug with scroll views overflowing is fixed in #1063 

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1775)


fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -1,23 +1,14 @@
...
-    }
-
-    arr = arr.includes(element) ? [element, ...arr.filter(item => item !== element)] : arr
-    return arr
-}
+import Dropdown from "@/ui/components/Dropdown"
+import { SynthesisIcons } from "@/ui/components/StyledComponents"
 
 const screenModeOptions = ["Windowed", "Fullscreen"]
 const qualitySettingsOptions = ["Low", "Medium", "High", "Ultra"]
```
Dhruv-0-Arora:
```suggestion
const qualitySettingsOptions = ["Low", "Medium", "High"]
```

fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -79,20 +70,16 @@ const SettingsModal: React.FC<ModalPropsImpl> = ({ modalId }) => {
...
-                options={moveElementToTop(
-                    screenModeOptions,
-                    PreferencesSystem.getGlobalPreference<string>("ScreenMode")
-                )}
+                defaultValue={PreferencesSystem.getGlobalPreference<string>("ScreenMode")}
+                options={screenModeOptions}
                 onSelect={selected => {
                     setScreenMode(selected)
                 }}
             />
```
Dhruv-0-Arora: Could you delete this there is no windowed vs full screen

fission/src/ui/modals/configuring/SettingsModal.tsx
Dhruv-0-Arora, line 1: - Could you center the the "Theme Editor" button
- Could you add more space above "Preferences" text
- Could you reduce the font size and space between stuff underneath "Preferences"

fission/src/ui/modals/configuring/theme-editor/ThemeEditorModal.tsx
Dhruv-0-Arora, line 1: The Theme Editor Modal seems to be a little broken. Could we add some space right above the "Theme Editor" text and remove the faint green line in the image attached?
Also the functionality of the theme editor is now broken

Screenshot 2024-07-29 at 2 00 11 PM

Dhruv-0-Arora: Screenshot 2024-07-29 at 2 02 04 PM

Dhruv-0-Arora (changes requested):

fission/src/ui/modals/configuring/SettingsModal.tsx
azaleacolburn: Graydon's rolling in his metaphorical grave right now

fission/src/ui/modals/configuring/theme-editor/ThemeEditorModal.tsx
LucaHaverty: The content extending off the modal will be fixed in #1063

Dhruv-0-Arora: ok 👍 
what about the theme editor being broken should that be its own PR?

fission/src/ui/modals/configuring/SettingsModal.tsx
LucaHaverty: @azaleacolburn is this better?

```ts
enum QualitySettingsCharacter {
    L,
    O,
    W,
    M,
    E,
    D,
    I,
    U,
    H,
    G,
}

const getSetting = (settings: QualitySettingsCharacter[]) => {
    if (
        settings[0] == QualitySettingsCharacter.L &&
        settings[1] == QualitySettingsCharacter.O &&
        settings[2] == QualitySettingsCharacter.W
    )
        return "Low"
    else if (
        settings[0] == QualitySettingsCharacter.M &&
        settings[1] == QualitySettingsCharacter.E &&
        settings[2] == QualitySettingsCharacter.D &&
        settings[3] == QualitySettingsCharacter.I &&
        settings[4] == QualitySettingsCharacter.U &&
        settings[5] == QualitySettingsCharacter.M
    )
        return "Medium"
    else if (
        settings[0] == QualitySettingsCharacter.H &&
        settings[1] == QualitySettingsCharacter.I &&
        settings[2] == QualitySettingsCharacter.G &&
        settings[3] == QualitySettingsCharacter.H
    )
        return "High"
}
```

(this is a joke btw)

azaleacolburn: You're a genius, think about it, the JIT compiler will read the enums faster, since each enum name is only one character long (this is a joke btw)

LucaHaverty: @azaleacolburn serious question time: Is there an advantage of using an enum instead of 
`type QualitySetting = "Low" | "Medium" | "High"`

azaleacolburn: Yes, but not a _ton_
- Enums can be iterated over
- Enums provide clear intent about what is happening (enums are for a list of valid options / variants, while optional types were introduced as a way of allowing variant types. The idea is that optional types should be used for type variation, rather than a list of variant values.) This is the important one for me.
- Enums are slightly faster at runtime since they're just numbers.
- Maybe my lsp is broken, but i get autocomplete on enums and not on enum-like string type variants.
- In every place where a variant is directly used, it's clearer about what the variant means with an enum, than if just a string literal is used. the string literal could mean anything, while the enum is clearly a variant of an some option. Think about the difference between seeing the literal `SimType.PWM` passed in compared to just `"PWM"`, to me, if I'm unfamiliar with the code, one is clearer.

Cons:
- Enum declaration syntax is uglier

azaleacolburn (approved):
LGTM

a-crowell (approved):
Great work! I love how organized the SynthesisIcons makes it.

fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -9,23 +9,14 @@ import Slider from "@/components/Slider"
...
-        return arr
-    }
-
-    arr = arr.includes(element) ? [element, ...arr.filter(item => item !== element)] : arr
-    return arr
-}
-
-const screenModeOptions = ["Windowed", "Fullscreen"]
-const qualitySettingsOptions = ["Low", "Medium", "High", "Ultra"]
+const qualitySettingsOptions: QualitySetting[] = ["Low", "Medium", "High"]
```
Dhruv-0-Arora: I feel like this and the QualitySettings type on line 32 of PreferenceTypes is kinda redundant. Would it be better to change it?

fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -76,91 +66,88 @@ const SettingsModal: React.FC<ModalPropsImpl> = ({ modalId }) => {
...
+                <Label size={LabelSize.Medium}>Screen Settings</Label>
+                <Dropdown
+                    label="Quality Settings"
+                    options={qualitySettingsOptions}
+                    defaultValue={PreferencesSystem.getGlobalPreference<string>("QualitySettings")}
+                    onSelect={selected => {
+                        setQualitySettings(selected)
+                    }}
+                />
+                <Box minHeight={"5px"} />
```
Dhruv-0-Arora: This might not be helpful but maybe we could add a Component that is called Spacer that just adds space
it might help for code readability

Dhruv-0-Arora (changes requested):

fission/src/ui/modals/configuring/SettingsModal.tsx
BrandonPacewic: https://github.com/Autodesk/synthesis/blob/79f677c413fa3fe0f534a11654fd55256b1ebc0a/fission/src/systems/preferences/PreferenceTypes.ts#L32

Dhruv-0-Arora (approved):
Yum 😋

PR #1053, opened 2024-07-26
Toolbar Buttons

azaleacolburn:
### Description
Add button that links to synthesis tutorials in the exporter toolbar.

### Testing Done
- [x] Click the button and go to the website

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1688)


PepperLola:
When the addon starts, I get an error saying that 'UI.Helper' has no attribute 'check_solid_open' in `Synthesis.py:138`.

If I change it to `lambda *_: True` like what the rest of the buttons have, the new button shows up and opens the website properly.

Also, could you run the formatter?

exporter/SynthesisFusionAddin/Synthesis.py
```diff
@@ -129,3 +130,14 @@ def register_ui() -> None:
     )
 
     gm.elements.append(apsButton)
+
+
+    websiteButton = HUI.HButton(
+        "Synthesis Website", 
+        work_panel,
+        Helper.check_solid_open,
```
BrandonPacewic: > When the addon starts, I get an error saying that 'UI.Helper' has no attribute 'check_solid_open' in `Synthesis.py:138`.

This was a merge conflict, left this to be changed in the HUI refactor.

```suggestion
        lambda *_: True,
```

BrandonPacewic (commented):
@azaleacolburn I would recommend running the format script in [`exporter/SynthesisFusionAddin/Synthesis.py`](https://github.com/Autodesk/synthesis/blob/f467f95d3e62fe23b319135c2ebdd7a68dca643d/exporter/SynthesisFusionAddin/tools/format.py) if you are not already 🙃

PepperLola (approved):
Works now

BrandonPacewic (changes requested):
Looks like there are still some left over formatting changes, for some reason some files are formatted to have max 80 character columns when it should be 120.

Reseting those files to their state on `dev` should do the trick.

BrandonPacewic (approved):
Just reverted the formatting changes that occurred 🙃

PR #1060, opened 2024-07-29
APS Management Modal

PepperLola:
### Description
Adds an APS modal that displays user information and includes a logout button. This opens instead of just logging out when clicking the APS button in the main HUD.

### Objectives
- [x] Show user information
- [x] Functional logout button

### Testing Done
- Make sure user information (pfp + name) displayed properly
- Make sure logout button works

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1698)

### DIFF
--- a/fission/src/Synthesis.tsx
+++ b/fission/src/Synthesis.tsx
@@ -66,6 +66,7 @@ import NewInputSchemeModal from "./ui/modals/configuring/theme-editor/NewInputSc
 import AssignNewSchemeModal from "./ui/modals/configuring/theme-editor/AssignNewSchemeModal.tsx"
 import AnalyticsConsent from "./ui/components/AnalyticsConsent.tsx"
 import PreferencesSystem from "./systems/preferences/PreferencesSystem.ts"
+import APSManagementModal from "./ui/modals/APSManagementModal.tsx"
 
 const worker = new Lazy<Worker>(() => new WPILibWSWorker())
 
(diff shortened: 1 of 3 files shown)


fission/src/Synthesis.tsx
```diff
@@ -66,6 +66,7 @@ import NewInputSchemeModal from "./ui/modals/configuring/theme-editor/NewInputSc
 import AssignNewSchemeModal from "./ui/modals/configuring/theme-editor/AssignNewSchemeModal.tsx"
 import AnalyticsConsent from "./ui/components/AnalyticsConsent.tsx"
 import PreferencesSystem from "./systems/preferences/PreferencesSystem.ts"
+import APSManagementModal from "./ui/modals/APSManagementModal.tsx"
```
BrandonPacewic, line 69: Can someone explain to me the difference between different `@` imports and seemingly relative imports? I understand that `@` refers to a module (or the javascript / typescript equivalent. Essentially a directory or collection of files), but we seem to switch between using an `@` and a `./` and I'm not familiar with why.

PepperLola: We specify what the @ ones point to in our tsconfig and vite config, and they make it easier to change imports if we were to do a refactor since we’d only need to change there (for the most part). I think we switch between them because our editors auto-import using relative imports and nobody bothers to change it. It might be worth going through everything at some point and standardizing it.

BrandonPacewic: I agree, not too big of a deal but would be nice if everything was the same.

BrandonPacewic (approved):
Can't really think of anything else to add to this panel. Definitely makes the logout much more clear. Could see how people would accidently log themselves out previously.

HunterBarclay (approved):

PR #1061, opened 2024-07-29
PWM Support

PepperLola:
### Description
Write to individual driver types when simulating with PWM instead of logging or assuming all are wheels.

I'm not sure if averaging out the output of each motor in a group is accurate to how it would behave in real life but we can adjust that later if necessary.

> [!NOTE]
> Motor max speeds are currently hardcoded but will be updated when the motor config task is done.

### Objectives
- [x] Write to drivers

### Testing Done
- Configure each type of driver to PWM group
- Run simulator and make sure the drivers move as expected

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1782)

### DIFF
--- a/fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
+++ b/fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
@@ -289,10 +289,10 @@ export class PWMGroup extends SimOutputGroup {
     }
 
     public Update(_deltaT: number) {
-        let average = 0;
+        let average = 0
         for (const port of this.ports) {
             const speed = SimPWM.GetSpeed(`${port}`) ?? 0
-            average += speed;
+            average += speed
             console.log(port, speed)
         }
         average /= this.ports.length
--- a/fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
+++ b/fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
@@ -6,6 +6,9 @@ import WPILibWSWorker from "./WPILibWSWorker?worker"
 import { SimulationLayer } from "../SimulationSystem"
 import World from "@/systems/World"
 import Driver from "../driver/Driver"
+import WheelDriver from "../driver/WheelDriver"
+import HingeDriver from "../driver/HingeDriver"
+import SliderDriver from "../driver/SliderDriver"
 
 const worker = new WPILibWSWorker()
 
@@ -286,18 +289,24 @@ export class PWMGroup extends SimOutputGroup {
     }
 
     public Update(_deltaT: number) {
-        // let average = 0;
+        let average = 0;
         for (const port of this.ports) {
             const speed = SimPWM.GetSpeed(`${port}`) ?? 0
-            // average += speed;
+            average += speed;
             console.log(port, speed)
         }
-        // average /= this.ports.length
+        average /= this.ports.length
 
-        // this.drivers.forEach(d => {
-        //     (d as WheelDriver).targetWheelSpeed = average * 40
-        //     d.Update(_deltaT)
-        // })
+        this.drivers.forEach(d => {
+            if (d instanceof WheelDriver) {
+                d.targetWheelSpeed = average * 40
+            } else if (d instanceof HingeDriver) {
+                d.targetVelocity = average * 40
+            } else if (d instanceof SliderDriver) {
+                d.targetVelocity = average * 40
+            }
+            d.Update(_deltaT)
+        })
     }
 }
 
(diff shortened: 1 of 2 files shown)


fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
```diff
@@ -286,18 +289,24 @@ export class PWMGroup extends SimOutputGroup {
...
-        // })
+        average /= this.ports.length
+
+        this.drivers.forEach(d => {
+            if (d instanceof WheelDriver) {
+                d.targetWheelSpeed = average * 40
+            } else if (d instanceof HingeDriver) {
+                d.targetVelocity = average * 40
+            } else if (d instanceof SliderDriver) {
+                d.targetVelocity = average * 40
```
Dhruv-0-Arora: Im assuming the reason you didn't just do 
```ts
} else if (d instanceof HingeDriver || d instanceof SliderDriver) {
```
is because you are going to change it in the future to be unique

fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
```diff
@@ -286,18 +289,24 @@ export class PWMGroup extends SimOutputGroup {
...
+            average += speed;
             console.log(port, speed)
         }
-        // average /= this.ports.length
-
-        // this.drivers.forEach(d => {
-        //     (d as WheelDriver).targetWheelSpeed = average * 40
-        //     d.Update(_deltaT)
-        // })
+        average /= this.ports.length
```
Dhruv-0-Arora: You could do something like
```ts
        const average =  this.ports.reduce((sum, port) => {
            const speed = SimPWM.GetSpeed(`${port}`) ?? 0
            console.log(port, speed)
            return sum + speed
        }, 0) / this.ports.length
```

Dhruv-0-Arora (changes requested):
Don't forget your 
```
TODO: integrate with motor config when max speeds are available instead
of hardcoding 40
```
and is the input system supposed to break for the robot when you attach the roborio?

fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
azaleacolburn: Might be better to do a switch statement, since this is looping and switch statements don't use comparisons which are quite slow when looped because of branch predicting. This probably won't matter, but might since this is in an update function that will only be called more often as more bodies are spawned.

PepperLola: I didn't realize they had the same name until after I'd separated the clauses but I'll change that to not repeat itself.

PepperLola:
> Don't forget your
> 
> ```
> TODO: integrate with motor config when max speeds are available instead
> of hardcoding 40
> ```
> 
> and is the input system supposed to break for the robot when you attach the roborio?

The input system is not supposed to work once you create a device because the brain is switched to the WPILibBrain which does not respond to user input. I think I'm going to come back to integrating it with the motor config later because it's not entirely blocked by it and it would be nice to have this merged soon.

fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
PepperLola: > Might be better to do a switch statement, since this is looping and switch statements don't use comparisons which are quite slow when looped because of branch predicting. This probably won't matter, but might since this is in an update function that will only be called more often as more bodies are spawned.

The only way I know of to make this a switch statement is
```ts
switch (true) {
  case d instanceof WheelDriver:
    // wheel
  case d instanceof HingeDriver:
  case d instanceof SliderDriver:
    // hinge and slider
}
```

I'm not sure if this has the same benefits of a regular switch statement, since the cases would still need to be evaluated at runtime so I don't think it could make a jump table. I also think this syntax is less immediately understandable.

We would also need to make it possible to simulate multiple robots at once for this to be impacted by the number of robots spawned, which I don't think is currently supported by our websocket implementation.

azaleacolburn: Oh you're right, a jump table probably isn't possible for whatever operation instanceof does since it doesn't match to a value 👍

azaleacolburn (approved):
LGTM

Dhruv-0-Arora (approved):
Fabuloso 🌹

fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts
azaleacolburn: The only way to do something like this would be to create a mapping object, then jumping with d.constructor.name, which seems excessive, I'll keep this in mind incase this update function ever becomes a bottleneck though.

azaleacolburn: [Seems like your implementation would be more efficient though](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html)

We could also match directly on `d.constructor` if d will never be a subclass of these drivers
https://stackoverflow.com/questions/36332665/how-to-use-instanceof-in-a-switch-statement

PR #1024, opened 2024-07-10
AARD-1741: Exporter PIP Dependency Updates

BrandonPacewic:
- [x] Fixed a bug related to our protobuf version being very old and having no way of updating it outside of reinstalling Fusion.
- [x] Will now check the versions of all installed dependencies upon error and update them if necessary.
- [x] Integration with #1010 for separate error messages depending on missing protoc files or improper dependencies.
- [x] Removed the need for `deps.py` to be a blank import.
  - Prep for #1013. 
- [x] Will now provide a separate error message if the complied proto files are missing rather than throw a could not install dependency error.

### Testing Done
- [x] Running on pre-existing Windows.
- [x] Running on pre-existing Mac.
- [x] Running on fresh-install Windows.
- [x] Running on fresh-install Mac.

### To Test For Reviewers:
- Open Fusion with the extension as normal.
- It should update / reinstall `python-protobuf` and ask you to restart Fusion.
  - Follow said instructions.
- Upon reopening, everything should continue to work as intended.
- If you like, update your system `protoc` version and run the build script (`build.sh`) from inside `exporter/SynthesisFusionAddin/proto`.
- Upon restarting Fusion AGAIN, everything should still work normally.
  - Previously compiling the proto files with the most recent version of `protoc` would have broken the exporter.

Closes: [JIRA Issue - Task](https://jira.autodesk.com/browse/AARD-1741)
Fixes: [JIRA Issue - Bug](https://jira.autodesk.com/browse/AARD-1734)

PR #1064, opened 2024-07-30
Exporter Formatting Workflow Fix

BrandonPacewic:


BrandonPacewic:
For some reason this is only affecting #1024.

PR #1065, opened 2024-07-30
APS Required Login Fix

BrandonPacewic:
Fix an issue where APS login was always required, even if upload was not selected as the export location.


azaleacolburn (approved):
I can't believe we missed this

BrandonPacewic:
> I can't believe we missed this

Yea... Me too...

PepperLola (approved):
Good Looking

PR #1066, opened 2024-07-30
Real Case Usage Exporter Fixes

BrandonPacewic:
- Fixed an issue where the Synthesis Addin would hang if `Run on Startup` was checked.
- Fixed an issue where the exporter assumed that all saved joint tokens would never be removed from the design, resulting in handlers not being added properly.


azaleacolburn (approved):
LGTM

PepperLola (approved):
Works on startup and doesn't crash when I load a bad design

PR #1069, opened 2024-07-31
Import Local Field

a-crowell:
### Description
The Import Local Mirabuf panel used to load every file as a robot, now it goes based on the dropdown.


Dhruv-0-Arora (approved):
Looks good 👍

As a future task, can we put the import local mira panel into the "spawn asset" modal rather than have it as a stand alone button on the MainHUD.
I also think this UI is a bit funky
Screenshot 2024-07-31 at 12 10 14 PM

a-crowell:
@Dhruv-0-Arora I agree this would be good to merge into Spawn Asset someday. Then we don't have to worry about that UI 👍

Dhruv-0-Arora:
> @Dhruv-0-Arora I agree this would be good to merge into Spawn Asset someday. Then we don't have to worry about that UI 👍

I shall bestow this responsibility upon myself and have it done today 🫡

LucaHaverty (approved):
LGTM 👍 👍 👍

PR #1070, opened 2024-07-31
Quality Configuration

Dhruv-0-Arora:
## Description
Creates a new panel that allows users to have more control over their graphics settings (specifically cascading shadows). Allows for enhancing graphics or improving performance.

Screenshot 2025-06-17 at 4 53 22 PM

Screenshot 2025-06-17 at 4 56 36 PM

### Features
- [x] Change light intensity
- [x] Toggle cascading shadows
- [x] Change Max Far (zoom out distance before the shadows are unrendered)
- [x] Change Shadow Map Size (shadow texture size/detail)
- [x] Reset cascading shadows values button
- [x] Anti-aliasing toggle (will hot reload the page)

### Notes
- Removed previous quality dropdown
Screenshot 2025-06-17 at 4 56 05 PM

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1780)


LucaHaverty:
Hovering over the '?' icon causes a jittering behavior for me because it shifts the panel over slightly

LucaHaverty:
Is this PR waiting for something else to be merged in or on hold?

LucaHaverty (commented):
I feel like the vast majority of users won't know what things like "cascades" and "shadow map size" are, so in my opinion it would be better to have preset quality levels that we know are stable rather than give them control over all the specific settings with the potential of breaking something or ending up with crazy artifacts.

BrandonPacewic:
> I feel like the vast majority of users won't know what things like "cascades" and "shadow map size" are, so in my opinion it would be better to have preset quality levels that we know are stable rather than give them control over all the specific settings with the potential of breaking something or ending up with crazy artifacts.

I agree, but I do think we should eventually have both as an option. Similar to how most games give you low, medium, and high presets but still let you change whatever specific option you want. (At your own risk of course).

BrandonPacewic (commented):
> I feel like the vast majority of users won't know what things like "cascades" and "shadow map size" are, so in my opinion it would be better to have preset quality levels that we know are stable rather than give them control over all the specific settings with the potential of breaking something or ending up with crazy artifacts.

I agree, but I do think we should eventually have both as an option. Similar to how most games give you low, medium, and high presets but still let you change whatever specific option you want. (At your own risk of course).

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -230,42 +233,89 @@ class SceneRenderer extends WorldSystem {
...
+        //     if (object.material) {
+        //         if (Array.isArray(object.material)) {
+        //             object.material.forEach(mat => {
+        //                 mat.needsUpdate = true
+        //             })
+        //         } else {
+        //             object.material.needsUpdate = true
+        //         }
+        //     }
+        // })
```
BrandonPacewic: Does this update still not need to be done?

fission/.eslintrc.cjs
BrandonPacewic, line 1: If possible we should avoid formatting these files beyond what has already been done. We don't enforce style here so your essentially putting your own auto formatting standards onto the repo.

fission/README.md
BrandonPacewic, line 1: Same as above. I'm not necessarily opposed to letting these changes go through since this file was already poorly formatted.

fission/postcss.config.js
BrandonPacewic, line 1: Same as before. We should leave this change in though since we use double quotes and trailing commas anyways.

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -86,11 +85,11 @@ class SceneRenderer extends WorldSystem {
         this._scene = new THREE.Scene()
 
         this._renderer = new THREE.WebGLRenderer({
-            // Following parameters are used to optimize post-processing
             powerPreference: "high-performance",
             antialias: false,
             stencil: false,
-            depth: false,
+            depth: !PreferencesSystem.getGraphicsPreferences().antiAliasing,
+            // logarithmicDepthBuffer: true,
```
BrandonPacewic: This still need to be here? Should add a comment if we want it to be a reference for later.

fission/src/ui/panels/GraphicsSettingsPanel.tsx
```diff
@@ -0,0 +1,184 @@
...
+                            />
+                        </Box>
+                    </>
+                ) : (
+                    <></>
+                )}
+                <div className="flex items-center justify-center mt-1 mb-0.5 mx-[5%]">
+                    <SectionLabel size={LabelSize.Medium} className="text-center">
+                        Hot-reload Settings
+                    </SectionLabel>
```
BrandonPacewic: I'm partial to the opinion that this label should be removed. At first I thought it was some sort of button or toggle and it seemed out of place.

fission/src/ui/panels/GraphicsSettingsPanel.tsx
```diff
@@ -0,0 +1,184 @@
...
+                <Checkbox
+                    label="Anti-Aliasing"
+                    defaultState={antiAliasing}
+                    onClick={checked => {
+                        // saving the new preference
+                        setAntiAliasing(checked)
+                        setReload(true)
+                    }}
+                    tooltipText="Requires a browser refresh"
+                />
```
BrandonPacewic: Should probably have some sort of user confirmation here before you auto refresh and reset the current scene. Could do this by adding some sort of confirmation after the user clicks accept with a change to this option or could be saved with some sort of "Refresh is required to apply some settings" message.

fission/tailwind.config.mjs
BrandonPacewic, line 1: Same as other formatting changes comment. While I agree that this file is also poorly formatted this should probably be its own pr.

fission/vite.config.ts
BrandonPacewic, line 1: Same as other formatting changes comment.

BrandonPacewic (changes requested):
Additionally I don't know if I'm the biggest fan of having the panel for settings open in a different place compared to the main settings panel. It does make sense from a visual perspective so you can see what changes your actually making. Open to feedback on that.

fission/src/systems/scene/SceneRenderer.ts
Dhruv-0-Arora: Removed the comments

fission/.eslintrc.cjs
Dhruv-0-Arora: Ill revert the changes to .eslintrc.cjs and ill keep the changes for the readme and postcss config

fission/tailwind.config.mjs
Dhruv-0-Arora: Reverted

fission/vite.config.ts
Dhruv-0-Arora: Reverted

BrandonPacewic (changes requested):
Small merge conflict.

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -90,20 +89,19 @@ class SceneRenderer extends WorldSystem {
...
         })
         this._renderer.setClearColor(CLEAR_COLOR)
         this._renderer.setPixelRatio(window.devicePixelRatio)
         this._renderer.shadowMap.enabled = true
         this._renderer.shadowMap.type = THREE.PCFSoftShadowMap
         this._renderer.setSize(window.innerWidth, window.innerHeight)
 
-        // Adding the lighting using quality preferences
-        this.ChangeLighting(PreferencesSystem.getGlobalPreference<string>("QualitySettings"))
+        // Adding the lighting uisng quality preferences
```
PepperLola:
```suggestion
        // Adding the lighting using quality preferences
```

PepperLola (commented):
I think it looks good. We might want to eventually make the settings tabbed rather than opening new panels with buttons but that should happen after the UI refactor. I suggested a change for a typo but other than that I think it's ready to merge.

Do we want to limit the maximum values that someone can choose? I think the upper end of the texture size values is almost impossible to run at an acceptable frame rate (especially since most people will probably be using chromebooks), so would it make sense to limit options that would tank performance like that to ones that are actually feasible to run? Especially if the settings are persisted, because it would be hard to reset the values if it takes a minute to click each button.

Dhruv-0-Arora:
> I think it looks good. We might want to eventually make the settings tabbed rather than opening new panels with buttons but that should happen after the UI refactor. I suggested a change for a typo but other than that I think it's ready to merge.
> 
> Do we want to limit the maximum values that someone can choose? I think the upper end of the texture size values is almost impossible to run at an acceptable frame rate (especially since most people will probably be using chromebooks), so would it make sense to limit options that would tank performance like that to ones that are actually feasible to run? Especially if the settings are persisted, because it would be hard to reset the values if it takes a minute to click each button.

@PepperLola 
Currently I have the max shadow map size set to the renderers maxTextureSize capability, so on slower computers it will automatically limit the user. We could also manually set the value.

Screenshot 2025-06-19 at 3 39 01 PM

BrandonPacewic (approved):
I think it's time to move onto something else. I have a few remaining issues that I think we should address after the UI refactor:

1. I agree that this should probably be a tab within the setting panel itself, rather than opening a new panel.
2. I think that we should consider splitting up the settings into collections (maybe a drop down) of basic and advanced.
3. I'm not that big of a fan of how we currently handle settings that need a refresh to take effect, again something that will be easier to handle (or at least implement) post UI refactor.
4. I think removing the presets all together is not what we want the final version of this to look like. I'm adding this as an additional task blocked by AARD-1781.

Since all my problems are essentially just UI nitpicks and the functionality is here I think its appropriate to make these changes after we finish the UI refactor.

PepperLola (approved):
I think it looks good except for the max texture size issues. In my experience it isn't properly limited and setting it too high can cause the app to become entirely unusable. I'm willing to leave that until later though, but I do think it should be addressed at some point.

BrandonPacewic:
> I think it looks good except for the max texture size issues. In my experience it isn't properly limited and setting it too high can cause the app to become entirely unusable. I'm willing to leave that until later though, but I do think it should be addressed at some point.

Noted AARD-1928.

PR #1068, opened 2024-07-30
Combined Configuration

LucaHaverty:
### Description
Merging all of the different assembly configuration panels into one. This includes intake config, ejector config, inputs, sequential joints, and scoring zones. 

### Objectives
- [x] Panel Navigation
- [x] Intake config support
- [x] Ejector config support
- [x] Input config support
- [x] Sequential join config support
- [x] Scoring zone config support
- [x] Input scheme assignment support

### No item selected
image

### Hover and click effects
image

### Item selected
image

### Ability to chain together for complex behavior
image

### Other features
- Custom header when no item is selected
- Custom message when there are no available options
- Optional delete behavior
- Optional add new behavior
- Custom data stored with options other than just the text displayed
- Optional indentation (used when chaining select menus together)

### Sequential joint configuration
image

### Control scheme configuration
image

image

### Assembly management
image

### Scroll views with persistent headers
https://github.com/user-attachments/assets/6f6a6f7a-a5a3-4710-ad6a-ee1a6787586c

> [!note]
> Some of this UI is still rough, so please leave any and all design feedback in reviews. Especially with the configuration interfaces for intake and ejector.

[Combined Config Issue](https://jira.autodesk.com/browse/AARD-1783)
[Sequential Elevator Issue](https://jira.autodesk.com/browse/AARD-1711)
[Remake Change Inputs Issue](https://jira.autodesk.com/browse/AARD-1776)
[UI Zoom Bug Issue](https://jira.autodesk.com/browse/AARD-1801)

PR #1071, opened 2024-07-31
Moving Import Local Mirabuf

Dhruv-0-Arora:
### Description
Removed the "Import Local Mirabuf" from MainHUD and moved it to "Spawn Asset" panel.

Screenshot 2024-07-31 at 1 19 34 PM

Screenshot 2024-07-31 at 1 20 25 PM

> **Note:** Merge #1069 first

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1787)


a-crowell (changes requested):
Looking good! Make sure the control schemes panel still pops up upon uploading a robot.

a-crowell (approved):
LGTM!

fission/src/ui/components/MainHUD.tsx
```diff
@@ -105,11 +105,11 @@ const MainHUD: React.FC = () => {
                         icon={SynthesisIcons.Gamepad}
                         onClick={() => openModal("change-inputs")}
                     />
-                    <MainHUDButton
+                    {/* <MainHUDButton
```
BrandonPacewic: Thinking we should remove this section now.

BrandonPacewic (commented):
This helps with one of our pain points from beta day about `mirabuf` being confusing. Having everything under spawn asset is definitely better imo.

LucaHaverty:
Do you all think it looks better for the button to be at the bottom of the panel instead of under "saved assets"? I think it looks a bit cleaner at the bottom, but here's a comparison:

Under saved assets             |  Bottom
:-------------------------:|:-------------------------:
image  |  image

Also, maybe we could add an icon to the button kind of like this:
image

fission/src/ui/components/MainHUD.tsx
Dhruv-0-Arora: Whoops

BrandonPacewic (commented):
~~I'm ready to approve this once we decide what we want to do with Luca's suggestions.~~

https://github.com/Autodesk/synthesis/issues/1071#issuecomment-2274481787

BrandonPacewic (changes requested):
We should wait to put this in until after #1068 since it will result in a merge conflict. Should be easier to fix in a smaller PR like this.

> [!note]
> - Waiting for #1068.

LucaHaverty (approved):

BrandonPacewic (approved):

PR #1073, opened 2024-07-31
Exporter Packaging and Installer `[AARD-1742]`

BrandonPacewic:
Now that we only have the exporter that needs to be installed locally. Previously we were installing the extension into the user directory for addins, we are now using the `ApplicationPlugins` folder which is standard Autodesk wide. It also has the advantage of not being able to be changed by the user. This, however, also means that the exporter will no longer have permissions within it's installed directory so we need to redirect things into the `Application Support` path. Hence the changes within the exporter that would appear to not be directly related to packaging.

Essentially this installer works exactly how installers work from the [Fusion Addin Store](https://apps.autodesk.com/FUSION/en/Home/Index) work.

- [x] For MacOS, new `.pkg` installer.
- [x] For Windows, new `.exe` installer.

When running the exporter from the installer logs can be found in:
- Mac: `~/.config/Autodesk/Synthesis/Logs`
- Win: `%appdata%/Autodesk/Synthesis/Logs`

Otherwise the log directory will be the same:
- `exporter/SynthesisFusionExporter/logs/`

This was done as we don't always have write permissions in the `ApplicationPlugins` directory.

Some other things:
- [x] ~~Install script will now mark the exporter with a version so you can tell them apart.~~
- [x] ~~Install script has an additional step to uninstall previous versions of the exporter.~~

> [!warning]
> Switching between running the version of the exporter within `ApplicationPlugins` and another manually added path will cause problems.
>
> Restart Fusion in-between runs of the exporter within different directories to avoid caching issues.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1742)


exporter/SynthesisFusionAddin/src/Logging.py
```diff
@@ -30,14 +30,19 @@ def cleanupHandlers(self) -> None:
...
+        logFileFolder = makeDirectories(os.path.join(f"{pathlib.Path(__file__).parent.parent}", "logs"))
+    else:
+        logFileFolder = makeDirectories(os.path.join(SUPPORT_PATH, "Logs"))
+
+    if not IS_RELEASE:
+        logFiles = [os.path.join(logFileFolder, file) for file in os.listdir(logFileFolder) if file.endswith(".log")]
+        logFiles.sort()
+        if len(logFiles) >= MAX_LOG_FILES_TO_KEEP:
+            for file in logFiles[: len(logFiles) - MAX_LOG_FILES_TO_KEEP]:
+                os.remove(file)
```
azaleacolburn: Is this split into multiple if statements for readability?

azaleacolburn (approved):
Works perfectly for me!

exporter/SynthesisFusionAddin/src/Logging.py
BrandonPacewic: More or less, I would be open to other suggestions if you think you have a better way to format this.

azaleacolburn: I mean I wouldn't have an issue with:

```suggestion
    if not IS_RELEASE:
        logFileFolder = makeDirectories(os.path.join(f"{pathlib.Path(__file__).parent.parent}", "logs"))

        logFiles = [os.path.join(logFileFolder, file) for file in os.listdir(logFileFolder) if file.endswith(".log")]
        logFiles.sort()
        if len(logFiles) >= MAX_LOG_FILES_TO_KEEP:
            for file in logFiles[: len(logFiles) - MAX_LOG_FILES_TO_KEEP]:
                os.remove(file)
    else:
        logFileFolder = makeDirectories(os.path.join(SUPPORT_PATH, "Logs"))    
```

PepperLola (approved):
Works well 👎

BrandonPacewic (commented):
> [!important]
> Should have someone test on Windows before we merge.

PR #1074, opened 2024-08-01
Bug: Friction Override Slider Shown on Startup

BrandonPacewic:
Fixed a bug that would cause the friction override slider to show on startup even if the friction override checkbox was set to false.

### DIFF
--- a/exporter/SynthesisFusionAddin/src/Parser/ExporterOptions.py
+++ b/exporter/SynthesisFusionAddin/src/Parser/ExporterOptions.py
@@ -51,7 +51,7 @@ class ExporterOptions:
     autoCalcGamepieceWeight: bool = field(default=False)
 
     frictionOverride: bool = field(default=False)
-    frictionOverrideCoeff: float | None = field(default=None)
+    frictionOverrideCoeff: float = field(default=0.5)
 
     compressOutput: bool = field(default=True)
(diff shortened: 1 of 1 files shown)


exporter/SynthesisFusionAddin/src/Parser/ExporterOptions.py
```diff
@@ -51,7 +51,7 @@ class ExporterOptions:
     autoCalcGamepieceWeight: bool = field(default=False)
 
     frictionOverride: bool = field(default=False)
-    frictionOverrideCoeff: float | None = field(default=None)
+    frictionOverrideCoeff: float = field(default=0.5)
```
azaleacolburn, line 54: This was causing the slider to always show up?

azaleacolburn (approved):

exporter/SynthesisFusionAddin/src/Parser/ExporterOptions.py
BrandonPacewic: Would cause the `__init__` of `GeneralOptionsTab` to crash before it could set the visibility of the slider since the Fusion API was not expecting to see a value of `None` for the slider position.

PepperLola (approved):
Friction override no longer enabled at startup

PR #1075, opened 2024-08-01
Operation Remove the Javascript

BrandonPacewic:
Previously over half our code was being counted as javascript even though we only have like 50 lines total. We want this to be a typescript repository, not a javascript one. This is the current workaround found by @azaleacolburn.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,6 @@
 * text=auto
+
+# Workaround for a bug it GitHub's language detection,
+# previously half of our code was being counted as JavaScript.. for some reason...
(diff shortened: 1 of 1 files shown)


HunterBarclay (approved):
We have been saved

azaleacolburn (approved):
This is beautiful

.gitattributes
```diff
@@ -1 +1,6 @@
 * text=auto
+
+# Workaround for a bug it GitHub's language detection,
```
BrandonPacewic:
```suggestion
# Workaround for a bug in GitHub's language detection,
```

PR #1078, opened 2024-08-02
Motors in Drivers

a-crowell:
### Description
Drivers have access to max velocity/acceleration and users can adjust them.

### Objectives
- [X] Drivers get and use max velocity/acceleration from preferences or the mirabuf file
- [X] Joint Configuration Panel with saving and canceling features

> [!Note]
> * Certain motors in the assemblies do not work. For example, 2018 sliders aren't given enough force to move. While I could clamp them as they come in, material is a big enough factor that it's easy enough to adjust in the panel.
> * While you'll see acceleration and force as umbrella terms for ease, SliderDrivers use force, HingeDrivers use torque, and WheelDrivers use acceleration. In the future, if we switch to using the vehicle controllers, WheelDrivers would use torque.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1703)


fission/src/systems/simulation/driver/HingeDriver.ts
```diff
@@ -25,13 +28,21 @@ class HingeDriver extends Driver {
...
 
-    public set minTorqueLimit(nm: number) {
-        const motorSettings = this._constraint.GetMotorSettings()
-        motorSettings.mMinTorqueLimit = nm
+    public get maxVelocity(): number {
+        return this._maxVelocity
+    }
+    public set maxVelocity(radsPerSec: number) {
+        this._maxVelocity = radsPerSec
     }
```
LucaHaverty: Instead of having a separate getter and setter method for `maxVelocity`, should it just be public? (same with all the others but I won't add individual comments on every one)

fission/src/systems/simulation/driver/HingeDriver.ts
```diff
@@ -53,31 +64,32 @@ class HingeDriver extends Driver {
...
     public Update(_: number): void {
         if (this._controlMode == DriverControlMode.Velocity) {
-            this._constraint.SetTargetAngularVelocity(this._targetVelocity)
+            this._constraint.SetTargetAngularVelocity(this._accelerationDirection * this._maxVelocity)
         } else if (this._controlMode == DriverControlMode.Position) {
+            let ang = this._targetAngle
+
+            if (ang - this._prevAng < -this.maxVelocity) ang = this._prevAng - this._maxVelocity
+            if (ang - this._prevAng > this.maxVelocity) ang = this._prevAng + this._maxVelocity
             this._constraint.SetTargetAngle(this._targetAngle)
```
LucaHaverty: Did you mean to set the target angle to `ang` instead of `this._targetAngle`?

fission/src/ui/panels/configuring/ConfigureJointsPanel.tsx
```diff
@@ -0,0 +1,259 @@
...
+                    maxVelocity: vel,
+                    maxForce: force,
+                })
+
+                PreferencesSystem.getRobotPreferences(robot.assemblyName).motors = removedMotor
+            }
+        }
+
+        PreferencesSystem.savePreferences()
+    }
```
LucaHaverty: I believe in react it's better practice to use a useCallback hook in cases like this, but @PepperLola or @HunterBarclay could you confirm? It would look something like this:

```ts
const onChange = useCallback((vel: number, force: number) => {
    // Function body
}, [driver, velocity, force]);
```

LucaHaverty (changes requested):
Love this! Everything I tested has worked great so far, but I left a couple code suggestions below. 

Also, to me it would make more sense for this panel to be called "motor configuration" instead of "joint configuration" because it also includes wheels, which I guess technically are joints but a lot of people won't think of wheels when they read "joints". Up to you though, might just be me

a-crowell:
@LucaHaverty I had the same debate last year, but decided on joints because a lot of these are multi-motor, especially the drive. Do you have other ideas for joints?

LucaHaverty:
> @LucaHaverty I had the same debate last year, but decided on joints because a lot of these are multi-motor, especially the drive. Do you have other ideas for joints?

Oh yeah that's a good point. Maybe mechanisms or subsystems? I'm not sure though, joints works too.

Dhruv-0-Arora:
> > @LucaHaverty I had the same debate last year, but decided on joints because a lot of these are multi-motor, especially the drive. Do you have other ideas for joints?
> 
> Oh yeah that's a good point. Maybe mechanisms or subsystems? I'm not sure though, joints works too.

subsystems 👽

a-crowell:
How did running bun run format (only changing the callback in subsystem panel) make it start failing fusion black format validation too?

BrandonPacewic (commented):
@a-crowell See #1081, the fusion formatter is just broken right now. Nothing you did.

a-crowell:
@BrandonPacewic Lovely, have fun with that.

BrandonPacewic:
> @BrandonPacewic Lovely, have fun with that.

Oh I will... 😢

HunterBarclay (changes requested):
It looks like the issue of the elevators and arms falling really fast to gravity is an issue again. @LucaHaverty what was the cause last time?

LucaHaverty:
> It looks like the issue of the elevators and arms falling really fast to gravity is an issue again. @LucaHaverty what was the cause last time?

@HunterBarclay The torque you set helps resist other forces like gravity and those of other mechanisms, so if you set the torque on a subsystem really high, and the velocity low you get snappier movement that resists other forces. I'm not sure if there's a better solution in this case, but that's the only thing I tried that worked.

a-crowell:
> @HunterBarclay The torque you set helps resist other forces like gravity and those of other mechanisms, so if you set the torque on a subsystem really high, and the velocity low you get snappier movement that resists other forces. I'm not sure if there's a better solution in this case, but that's the only thing I tried that worked.

@LucaHaverty That makes sense. Gravity is the kind of thing designers need to pay attention to when it comes to power behind their motors.

LucaHaverty:
@a-crowell I have a small UI thing. Do you think you could add dividers between the subsystems to have visual continuity with the new input configuration panel?

image

LucaHaverty:
It seems like with the arm on mean machine, turning the velocity any higher than 20 doesn't change much because it's limited by the torque. Maybe we should reduce the max allowed velocity because I don't think anything much higher than that is realistic, and would require an extremely high torque. Same goes for the elevators.

a-crowell:
> @a-crowell I have a small UI thing. Do you think you could add dividers between the subsystems to have visual continuity with the new input configuration panel?
> 
> image

Yeah, what do you think of adding dividers to scoring zones too? I'm torn between adding them or not.

LucaHaverty:
> > @a-crowell I have a small UI thing. Do you think you could add dividers between the subsystems to have visual continuity with the new input configuration panel?
> > image
> 
> Yeah, what do you think of adding dividers to scoring zones too? I'm torn between adding them or not.

I think we're planning on reworking the visuals of the scoring zones panel in the near future, and I think when we do that we should include the dividers to stay consistent with the rest of the UI.

a-crowell:
> I think we're planning on reworking the visuals of the scoring zones panel in the near future, and I think when we do that we should include the dividers to stay consistent with the rest of the UI.

Sounds good. I'll leave them alone until then.

LucaHaverty (changes requested):
When I spawn in mean machine, enable realistic gravity, then modify the arm joint torque, the arm moves up against gravity all the way to it's limit and gets stuck there.

a-crowell:
> When I spawn in mean machine, enable realistic gravity, then modify the arm joint torque, the arm moves up against gravity all the way to it's limit and gets stuck there.

What was the torque value?

LucaHaverty:
> > When I spawn in mean machine, enable realistic gravity, then modify the arm joint torque, the arm moves up against gravity all the way to it's limit and gets stuck there.
> 
> What was the torque value?

It happens with any torque value as soon as I change it. Doesn't seem to matter what the value is or how much I change it by.

a-crowell:
> It happens with any torque value as soon as I change it. Doesn't seem to matter what the value is or how much I change it by.

@LucaHaverty Forgot a negative. Fixed now!

LucaHaverty (approved):
Amazing work! Can't wait to mess around with this more once swerve is in

fission/src/mirabuf/MirabufParser.ts
```diff
@@ -240,7 +240,7 @@ class MirabufParser {
             console.log("Failed to get part definitions")
             return
         }
-        console.log(partDefinitions)
+        console.debug(partDefinitions)
```
BrandonPacewic:
```suggestion
```

One little merge conflict.

BrandonPacewic (commented):
I could not seem to find where to pull up the config panel. The subsystem button on main HUD appeared to do nothing and I failed to find where the driver config is in the new configure objects panel.

LucaHaverty:
> I could not seem to find where to pull up the config panel. The subsystem button on main HUD appeared to do nothing and I failed to find where the driver config is in the new configure objects panel.

Looks like the panel got deleted from `Synthesis.tsx` in a merge conflict :(

a-crowell:
@BrandonPacewic @LucaHaverty Forgot to stage the Synthesis file. Thanks for catching it!

fission/src/ui/panels/configuring/ConfigureSubsystemsPanel.tsx
```diff
@@ -0,0 +1,277 @@
...
+                            onChange={(_, _force: number | number[]) => {
+                                setForce(_force as number)
+                                onChange(velocity, _force as number)
+                            }}
+                            step={0.01}
+                        />
+                    ) : driver instanceof HingeDriver ? (
+                        <Label>Select Realistic Gravity in Settings for torque config</Label>
+                    ) : (
+                        <Label>Select Realistic Gravity in Settings for force config</Label>
```
BrandonPacewic, line 127: I'm not sure I'm a fan of these labels. I agree it would be helpful to add some sort of description but, at least to me, it feels weird to 'Select (a) Realistic Gravity' when I don't have any sort of indication of what the units here are.

a-crowell: I could take them out entirely and let users figure out that torque/force config appears when they enable realistic gravity.  That could be in the tutorial. I like enable wording better at least if we do keep it in there. Do you have other wording ideas?

BrandonPacewic: Is `Enable Realistic Gravity` a setting?

a-crowell: Yep, in General Settings. It overrides force/torque though, so users can't modify that acceleration if Realistic Gravity isn't enabled. That's what the note is for.

BrandonPacewic: Got it, for now I think it's fine. We should probably add something later for people like me who didn't know about the `Realistic Gravity` option 🙃

LucaHaverty (approved):
Looks great!

fission/src/ui/panels/configuring/ConfigureSubsystemsPanel.tsx
a-crowell: After in-person discussion, the plan is to merge this in then add tooltip functionality for the note when the UI is moved into Configure in a new PR.

BrandonPacewic (approved):
🪨 🧗

BrandonPacewic (approved):

HunterBarclay (approved):

PR #1080, opened 2024-08-03
Black Format Workflow Patch

HunterBarclay:
### Description
Added specific versions of `black` and `isort` to the format validation workflow for the exporter.


HunterBarclay:
Didn't work

PR #1081, opened 2024-08-06
Exporter Formatting Fix

BrandonPacewic:
For some reason the isort formatter on mac wants to add a line between these imports:

```python
import adsk.core
import adsk.fusion
from google.protobuf.json_format import MessageToJson

from proto.proto_out import assembly_pb2, types_pb2
```

And on linux it wants to remove this line:

```python
import adsk.core
import adsk.fusion
from google.protobuf.json_format import MessageToJson
from proto.proto_out import assembly_pb2, types_pb2
```

¯\\_(ツ)_/¯

Edit:

This is actually because our local machines know that `proto.proto_out` is a local path but because we don't track that with git the server can't find it as a local path so it thinks that it's a package import.

### DIFF
--- a/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
+++ b/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
@@ -6,7 +6,6 @@ from typing import *
 
 import adsk.core
 import adsk.fusion
-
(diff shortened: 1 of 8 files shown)


azaleacolburn (approved):
LGTM

azaleacolburn (approved):
I didn't mean to re-request a review

Dhruv-0-Arora (approved):
👀

PR #1082, opened 2024-08-06
Mirabuf Caching OPFS Back Up

a-crowell:
### Description
We cannot create writables in OPFS on Safari/iOS, so this PR puts the ArrayBuffers in a hashMap so there is at least temporary storage.

### Testing Done
- [X] Chrome/Firefox on Windows (with canOPFS directly set true/false)
- [X] Safari (thanks @Dhruv-0-Arora)

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1797)

### DIFF
--- a/fission/src/mirabuf/MirabufLoader.ts
+++ b/fission/src/mirabuf/MirabufLoader.ts
@@ -32,17 +32,14 @@ const fieldFolderHandle = await root.getDirectoryHandle(fieldsDirName, { create:
 export const backUpRobots: Map<MirabufCacheID, ArrayBuffer> = new Map<MirabufCacheID, ArrayBuffer>()
 export const backUpFields: Map<MirabufCacheID, ArrayBuffer> = new Map<MirabufCacheID, ArrayBuffer>()
 
-const canOPFS = await (async() => {
+const canOPFS = await (async () => {
     try {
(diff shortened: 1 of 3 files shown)


Dhruv-0-Arora (approved):
Looks good 👍

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -353,7 +401,7 @@ class MirabufCachingService {
 
 export enum MiraType {
     ROBOT = 1,
-    FIELD = 2,
+    FIELD,
```
BrandonPacewic, line 404: Looks like we got safari working! One minor note, was this supposed to be changed?

```suggestion
    FIELD = 2,
```

LucaHaverty (approved):
Worked on chrome, microsoft edge (🤮), and firefox. Nice work!

fission/src/mirabuf/MirabufLoader.ts
a-crowell: Yep! I just learned the 2 was unnecessary because it increments based on the last one.

BrandonPacewic: If it works it works.

BrandonPacewic (approved):
Now I can use my favorite web browser!

PR #1084, opened 2024-08-07
Fission Unit Tests Playwright Update

a-crowell:
### Description
Fission unit tests currently failing for everyone due to Playwright update to 1.46.


a-crowell:
Yeah this didn't work. Bun install got the new version and tried clearing browser cache, but still getting the same error.

PR #1085, opened 2024-08-07
Transform Gizmo Refactor

Dhruv-0-Arora:
## Description
Refactoring the Transform Gizmos to be a `SceneObject`.

### Changes
- All of the transform gizmo computations happen in the `GizmoSceneObject` file. This includes mapping transformations of the mesh to the mirabuf.
- Changed implementation in ejector, intake, and scoring zones configurations.
- Removed implementation from `SceneRenderer`

### Hunter's Additions
- Reworked how a brain is selected for modification/scheme assignment.
- Added new panel specifically for newly spawned in robots.
- Moved scheme selection into the config panel.
- Added `GlobalUIControls` and `GlobalUIComponent` to provide access to context controls outside of react components. Probably not the best practice, but make sense considering these context controls will be relevant for the entire lifetime of the application, not just a single component/menu.
- All current uses of the `GizmoSceneObject` are via the `TransformGizmoControl` component. This allows were easy lifetime controls when using the gizmos.
- Added a move option to the config menu for moving and orientating a robot/field after spawning it.
- Added "Reset Orientation" button to the transform gizmo controls for resetting rotation back to default.
- Changed out default mesh for gizmos to a blank mesh instead of a ball.
- Moved gizmo to the center of the assembly you are moving (center of mass for robots, center of the AA box for fields).
- Moved newly spawned assemblies to `(0, 0 + half of bounding box height, 0)` anchored around the center of their bounding box.
- Removed enter and escape controls for spawning placement/cancellation. Debating adding them back in as the controls for the initial config panel.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1798)


LucaHaverty (changes requested):
Couple Issues:

Pressing enter when configuring the intake, ejector, or scoring zones makes the gizmo disappear

Spawning a robot, then deleting the assembly before pressing enter leaves the gizmo behind

Dhruv-0-Arora:
> Couple Issues:
> 
> Pressing enter when configuring the intake, ejector, or scoring zones makes the gizmo disappear
> 
> Spawning a robot, then deleting the assembly before pressing enter leaves the gizmo behind

🫠

BrandonPacewic (commented):
Looks like this has some pretty extensive merge conflicts from #1068.

a-crowell (approved):
One more merge conflict. Other than that, looks good.

BrandonPacewic (commented):
I'm still not sure that this 100% addresses the concerns that @HunterBarclay had.

Dhruv-0-Arora:
> Looks like this has some pretty extensive merge conflicts from #1068.

Yep I just added all the refactoring changes that I could remember

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
```diff
@@ -217,8 +221,8 @@ const ConfigureGamepiecePickupInterface: React.FC<ConfigPickupProps> = ({ select
                         const robotTransformation = JoltMat44_ThreeMatrix4(
                             World.PhysicsSystem.GetBody(selectedRobot.GetRootNodeId()!).GetWorldTransform()
                         )
-                        transformGizmo.mesh.position.setFromMatrixPosition(robotTransformation)
-                        transformGizmo.mesh.rotation.setFromRotationMatrix(robotTransformation)
+                        transformGizmo.obj.position.setFromMatrixPosition(robotTransformation)
+                        transformGizmo.obj.rotation.setFromRotationMatrix(robotTransformation)
```
HunterBarclay: Why not build a function inside of GizmoSceneObject to set these? Simply pass in either a matrix or the individual components (translation, rotation, scale) and go from there. Same with getters.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -49,6 +47,8 @@ class MirabufSceneObject extends SceneObject {
     private _ejectable?: EjectableSceneObject
     private _scoringZones: ScoringZoneSceneObject[] = []
 
+    private _gizmo: GizmoSceneObject | undefined
```
HunterBarclay: Since we've decided that only once gizmo will exist, this should be some kind of static either in the SceneRenderer or GizmoSceneObject itself.

HunterBarclay (changes requested):
In addition to my comment of having the gizmo be handled globally rather than locally with each mirabuf, you should still add the actual react component that can be used to handle the gizmo, and it to the panel for spawning and additional moving.

Take how gizmos are used in zone configuration, make it a react component, use it **everywhere** we use transform gizmos, **including** when spawning in an asset.

fission/src/mirabuf/MirabufSceneObject.ts
BrandonPacewic: I can think of a case where we would want more than one gizmo in a scene at a time. For instance, what about placing more than one robot during the initialization of match mode?

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+        this._gizmo.addEventListener("dragging-changed", (event: { target: TransformControls; value: unknown }) => {
+            // disable orbit controls when dragging the transform gizmo
+            const gizmoDragging = World.SceneRenderer.IsAnyGizmoDragging()
+            if (!event.value && !gizmoDragging) {
+                World.SceneRenderer.orbitControls.enabled = true // enable orbit controls when not dragging another transform gizmo
+            } else if (!event.value && gizmoDragging) {
+                World.SceneRenderer.orbitControls.enabled = false // disable orbit controls when dragging another transform gizmo
+            } else {
+                World.SceneRenderer.orbitControls.enabled = !event.value // disable orbit controls when dragging transform gizmo
+            }
```
PepperLola:
```suggestion
            World.SceneRenderer.orbitControls.enabled = !event.value && !gizmoDragging
```

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+                // disable other gizmos when translating
+                Array.from(World.SceneRenderer.sceneObjects.values())
+                    .filter(obj => obj instanceof GizmoSceneObject)
+                    .forEach(obj => {
+                        if (obj.gizmo.object === event.target.object && obj.gizmo.mode !== "translate") {
+                            obj.gizmo.dragging = false
+                            obj.gizmo.enabled = !event.value
+                            return
+                        }
+                    })
```
PepperLola: Since you have the `SceneRenderer._gizmosOnMirabuf` map, you should no longer filter `sceneObjects`. In this case, you should make it so that you either make `_gizmosOnMirabuf` publicly available or have a method in SceneRenderer to test if some object has a gizmo (which would be as simple as `this._gizmosOnMirabuf.has(mirabuf_id)` from inside SceneRenderer).

As long as the mirabuf instance id that would be the key in `gizmosOnMirabuf` is available inside this function that I'm commenting on, you can then reduce this from O(n) to O(probably 1) since it would only require a Map lookup instead of an array filter and then foreach.

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+                        if (
+                            obj.gizmo.mode === "scale" &&
+                            event.target !== obj.gizmo &&
+                            obj.gizmo.object === event.target.object
+                        ) {
+                            obj.gizmo.dragging = false
+                            obj.gizmo.enabled = !event.value
+                            return
+                        }
+                    })
```
PepperLola: Same thing here - use the `_gizmosOnMirabuf` map.

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+        })
+    }
+
+    public Update(): void {
+        this._gizmo.updateMatrixWorld()
+
+        // updating the size of the gizmo based on the distance from the camera
+        const mainCameraFovRadians = (Math.PI * (this._mainCamera.fov * 0.5)) / 180
+        this._gizmo.setSize(
+            (this._size / this._mainCamera.position.distanceTo(this.gizmo.object!.position)) *
```
PepperLola, line 162: Are we sure object is always going to be non-null?

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+                    this.UpdateBodyPositionAndRotation(rn)
+                    this._parentObject?.UpdateNodeParts(rn, this.obj.matrix)
+                })
+            }
+        }
+
+        // creating enter key and escape key event listeners
+        if (InputSystem.isKeyPressed("Enter") && this._parentObject) {
+            // confirming placement of object
+            if (this._parentObject !== undefined) this._parentObject.EnablePhysics()
```
PepperLola:
```suggestion
            if (this._parentObject) this._parentObject.EnablePhysics()
```

I know `this._parentObject` can only be a `MirabufSceneObject` or `undefined` but I still prefer a truthiness check rather than checking if it's not exactly undefined. It doesn't really matter very much in this instance though.

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+import { Object3D, PerspectiveCamera } from "three"
+import { ThreeMatrix4_JoltMat44, ThreeQuaternion_JoltQuat } from "@/util/TypeConversions"
+import { RigidNodeReadOnly } from "@/mirabuf/MirabufParser"
+
+export type GizmoMode = "translate" | "rotate" | "scale"
+
+class GizmoSceneObject extends SceneObject {
+    private _gizmo: TransformControls
+    private _obj: Object3D
+    private _parentObject: MirabufSceneObject | undefined
```
PepperLola:
```suggestion
    private _parentObject?: MirabufSceneObject
```

I prefer this just because it's cleaner and is also what you do in the constructor.

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+                            obj.gizmo.mode === "scale" &&
+                            event.target !== obj.gizmo &&
+                            obj.gizmo.object === event.target.object
+                        ) {
+                            obj.gizmo.dragging = false
+                            obj.gizmo.enabled = !event.value
+                            return
+                        }
+                    })
+            }
```
PepperLola: Would a switch be better here? I'm not sure if the if will be optimized, especially since you're checking inputs in one of the conditions.

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+            // confirming placement of object
+            if (this._parentObject !== undefined) this._parentObject.EnablePhysics()
+            World.SceneRenderer.RemoveSceneObject(this.id)
+            return
+        } else if (InputSystem.isKeyPressed("Escape") && this._parentObject) {
+            // cancelling the creation of the mirabuf scene object
+            World.SceneRenderer.RemoveSceneObject(this._parentObject.id)
+            World.SceneRenderer.RemoveSceneObject(this.id)
+            return
+        }
```
PepperLola: Should `Dispose` be called somewhere in here instead of one of these calls? I'm genuinely asking since I don't know the answer, it very well could be no.

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+    /** changes the mode of the gizmo */
+    public SetMode(mode: GizmoMode) {
+        this._gizmo.setMode(mode)
+    }
+
+    /** updates body position and rotation for each body from the parent mirabuf */
+    public UpdateBodyPositionAndRotation(rn: RigidNodeReadOnly) {
+        if (!this._parentObject) return
+        World.PhysicsSystem.SetBodyPosition(
+            this._parentObject.mechanism.GetBodyByNodeId(rn.id)!,
```
PepperLola: Are we certain that the result of `GetBodyByNodeId` will be defined? I'd definitely hope so, but would it be worth instead checking that value and then calling `World.PhysicsSystem.SetBodyPosition` only if it's defined?

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -0,0 +1,194 @@
...
+            ThreeMatrix4_JoltMat44(this._obj.matrix).GetTranslation()
+        )
+        World.PhysicsSystem.SetBodyRotation(
+            this._parentObject.mechanism.GetBodyByNodeId(rn.id)!,
```
PepperLola: Same here

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -236,13 +231,29 @@ class SceneRenderer extends WorldSystem {
         return id
     }
 
+    /** Registers gizmos that are attached to a parent mirabufsceneobject  */
+    public RegisterGizmoSceneObject(obj: GizmoSceneObject): number {
+        if (obj.HasParent()) this._gizmosOnMirabuf.set(obj.parentObjectId!, obj)
```
PepperLola, line 243: Will `obj.parentObjectId` necessarily be defined?

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -236,13 +231,29 @@ class SceneRenderer extends WorldSystem {
...
         const obj = this._sceneObjects.get(id)
+
+        // If the object is a mirabuf object, remove the gizmo as well
+        if (obj instanceof MirabufSceneObject) {
+            const objGizmo = this._gizmosOnMirabuf.get(id)
+            if (this._gizmosOnMirabuf.delete(id)) objGizmo!.Dispose()
+        } else if (obj instanceof GizmoSceneObject && obj.HasParent()) {
+            this._gizmosOnMirabuf.delete(obj.parentObjectId!)
+        }
+
```
PepperLola, line 263: Will these necessarily be defined?

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -236,13 +231,29 @@ class SceneRenderer extends WorldSystem {
...
         if (this._sceneObjects.delete(id)) {
             obj!.Dispose()
```
PepperLola, line 265: Or `obj`?

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -328,83 +339,11 @@ class SceneRenderer extends WorldSystem {
...
-                this._scene.remove(tc)
-                this._transformControls.delete(tc)
-            }
-        })
+    /** returns whether any gizmos are being currently dragged */
+    public IsAnyGizmoDragging(): boolean {
+        return Array.from(this._sceneObjects.values())
+            .filter(obj => obj instanceof GizmoSceneObject)
+            .some(obj => obj.gizmo.dragging)
     }
```
PepperLola:
```suggestion
    public IsAnyGizmoDragging(): boolean {
        return Array.from(this._gizmosOnMirabuf.values())
            .some(gizmo => gizmo.dragging)
    }
```

I'm not sure if I got the syntax/names right but you can skip filtering since you're already keeping a map of active gizmos.

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
```diff
@@ -217,8 +219,8 @@ const ConfigureGamepiecePickupInterface: React.FC<ConfigPickupProps> = ({ select
...
```
PepperLola:
```suggestion
                        transformGizmo.UpdateGizmoObjectPositionAndRotation(robotTransformation)
```

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
```diff
@@ -107,7 +112,7 @@ const ConfigureGamepiecePickupInterface: React.FC<ConfigPickupProps> = ({ select
             return
         }
 
-        transformGizmo.mesh.scale.set(zoneSize, zoneSize, zoneSize)
+        transformGizmo.obj.scale.set(zoneSize, zoneSize, zoneSize)
```
PepperLola: I don't like that we're manipulating internal gizmo objects from outside of it. Maybe there should be a method to set the scale of the gizmo from out here? That way if the implementation of the gizmo changes, we would only need to make sure all the methods inside the class still work as expected, rather than needing to track down all the places where we made assumptions like this about the internal implementation of the class.

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
```diff
@@ -117,16 +122,16 @@ const ConfigureGamepiecePickupInterface: React.FC<ConfigPickupProps> = ({ select
...
-            )
+            ),
+            "translate",
+            1.5
         )
 
-        ;(gizmo.mesh.material as THREE.Material).depthTest = false
-        gizmo.AddMeshToScene()
-        gizmo.CreateGizmo("translate", 1.5)
+        ;((gizmo.obj as THREE.Mesh).material as THREE.Material).depthTest = false
```
PepperLola: Same thing here. We could avoid the whole ESLint vs Prettier disagreement about lines starting with a semicolon as well as the double cast and generally ugly line by creating a function to set the `depthTest` field of the material. Also, is this something that needs to happen every time the gizmo is created? If so, it would be worth putting it inside of some function in the gizmo object so that we don't need to remember to call it every time.

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureShotTrajectoryInterface.tsx
```diff
@@ -108,21 +108,17 @@ const ConfigureShotTrajectoryInterface: React.FC<ConfigEjectorProps> = ({ select
...
+            new THREE.ConeGeometry(0.1, 0.4, 4).rotateX(Math.PI / 2.0).translate(0, 0, 0.2),
+            World.SceneRenderer.CreateToonMaterial(ReactRgbaColor_ThreeColor(theme.HighlightSelect.color))
         )
 
-        ;(gizmo.mesh.material as THREE.Material).depthTest = false
-        gizmo.AddMeshToScene()
-        gizmo.CreateGizmo("translate", 1.5)
-        gizmo.CreateGizmo("rotate", 2.0)
+        if (!mesh) return
+        ;(mesh.material as THREE.Material).depthTest = false
```
PepperLola:
```suggestion
        const material: THREE.Material = mesh.material
        material.depthTest = false
```

I like the look of this better but it doesn't really matter.

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureShotTrajectoryInterface.tsx
```diff
@@ -208,12 +208,12 @@ const ConfigureShotTrajectoryInterface: React.FC<ConfigEjectorProps> = ({ select
...
                 onClick={() => {
-                    if (transformGizmo) {
+                    if (ejectorMesh) {
                         const robotTransformation = JoltMat44_ThreeMatrix4(
                             World.PhysicsSystem.GetBody(selectedRobot.GetRootNodeId()!).GetWorldTransform()
                         )
-                        transformGizmo.mesh.position.setFromMatrixPosition(robotTransformation)
-                        transformGizmo.mesh.rotation.setFromRotationMatrix(robotTransformation)
+                        ejectorMesh.position.setFromMatrixPosition(robotTransformation)
+                        ejectorMesh.rotation.setFromRotationMatrix(robotTransformation)
```
PepperLola: There's a lot of repetition of these two functions. Would it be worth creating a function to do that for you? There's already one in the gizmo class but that only works if the mesh is stored in the gizmo.

fission/src/ui/panels/configuring/assembly-config/interfaces/scoring/ZoneConfigInterface.tsx
```diff
@@ -71,7 +71,7 @@ function save(
     const translation = new THREE.Vector3(0, 0, 0)
     const rotation = new THREE.Quaternion(0, 0, 0, 1)
     const scale = new THREE.Vector3(1, 1, 1)
-    gizmo.mesh.matrixWorld.decompose(translation, rotation, scale)
+    gizmo.obj.matrixWorld.decompose(translation, rotation, scale)
```
PepperLola, line 73: I still don't like that we're doing this from outside the gizmo.

fission/src/ui/panels/configuring/assembly-config/interfaces/scoring/ZoneConfigInterface.tsx
```diff
@@ -170,13 +170,13 @@ const ZoneConfigInterface: React.FC<ZoneConfigProps> = ({ selectedField, selecte
...
+        const gizmo = new GizmoSceneObject(
+            new THREE.Mesh(new THREE.BoxGeometry(1, 1, 1), zone.alliance == "blue" ? blueMaterial : redMaterial),
+            "translate",
+            1.5
         )
 
-        ;(gizmo.mesh.material as THREE.Material).depthTest = false
-        gizmo.AddMeshToScene()
-        gizmo.CreateGizmo("translate", 1.5)
+        ;((gizmo.obj as THREE.Mesh).material as THREE.Material).depthTest = false
```
PepperLola: Same as above

fission/src/ui/panels/configuring/assembly-config/interfaces/scoring/ZoneConfigInterface.tsx
```diff
@@ -190,14 +190,14 @@ const ZoneConfigInterface: React.FC<ZoneConfigProps> = ({ selectedField, selecte
         const fieldTransformation = JoltMat44_ThreeMatrix4(World.PhysicsSystem.GetBody(nodeBodyId).GetWorldTransform())
         const props = DeltaFieldTransforms_VisualProperties(deltaTransformation, fieldTransformation)
 
-        gizmo.mesh.position.set(props.translation.x, props.translation.y, props.translation.z)
-        gizmo.mesh.rotation.setFromQuaternion(props.rotation)
-        gizmo.mesh.scale.set(props.scale.x, props.scale.y, props.scale.z)
+        gizmo.obj.position.set(props.translation.x, props.translation.y, props.translation.z)
+        gizmo.obj.rotation.setFromQuaternion(props.rotation)
+        gizmo.obj.scale.set(props.scale.x, props.scale.y, props.scale.z)
```
PepperLola: I don't think we should be directly mutating properties of the gizmo from outside.

PepperLola (changes requested):
For any of the non-null assertions, if you're absolutely certain that they're valid and will never cause issues, please respond with your justification and then resolve the comment. There are also some stylistic suggestions that I made that you don't need to accept if you or other people prefer how they are now.

fission/src/systems/scene/SceneRenderer.ts
HunterBarclay: `obj.HasParent()` essentially checks `obj.parent != undefined`.

HunterBarclay: I suppose you could technically add an unregistered object to the gizmo

fission/src/systems/scene/SceneRenderer.ts
HunterBarclay: Again, see the earlier comment.

fission/src/systems/scene/SceneRenderer.ts
HunterBarclay: If delete succeeds, then `obj` would exist

fission/src/ui/panels/configuring/assembly-config/interfaces/scoring/ZoneConfigInterface.tsx
HunterBarclay: At the end of the day, whatever uses the gizmo needs to have the freedom to do whatever it needs with the transform data it provides. I don't think there will ever be a way around this purely.

fission/src/mirabuf/MirabufSceneObject.ts
HunterBarclay: Honestly, I think for initial robot placement, we should move away from using the gizmo tool altogether and move towards Luca's placement for bots for match mode in v6 where you can click and place. In which case, gizmo controls should be reserved for fine movement/orientation/scale control.

HunterBarclay (approved):

PR #1086, opened 2024-08-08
Add `Barclay/JoltPhysics.js` submodule

BrandonPacewic:
We still have `mirabuf` as a submodule even though we don't really need it anymore. It does however make the repo look cooler and just that much more complex. So why not also add Hunter's fork of jolt that we use? We don't really need it here, but it looks cool.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "mirabuf"]
 	path = mirabuf
 	url = https://github.com/HiceS/mirabuf.git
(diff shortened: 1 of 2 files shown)


azaleacolburn:
I fully support this PR being merged

azaleacolburn (approved):
I fully support this PR being merged

Dhruv-0-Arora (approved):
If Azalea supports then I do too

a-crowell (approved):
Then, I do three.

PR #1088, opened 2024-08-09
Track Compiled Proto Files

BrandonPacewic:
Just one less step in getting the exporter up and running. We also need these files tracked to keep the formatter consistent between developer computers and the github action bot. Also also, this will be needed for adding type checking with `mypy` which I am currently working on.


HunterBarclay (commented):
Should we update the README to match the current setup instructions. It still mentions the init script.

HunterBarclay:
@BrandonPacewic When running the addin, it pops up with the dependency installation window, then it says the dependencies have been install successfully, and I need to restart. I restart, and then it repeats. It's done this 4 times in a row now. For context, I git cleaned my repo.

HunterBarclay (approved):

Dhruv-0-Arora (approved):
sublime

PR #1092, opened 2024-08-13
File Save Permission Detection `[AARD-1724]`

BrandonPacewic:
Fixed a bug that would cause the exporter to crash upon the selection of an export location that was readonly. Now will re-prompt for a different directory if this happens.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1724)


azaleacolburn (approved):

PepperLola (approved):
Very nice 💯

PR #1093, opened 2024-08-13
Exporter Typing with `mypy` `[AARD-1765]`

BrandonPacewic:
"Fixed" 400+ type errors...

In addition to the type error fixes:
- Added a `mypy` typing workflow.
- Added `.pyi` typing files for our compiled proto files for lsp and intellisense.
- Fixed an issue with the [`GlobalManager`](https://github.com/Autodesk/synthesis/blob/5807353608c7f9bd7ac9f186b5f7dca242b54db1/exporter/SynthesisFusionAddin/src/GlobalManager.py#L12-L62) where the internal class instance was the only thing ever being referenced.
https://github.com/Autodesk/synthesis/blob/5807353608c7f9bd7ac9f186b5f7dca242b54db1/exporter/SynthesisFusionAddin/src/GlobalManager.py#L56
  - Essentially because of the return type of the [`__new__`](https://github.com/Autodesk/synthesis/blob/5807353608c7f9bd7ac9f186b5f7dca242b54db1/exporter/SynthesisFusionAddin/src/GlobalManager.py#L52-L56) method the class was simply not working how it was originally designed.
- Made note of several things todo in future refactors including:
  - [`OString`](https://github.com/Autodesk/synthesis/blob/5807353608c7f9bd7ac9f186b5f7dca242b54db1/exporter/SynthesisFusionAddin/src/Types.py#L123-L311) needing more defined behaviour instead of revolving around the python 2 `object` type.
https://github.com/Autodesk/synthesis/blob/5807353608c7f9bd7ac9f186b5f7dca242b54db1/exporter/SynthesisFusionAddin/src/Types.py#L281
  - `JointHierarchy.py`'s class structure being poorly defined.
    - I resolved the apparent errors that `mypy` pointed out but the whole structure needs to be rewritten, particularly when it comes to how the various type of nodes are currently being used, many of which I'm sure can just be removed.
    - I also fixed some typing errors that seemingly looked like syntax errors related to the Fusion API and other functions within the module.
- Provided `default_factory`s for `ExporterOptions`' `list` default `field`s.
- Replaced default values for `ExporterOptions` `field`s of `None` where appropriate. 
- Noted various instances of unused code that, given the observed type errors, would never run even if we wanted to use it.
  - This code should be removed in the future as a part of separate refactors.
  - See included `Transition: AARD-1765` comment tags for references to individual code sections.
- Fixed an issue within [`construct_info`](https://github.com/Autodesk/synthesis/blob/5807353608c7f9bd7ac9f186b5f7dca242b54db1/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Utilities.py#L26-L59) that would cause `AttributeError`s to be thrown during exporting making debugging painful. https://github.com/Autodesk/synthesis/blob/5807353608c7f9bd7ac9f186b5f7dca242b54db1/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Utilities.py#L54-L59

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1765)


azaleacolburn (approved):
Amazing Work!

PepperLola (approved):
Good Looking 🌹

PR #1089, opened 2024-08-12
UI Continuity

LucaHaverty:
### Description
Cleaned up all of the UI and standardized it across the whole product.

### Objectives
- [x] Remove access to the theme editor (sorry Julian, it's very cool but we don't have the time to fix it at the moment 😢)
- [x] Tweak colors to be closer to what's on the figma
- [x] Rename everything to be less technical and more accessible to young robotics students
- [x] Add tooltips on settings that users found confusing or misunderstood
- [x] Make UI more dynamic for a more enjoyable user experience (hover effects, click effect, etc...)
- [x] Fix the bug causing a robots scheme to not be set if it was selected while the robot was importing
- [x] Fix the scroll zoom bug

> [!note]
> I am not a visual designer and I think there is still a lot of room for improvement, so please comment on anything you think should change, no matter how big or small, both in the UI design and user experience.

## Screenshots
(A little has changed since the screenshots were taken so definitely pull down the branch and mess around with it)

### Tooltips
image

### Settings Modal
Before | After
:-------------------------:|:-------------------------:
image  |  image

### Spawn Asset Panel (demonstration of the new colors)
image

### Dynamic UI Demo
https://github.com/user-attachments/assets/3d701e3c-88df-4fbd-a32e-9cab5b4f859f

### Questions for reviewers:

> [!Important] 
Feedback I'm looking for:
> - I'm not a big fan of what I did with the accordion button, let me know if you have better ideas!
> - Do you all think the name "configure assets" and "general settings" will make sense to users lacking technical experience?
> - Do we want to keep debug tools in the build that is released? If so, we might want to add info tooltips to all of the options.
> - Should we keep sequential joints as a separate configuration interface than configuring motor speeds and torques or merge them?
> - Any suggestions for the intake, ejector, and scoring zone config panels?

[Control scheme bug Jira issue](https://jira.autodesk.com/browse/AARD-1799)
[UI Fixes Jira issue](https://jira.autodesk.com/browse/AARD-1794)
[Info Buttons Jira issue](https://jira.autodesk.com/browse/AARD-1809)
[Scroll Zoom Bug Jira issue](https://jira.autodesk.com/browse/AARD-1801)

PR #1097, opened 2024-08-19
Move subsystem config to the combined configuration panel

LucaHaverty:
### Description
Added all of the subsystem config functionality to the combined configuration panel.

> [!warning]
> Merge after #1089 

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1849)


a-crowell (changes requested):
I like how you made each joint separate and moved inversion to each one. That makes a lot of sense.

However, the combination with the intake and ejector isn't very intuitive. It makes me realize Joints or Motors might be a better term for the joints because the armavator is the intake/ejector subsystem. This could look like intake/ejector being lumped into Gamepiece Manipulation > (Intake / Ejector) && Subsystems (Drivetrain / Arm / Elevator) or something like Intake && Ejector && Joints.

Having the inversion out makes the Sequential Joints much more intuitive too! Great work.

a-crowell (approved):
Works great!

BrandonPacewic (approved):
LGTM!

PR #1101, opened 2024-08-20
Utilize Fusion's Unit Manager `[AARD-1773]`

BrandonPacewic:
Instead of keeping track of a separate option for weight units, the Synthesis Exporter now will use the default Fusion units of a given design. For example: If you have inches (`in`) as your selected distance unit, the exporter will use pounds (`lbs`) for it's weight unit. And vice-versa with things like millimeters (`mm`) and centimeters (`cm`) for using kilograms (`kg`).

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1773)


azaleacolburn (approved):
LGTM

Dhruv-0-Arora (approved):
Exceptional 🥪

PR #1050, opened 2024-07-26
CAN Support

azaleacolburn:
### Description
Send, receive, and use CAN device data between a robot project and fission.

### Objectives
- [x] Send TalonFX Data
- [x] Send CANSparkMax Data
- [x] Make simulated motor fields appear as CAN devices rather than SimDevices
- [x] CAN Motor Brain Mappings
- [x] CANEncoder support

### Testing Done
- [x] Test TalonFX in WS Viewer
- [x] Test CANSparkMax in WS Viewer
- [x] Make sure CAN Motor inputs map into the brain properly
- [x] Test CANEncoders on SparkMaxes

For testing:
1. Build and publish SyntheSim with `./gradlew build` and `./gradlew publishToMavenLocal`.
2. Build and run the JavaSample with `./gradlew build` and `./gradlew simulateJava`.
3. Run fission
4. Open fission, spawn a robot, then click the WS Test button in the debug menu (this must be done first).
5. Verify the connection by opening the WS Viewer panel, you should see a long list of motors and encoders.
6. Click the RoboRio button to setup CAN output group, select all the CANSparkMax motors and all the WheelDrivers, then click Done
7. Do the same but for Encoders, this time, just select CANSparkMax[1] and any of the wheel stimuli.
8. In the robotSimulation screen, click Autonomous, you should see the robot drive forward 20 (feet?), then stop.

We need to have a discussion about how we want device mapping to work, right now every motor in a group's output is averaged and applied to each driver. We need to make it easier to do one to one mappings between motors and drivers, because right now, clicking through the whole menu ~10 times to make an output group for each motor is ridiculous.

> [!note]
> Pulled from #1077 for it's encoder framework, so I think just turning that PR into a gyro / joystick PR is the way to go.

> [!note]
> CANMotor inputs (eg. supplyCurrent) have been ignored, since they're not simulated in fission.

This PR is a combination of JIRA Issues [1747](https://jira.autodesk.com/browse/AARD-1747), [1748](https://jira.autodesk.com/browse/AARD-1748), and [1749](https://jira.autodesk.com/browse/AARD-1749) for a total of 18 story points.

PR #1103, opened 2024-08-21
More Simulation Support Pt.2

PepperLola:
### Description
Added support for more of the simulation devices supported by WPILib. Also fixed issue with resetting the brain of the robot every time a PWM Output Group was created so it's now possible to have more than one output group.

The aim of this PR is to add support for the most important items in the [Hardware Messages](https://github.com/wpilibsuite/allwpilib/blob/main/simulation/halsim_ws_core/doc/hardware_ws_api.md#hardware-messages) section of the WPILib Hardware WS API docs. This does not include CAN motor support, which is being handled in #1050. All planned devices are listed in [Objectives](#objectives) and are what I deem to be most important and worthwhile right now, but these could be subject to change and the list is open to suggestions.

> [!NOTE]
> It's worth noting that controls are managed entirely by the HALSim GUI. The available USB devices are assigned to ports through the GUI, which are then read and fed to the robot code. The robot code sets motor outputs depending on the controller input, and the motor values are what's actually sent to fission.

### <a name="objectives"></a>Objectives
- [x] Encoder support
- [x] Joystick support through Sim GUI
- [x] Gyroscope and Accelerometer support
- [x] Digital and analog sensor support

### Testing Done
- Test driving robot with two different PWM motors
  -  Ensure encoders update appropriately

> [!IMPORTANT]
> In order for a controller to be appropriately mapped in the Sim GUI, you must enable the checkbox called "Map gamepad" in the Joystick menu. However, this does not appear when running the application on MacOS.
>
> When unchecked, the joystick axes may not be accurate (e.g. the right joystick and L/R trigger axes may be swapped).

When testing, you can uncomment the commented out devices in `WPILibBrain.ts`. Make sure the inputs are updating appropriately in the simulator. You can see how the inputs are meant to update in in those lines you uncomment, and how the outputs are meant to update in the sample robot code.
https://github.com/Autodesk/synthesis/blob/18cd843fd0a9cdcd5f2aeda74cca8e5b0d87f9ba/fission/src/systems/simulation/wpilib_brain/WPILibBrain.ts#L382-L387

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1766)


HunterBarclay (approved):
Good progress on Simulation. I'll be opening another branch to start working on the UI.

PR #1099, opened 2024-08-20
Catch When No Design is Selected

azaleacolburn:
### Description
Add catch and messagebox when no design is open in the exporter.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1851)

PR #1107, opened 2024-08-22
The Dreaded `ConfigCommand.py` Refactor `[AARD-1764]`

BrandonPacewic:
I am happy to report that this file is now less than 240 lines. 😊 

- Set a default size for the main config panel:
  - This was done to solve an issue with the joint and gamepiece config tables not showing up correctly.
- Removed unused handler code.
- Remove unused code within `ConfigCommand.py`.
- Removed preselect handler as it simply was not working as originally intended.
- Updated the help icon redirect to be directly to the exporter codelab.
- Changed the default location for export to be local download.
- Moved handler creation to the top of create:
  - Results in more stable behaviour if we crash later when adding joints, wheels and gamepieces.
- Added `PersistentEventHandler` class that will auto add a handler to the global list to keep it from falling out of scope.
- Added better typing with `adsk.fusion.Design.cast(adsk.core.Application.get().activeProduct)` for the design getter call.
- Fixed a persistent issue from #1099.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1764)


PepperLola (approved):
I like the default export location change. Nice job getting it down to so few lines 🥇 

One thing I noticed is that it no longer includes the version number in the exported file name. Not sure if this was intentional and I don't mind either way, just something to note.

BrandonPacewic (commented):
> One thing I noticed is that it no longer includes the version number in the exported file name. Not sure if this was intentional and I don't mind either way, just something to note.

Unfortunately, this was not intended. 🫤

exporter/SynthesisFusionAddin/src/UI/ConfigCommand.py
```diff
@@ -231,111 +114,41 @@ def notify(self, args: adsk.core.CommandCreatedEventArgs) -> None:
...
+    def notify(self, _: adsk.core.CommandEventArgs) -> None:
+        design = adsk.fusion.Design.cast(adsk.core.Application.get().activeProduct)
         exporterOptions = ExporterOptions().readFromDesign()
 
-        if args.executeFailed:
-            logger.error("Could not execute configuration due to failure")
-            return
+        designNameSplit = design.rootComponent.name.split(" ")
+        if len(designNameSplit) > 1:
+            docName, docVersion = designNameSplit[:2]
```
PepperLola: It looks like this line doesn't work properly with multi-word document names. When I try exporting the utility knife, `docName` becomes "Utility" and `docVersion` becomes "Knife".

BrandonPacewic: 😬 Nice catch.

PepperLola (approved):
Still works and includes the full name and version this time. I will say that previously it would only concatenate the name and version with an underscore and the rest would be the raw name with spaces (i.e. `Utility Knife_v2.mira` instead of the `Utility_Knife_v2.mira` that happens now). I don't care either way and this is definitely less important than including the version number and full name.

BrandonPacewic (commented):
> I will say that previously it would only concatenate the name and version with an underscore and the rest would be the raw name with spaces (i.e. `Utility Knife_v2.mira` instead of the `Utility_Knife_v2.mira` that happens now). I don't care either way and this is definitely less important than including the version number and full name.

Yea this was on purpose. Was simpler to write it this way and I figured it was fine to replace all spaces with underscores.

azaleacolburn (approved):
This is great, thanks for all your hard work!

Dhruv-0-Arora (commented):
This might be an issue

Screenshot 2025-06-19 at 10 43 26 AM

Dhruv-0-Arora (approved):

PR #1108, opened 2024-08-30
[BUG] Vite Server Crashing Chrome

HunterBarclay:
### Description
Vite server has been crashing Google Chrome since the introduction of the new configuration panel. It appears that the configuration files had a number of circular dependencies that either couldn't be resolved by Google Chrome's debugger or ran into a vulnerability it Google Chrome's debugger that causes a crash.

### Solution
I move a handful of share objects between the configuration panel and it's interfaces into separate classes to remedy the circular dependencies

### Other Notes
I intentionally created circular dependencies to test my theory that simply circular dependencies were the issue, but they were resolved without issue. This makes me believe that either the mildly cursed work arounds we are using in React or the specific complexity of the circular dependencies were what caused the crash. Other than that, I'm not entirely sure what caused the crash or how we can directly avoid going forward, other than the obviously of having no circular dependencies.

[JIRA Issue - 1859](https://jira.autodesk.com/browse/AARD-1859)


BrandonPacewic (approved):
Weird problem. Nice fix. 🥇

PepperLola (approved):
This PR prevented me from defenestrating my computer.

PR #1115, opened 2025-01-04
HALSim websocket server only listens on localhost

virtuald:
I personally haven't tried this (so you should verify before merging!) but the websocket server only listens on localhost, so that environment variable specified in build.gradle serves no purpose (it's a client setting, not a server setting).

Reference: https://github.com/wpilibsuite/allwpilib/blob/b65f159c3fdfdfce9ec126cf79dfd9a16d1e66e9/simulation/halsim_ws_server/src/main/native/cpp/HALSimWeb.cpp#L120

### DIFF
--- a/simulation/samples/JavaAutoSample/build.gradle
+++ b/simulation/samples/JavaAutoSample/build.gradle
@@ -91,7 +91,6 @@ test {
 wpi.sim.addGui().defaultEnabled = true
 wpi.sim.addDriverstation()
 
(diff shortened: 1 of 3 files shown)


HunterBarclay:
@virtuald 
Thanks for the catch. It looks like that's left over from native Synthesis where we actually handled the server and the user used the client extension. I have a [PR](https://github.com/virtuald/synthesis/pull/1) ready for your branch which removes the use of this environment variable from our sample projects.

PR #1116, opened 2025-01-04
Add RobotPy instructions

virtuald:
I haven't actually tried a robotpy project with synthesis, but I have verified that running these instructions will start a robot program with a websocket server.


HunterBarclay (approved):
Looks great, thanks for including this. I've never actually messed with RobotPy before so I might take a bit and start working on the python SyntheSim.

PR #1126, opened 2025-06-17
New Default Scene and Logo

itspvty:
Modified the baseplate (ground) that is created on page load to be white instead of blue and to render the Autodesk logo on the top (+Y) face. The icon-only logo used was added to assets.

image


Dhruv-0-Arora (changes requested):
We floating 🙌
Maybe make the thickness the same as before to remove this or talk to Azalea to start on fixing suspension and physics

image

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -450,6 +485,7 @@ class SceneRenderer extends WorldSystem {
 
         ContextSupplierEvent.Dispatch(miraSupplierData, e.position)
     }
+
```
Dhruv-0-Arora:
```suggestion
```

Dhruv-0-Arora (approved):
LGTM

BrandonPacewic (changes requested):
It appears that robots are clipping into the ground. Additionally, parts of the ground have no collision.

image

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -98,14 +101,45 @@ class SceneRenderer extends WorldSystem {
...
+            logoMaterial, // Logo on top face only
+            solidMaterial,
+            solidMaterial,
+            solidMaterial,
+        ]        
+        
+        // Set up materials for CSM
+        materials.forEach(material => {
+            if (this._light instanceof CSM) this._light.setupMaterial(material)
+        })
```
Dhruv-0-Arora: This is redundant since the CreateToonMaterial function already runs the setupMaterial function. You only need to run the logoMaterial through the setupMaterial function

Dhruv-0-Arora (changes requested):

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -98,14 +101,40 @@ class SceneRenderer extends WorldSystem {
         this._renderer.shadowMap.type = THREE.PCFSoftShadowMap
         this._renderer.setSize(window.innerWidth, window.innerHeight)
 
-        // Adding the lighting uisng quality preferences
+        // Adding the lighting using quality preferences
```
BrandonPacewic, line 104: What a lovely typo 🙃

fission/src/systems/scene/SceneRenderer.ts
BrandonPacewic: Small catch but I think its important enough.

BrandonPacewic (approved):
Cool looks good to me 🦜

fission/src/systems/scene/SceneRenderer.ts
Dhruv-0-Arora: Whoops 😬

Dhruv-0-Arora (changes requested):
When materials are created, they need to be set up with the cascading shadow map otherwise there are some artifacts of the shadows

image

fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -98,14 +101,40 @@ class SceneRenderer extends WorldSystem {
...
+        logoTexture.wrapT = THREE.ClampToEdgeWrapping
+        logoTexture.center.set(0.5, 0.5) // Size Adjustment
+        logoTexture.repeat.set(2, 2)
+
+        const logoMaterial = new THREE.MeshToonMaterial({
+            map: logoTexture,
+            color: GROUND_COLOR,
+            shadowSide: THREE.DoubleSide,
+        })
+
```
Dhruv-0-Arora:
```suggestion
        })
        if (this._light instanceof CSM) this._light.setupMaterial(logoMaterial)

```

Dhruv-0-Arora (approved):

PR #1127, opened 2025-06-17
Delete old field on import

AlexD717:
### Description
When spawning a new game field all other game fields are removed, but robots are kept

### Testing Done
- [ ] Spawned field with no fields spawned yet
- [ ] Spawned field with a field already spawned
- [ ] Spawned field with a robot already spawned
- [ ] Spawned field with a robot and field already spawned

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1876)
Jira Issue 1876


fission/src/systems/scene/SceneRenderer.ts
```diff
@@ -302,6 +303,16 @@ class SceneRenderer extends WorldSystem {
         }
     }
 
+    public RemoveAllFields() {
+        for (const [id, obj] of this._sceneObjects) {
+            if (obj instanceof MirabufSceneObject) {
+                if (obj.miraType == MiraType.FIELD) {
+                    this.RemoveSceneObject(id)
+                }
+            }
```
Dhruv-0-Arora:
```suggestion
            if (obj instanceof MirabufSceneObject && obj.miraType == MiraType.FIELD) {
                this.RemoveSceneObject(id)
            }
```

Dhruv-0-Arora (changes requested):
@BrandonPacewic @PepperLola What are your thoughts on creating a confirmation panel notifying the user that it will remove all the fields

fission/src/systems/scene/SceneRenderer.ts
Dhruv-0-Arora: Note: Brandon thinks it is a good idea

Dhruv-0-Arora:
Note: Previous commit doubles the timeout time for unit tests to 10,000 ms

AlexD717:
> @BrandonPacewic @PepperLola What are your thoughts on creating a confirmation panel notifying the user that it will remove all the fields

Conformation panel is not feasible with the current UI implementation, we will do a UI refactor later and then implement a conformation panel (Paraphrased from Julian)

Dhruv-0-Arora (approved):
LGTM

Dhruv-0-Arora (changes requested):
> **Note:** Need to revert all the formatting changes that I included.

Dhruv-0-Arora (approved):
Ready 👍

PepperLola (approved):

PR #1128, opened 2025-06-17
Main Menu Screen `[AARD-1878]`

rutmanz:
### Description
Adds a main menu that precedes world initialization and requires a choice of singleplayer or multiplayer. 

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1878)


BrandonPacewic (changes requested):
Think it might be better to use the toast system for the multiplayer alert (I know we are probably going to be changing it rather soon). Additionally some weird behaviour happens when you click outside the main menu modal when it first pops up.

Dhruv-0-Arora (changes requested):
Would it be possible for you to render it over the scene instead of a gray frame?

Screenshot 2025-06-17 at 5 27 02 PM

rutmanz:
The ticket specified that this menu should precede loading the scene, presumably so that multiplayer could have an entirely different initialization. We could load it anyway or do something else with the background. I agree it's not great to have blank

fission/src/ui/modals/MainMenuModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+                        startSingleplayerCallback()
+                    }}
+                    className="w-full my-1"
+                />
+                <Button
+                    value={"Multiplayer"}
+                    onClick={() => {
+                        // todo
+                        Global_AddToast?.("error", "Multiplayer Not Yet Supported", "Come back soon!")
+                        // closeModal()
```
BrandonPacewic: Should clean up this todo and leftover comment.

BrandonPacewic (changes requested):
Looks good. Sounds like we have a UI refactor coming so we are probably going to end up changing some stuff here.

Two comments I had:
1. We should be hiding the side panel since it doesn't do anything till you actually launch the sim.
2. Scoreboard should probably also be hidden.

Like I mentioned, UI refactor coming so we are probably going to change this anyways. Run the auto formatter and we can mark this as done.

BrandonPacewic (approved):
Notes for end of summer UI polish:

- Should consider the difference between launching into the sim and displaying a main menu.
- For parallel with fusion some form of instant sim access seems better.
- Should have a UX conversation about what we want the final user flow on launch to be.

Looks good to me!

AlexD717 (approved):

fission/src/ui/modals/MainMenuModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+                    onClick={() => {
+                        closeModal()
+                        startSingleplayerCallback()
+                    }}
+                    className="w-full my-1"
+                />
+                <Button
+                    value={"Multiplayer"}
+                    onClick={() => {
+                        Global_AddToast?.("error", "Multiplayer Not Yet Supported", "Come back soon!")
```
Dhruv-0-Arora:
```suggestion
                        Global_AddToast?.("error", "Not Supported", "Multiplayer is not yet supported. Come back soon!")
```

Dhruv-0-Arora: This is not necessary but I think it personally looks cleaner. Thoughts?

Dhruv-0-Arora (changes requested):

PepperLola (approved):
Looks good. I agree about hiding the other UI elements but that can happen after the UI refactor.

Dhruv-0-Arora (approved):

PR #1131, opened 2025-06-18
Keep camera focus on robot when field is loaded

itspvty:
The robot now keeps its assigned focus when a field asset is spawned.

After spawning robot:
image

After then spawning field:
image


package-lock.json
Dhruv-0-Arora, line 1: you might have accidentally created this package lock. Can you remove it?

Dhruv-0-Arora (changes requested):

Dhruv-0-Arora (approved):
Looks good

Dhruv-0-Arora:
Actually one change, instead of adding the package-lock to the git ignore, you might want to just delete it since this package-lock is in the root directory instead of the one in fission. We want to keep the one located in fission.

Dhruv-0-Arora (changes requested):
See above

Dhruv-0-Arora (approved):
lgtm

BrandonPacewic (approved):
Sweet 🎥

PR #1132, opened 2025-06-18
Open correct tab in spawn panel from configuration panel

rutmanz:
### Description
When opening the mirabuf import panel through the Configure Assets panel, it now opens to the import mode matching the configure mode. For example, when the Fields tab is selected in Configure Assets and the user clicks the plus button, the spawn panel opens to the fields tab instead of robots.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1891)


BrandonPacewic (approved):
Yes, indeed very good. Very good.

azaleacolburn (approved):
lgtm

PR #1137, opened 2025-06-18
Center camera on all axis on focus

itspvty:
When an object is created, it is now focused to the middle of the robot assembly as opposed to just its base. In other words, the automatic centering that occurs on focus now takes into account the Y axis.

Old:
image

New:
image


Dhruv-0-Arora (changes requested):
Could you merge dev?

Dhruv-0-Arora (commented):
Looks good to me. The only potential problem with this implementation is that since you are pivoting from a physics based approach, which is more precise, to a mesh based approach. With a weirdly exported robot it might cause some issues but I can't seem to find any with our current implementation.

Another note is that when you start trying to topple the robot, the camera will start shaking due to the bounding box implementation rather than the jolt physics based implementation. It doesn't seem to be too big of a problem though. Additionally, the camera moves if you begin to move the mechanisms/arms which could be unwanted yet again doesn't seem to be too big of an issue.

@BrandonPacewic @PepperLola thoughts?

Dhruv-0-Arora (approved):
Discussed; alternatively we could calculate the bounding box when the assembly is spawned and keep it constant but I personally think this looks better.

BrandonPacewic (approved):
> Looks good to me. The only potential problem with this implementation is that since you are pivoting from a physics based approach, which is more precise, to a mesh based approach. With a weirdly exported robot it might cause some issues but I can't seem to find any with our current implementation.
> 
> Another note is that when you start trying to topple the robot, the camera will start shaking due to the bounding box implementation rather than the jolt physics based implementation. It doesn't seem to be too big of a problem though. Additionally, the camera moves if you begin to move the mechanisms/arms which could be unwanted yet again doesn't seem to be too big of an issue.
> 
> @BrandonPacewic @PepperLola thoughts?

Maybe, but this is such a small change (with such an obvious benefit) that we should just make it and revisit later if we really feel like it could be significantly better than this. 

Good work, nice easy change 🛰️

PR #1139, opened 2025-06-18
Update Jolt Submodule Owner

azaleacolburn:
### Description
Use the Azalea-maintained Jolt repo.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1904)


BrandonPacewic (changes requested):
Unfortunately it's not this easy to update (It probably should be but it is what it is).

1. The submodule pointer itself needs to be updated in addition to updating the `.gitmodules` file:
  - This can be done by running:
```
git submodule sync                # update your local .git/config
git submodule foreach 'git fetch' # ensure the new remote is known
```
  - After that your git diff should look something like this:
```
diff --git a/JoltPhysics.js b/JoltPhysics.js
index a1b2c3d..d4e5f6g 160000
--- a/JoltPhysics.js
+++ b/JoltPhysics.js
@@ -1 +1 @@
-Subproject commit a1b2c3d4e5f6g7h8i9j0k
+Subproject commit d4e5f6g7h8i9j0ka1b2c3d
```
(I'm not exactly sure what it will look like since the folder name is different than the GitHub project name).

2. Since the submodule is just there as a reference (we don't actually build from it, its just there to look cool) we need to update all references to Hunter's `jolt-physics`. Such as here:
https://github.com/Autodesk/synthesis/blob/0e270b0a900ed48d98f094304e147811a8aa38b4/fission/package.json#L24
And here:
https://github.com/Autodesk/synthesis/blob/0e270b0a900ed48d98f094304e147811a8aa38b4/fission/src/systems/physics/PhysicsSystem.ts#L15
Just to name a few, there are *lots* of such occurrences. I'm not 100% sure if Hunter had to do something to make his project available to be referenced in this way, something else that will need to be confirmed.

3. I'm also going to request that we use https for the `.gitmodules` file. It's what we have done thus far and if someone doesn't have ssh set up I don't want this to be an additional part of dev environment setup that we need to worry about.

rutmanz:
Looks like Hunter had to put [his fork on NPM](https://www.npmjs.com/package/@barclah/jolt-physics)

Dhruv-0-Arora (approved):
Amazing

.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/HiceS/mirabuf.git
 [submodule "jolt"]
 	path = jolt
-	url = https://github.com/HunterBarclay/JoltPhysics.js.git
+	url = https://github.com:azaleacolburn/JoltPhysics.js.git
```
rutmanz:
```suggestion
	url = https://github.com/azaleacolburn/JoltPhysics.js.git
```

rutmanz (changes requested):

PepperLola (approved):
Looks good, I couldn't find any missed imports. Forgot I had to reinstall packages before running but after that everything worked perfectly.

azaleacolburn:
> Looks good, I couldn't find any missed imports. Forgot I had to reinstall packages before running but after that everything worked perfectly.

`rg barclah . --files-with-matches | xargs sed -i 's/barclah/azaleacolburn/g'` went hard lol

rutmanz (approved):

BrandonPacewic (approved):
🔗

PR #1141, opened 2025-06-19
Robot Import Cache Failure

ryanzhangofficial:
Avoids failing robot import if cache building fails. Return raw buffer wrapped in MirabufCacheInfo as a fallback.


BrandonPacewic (changes requested):
It appears that in this updated version of this function we are always downloading the robot and updating the cache, even if we already have this robot cached, which is not quite the exact behaviour we want. 

We want some form of "cache-first" logic, where we double check if the cache of the file is valid, then use that to avoid downloading it again.

Then, if we need to download the robot, do so, and then in that case if the cache again fails, return the robot that we just downloaded instead of doing nothing.

(I hope that makes sense)

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -118,24 +118,40 @@ class MirabufCachingService {
      * @returns {Promise<MirabufCacheInfo | undefined>} Promise with the result of the promise. Metadata on the mirabuf file if successful, undefined if not.
      */
     public static async CacheRemote(fetchLocation: string, miraType?: MiraType): Promise<MirabufCacheInfo | undefined> {
-        if (miraType) {
+        if (miraType !== undefined) {
             const map = MirabufCachingService.GetCacheMap(miraType)
-            const target = map[fetchLocation]
+            const hit = map[fetchLocation]
+            if (hit) return hit
```
Dhruv-0-Arora: I personally think we should keep this named `target`

Dhruv-0-Arora (changes requested):

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -118,24 +118,40 @@ class MirabufCachingService {
...
+            if (target) return target
+        }
+        try {
+            // grab file remote
+            const miraBuff = await fetch(
+                encodeURI(fetchLocation),
+                import.meta.env.DEV ? { cache: "no-store" } : undefined
+            )
+                .then(x => x.blob())
+                .then(x => x.arrayBuffer())
```
BrandonPacewic: Right now you assume `fetch(...)` always resolves to a 2xx. If the server returns 404 or 500, you’ll end up turning that error into a blob/arrayBuffer error later. We should throw in a quick:

```ts
if (!resp.ok) throw new Error(...)
```

Before we do that just to make sure we clearly fail on a bad response.

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -118,24 +118,40 @@ class MirabufCachingService {
...
+
+            if (cached) return cached
+
+            // fallback: return raw buffer wrapped in MirabufCacheInfo
+            return {
+                id: Date.now().toString(),
+                miraType: miraType ?? (this.AssemblyFromBuffer(miraBuff).dynamic ? MiraType.ROBOT : MiraType.FIELD),
+                cacheKey: fetchLocation,
+                buffer: miraBuff,
             }
```
BrandonPacewic, line 155: Can we add a error toast right before we get here? Some sort of error display so we know we reached the fallback.

BrandonPacewic (changes requested):
Awesome this is exactly what I had in mind. Few small changes and then we should be good to go.

Dhruv-0-Arora (commented):
Is there some way to reproduce this error to test it?

BrandonPacewic (commented):
> Is there some way to reproduce this error to test it?

This might actually be a good place to write a unit test 🤔 

@Dhruv-0-Arora For now just look at the code logic itself and compare it to the original. When stepping through the function it should be fairly obvious where the problem arises.

ryanzhangofficial:
Is there currently an obvious problem that I'm not seeing?

BrandonPacewic:
> Is there currently an obvious problem that I'm not seeing?

Nope your version looks good, I was referencing the original version of the code. Not the changes you have here.

Dhruv-0-Arora (approved):

BrandonPacewic (approved):
Awesome! Will add a unit test to this as a part of AARD-1940

PR #1142, opened 2025-06-19
Light Reposition to fix Low Graphics Shader bug

itspvty:
The directional light by ThreeJS is now positioned by taking the normalized light direction and multiplying by -20 (placing it 20 units away in the opposite direction from the origin), and now both lighting modes use the same direction vector (1.0, -3.0, -2.0) normalized. The light target remains at (0, 0, 0) to center the shadow camera on the ground plane.

image


azaleacolburn (approved):
Nice addition, lgtm

BrandonPacewic (approved):
Cool nice and easy 🌤️

PR #1143, opened 2025-06-19
Don't allow importing multiple assets at once `[AARD-1800]`

rutmanz:
### Description
Prevents importing a new asset when there is still a previous asset being configured. It does this by showing a toast and preventing the import mirabuf panel from being opened until the previous asset's position is confirmed.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1800)


fission/src/ui/panels/mirabuf/ImportMirabufPanel.tsx
```diff
@@ -162,7 +162,12 @@ const ImportMirabufPanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
         }
     }, [])
 
-    useEffect(() => {
+    useLayoutEffect(() => {
+        if (mirabufPanelState.hasUnconfirmedImport) {
+            closePanel("import-mirabuf")
+            Global_AddToast?.("warning", "You're already importing a model", "Confirm that one first")
```
PepperLola:
```suggestion
            Global_AddToast?.("warning", "You're already importing a model!", "Confirm that one before importing another.")
```

PepperLola (approved):
I just have a nitpick for the toast contents. Other than that it looks good and works well. For some reason trying to import another model and trigger the toast would cause the website to crash before I updated dependencies, but after that it worked.

BrandonPacewic (approved):
Committed @PepperLola's toast description change suggestion. Also changed the PR title to have the Jira issue at the end as to not clutter the title by reading reading order.

PR #1145, opened 2025-06-19
Created Drag Mode (previously god mode)

itspvty:
Added a drag mode system that allows users to move and manipulate all dynamic objects including robots and game pieces. The system intercepts mouse interactions and moves bodies via force-based dragging. The system automatically detects draggable objects and excludes static bodies and field structures, with special handling for complex objects like robots by temporarily disabling their physics during drag operations. The drag also includes velocity damping, distance-based speed control, and mass-proportional force application. A visual drag mode indicator appears when drag mode is active and camera controls are automatically disabled during drag operations. The functionality is accessible through the debug panel with toggle controls and toast notifications for mode status changes. These changes collectively represent the deprecation of "God Mode" and replaces it with "Drag Mode".

image

image

(Difficult to demonstrate actual behavior through screenshots)


fission/src/systems/scene/DragModeSystem.ts
```diff
@@ -0,0 +1,279 @@
...
+
+    public Destroy(): void {
+        this.enabled = false
+    }
+
+    private hookInteractionHandlers(): void {
+        const handler = World.SceneRenderer.renderer.domElement.parentElement?.querySelector("canvas")
+        if (!handler) return
+
+        const screenHandler = (World.SceneRenderer as any)._screenInteractionHandler
```
BrandonPacewic: Any chance we could specifiy a type here? ESLint *really* does not like `any`.

fission/src/systems/scene/DragModeSystem.ts
```diff
@@ -0,0 +1,279 @@
...
+        this._originalInteractionMove = screenHandler.interactionMove
+        this._originalInteractionEnd = screenHandler.interactionEnd
+
+        screenHandler.interactionStart = (interaction: InteractionStart) => this.onInteractionStart(interaction)
+        screenHandler.interactionMove = (interaction: InteractionMove) => this.onInteractionMove(interaction)
+        screenHandler.interactionEnd = (interaction: InteractionEnd) => this.onInteractionEnd(interaction)
+    }
+
+    private unhookInteractionHandlers(): void {
+        const screenHandler = (World.SceneRenderer as any)._screenInteractionHandler
```
BrandonPacewic: Any chance we could specifiy a type here? ESLint *really* does not like `any`.

rutmanz:
If you're going to have a button-shaped indicator that says "drag mode" it would make sense to have it be a button to deactivate drag mode (or look less like a button). It also is covered by the fps indicator, the bottom left might be a better place for it

rutmanz (changes requested):
If you're going to have a button-shaped indicator that says "drag mode" it would make sense to have it be a button to deactivate drag mode (or look less like a button). It also is covered by the fps indicator, the bottom left might be a better place for it

PepperLola (changes requested):
I think it would be nice if whichever element you're dragging stayed fixed to the mouse cursor, since right now it drifts the farther you drag it.

fission/src/ui/components/DragModeIndicator.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+        window.addEventListener("dragModeToggled", handleDragModeToggle as EventListener)
+
+        return () => {
+            window.removeEventListener("dragModeToggled", handleDragModeToggle as EventListener)
+        }
+    }, [])
+
+    const handleClick = () => {
+        window.dispatchEvent(new CustomEvent("disableDragMode"))
```
rutmanz, line 22: This doesn't seem to trigger the "disabled" toast, not a big deal but it should probably be consistent with disabling in the menu

rutmanz (changes requested):

rutmanz (approved):

BrandonPacewic (commented):
Ran into this, not 100% sure if this is intended or not. Happens on trackpad. 

https://github.com/user-attachments/assets/b8afee67-aad4-465c-9842-3ae7a6e6e75d

ariesninjadev:
> Ran into this, not 100% sure if this is intended or not. Happens on trackpad.
> 
>  Screen.Recording.2025-06-20.at.15.00.22.mov

Oooooh! This is going to take me forever to fix.

itspvty:
> Ran into this, not 100% sure if this is intended or not. Happens on trackpad.
> 
>  Screen.Recording.2025-06-20.at.15.00.22.mov

@BrandonPacewic I partially fixed the issue here, where now game pieces are actually draggable but they jitter a lot when dragged. I believe that this bug is beyond the scope of my PR but feel free to take a look and see if this jitter might be caused by an oversight in my code.

PepperLola (dismissed):
It still doesn't look like the robot is fixed to the cursor, since it's possible to move the cursor off of the robot while in drag mode.

https://github.com/user-attachments/assets/f66bcdac-37a8-4c7c-8b2c-04ca1d8fb536

I also don't think this was part of the task, but in previous versions you were able to use the scroll wheel to move whatever you're holding towards and away from the camera which is nice for accurately placing something.

BrandonPacewic (approved):
Code here is quite interesting, seems to work well. Nice job :godmode: 

I also believe that @PepperLola's problems have been properly addressed.

AlexD717 (approved):

PepperLola (approved):
snazzy

Dhruv-0-Arora (approved):

PR #1149, opened 2025-06-19
Fix Scoring Zone Scaling `[AARD-1913]`

rutmanz:
### Description
Scoring zones previously could not be scaled non-uniformly (without holding shift), the page would crash if the scalars were negative, and would leave ghosts behind while moving an existing scoring zone. They now no longer suffer from those problems

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1913)

### DIFF
--- a/fission/src/mirabuf/MirabufSceneObject.ts
+++ b/fission/src/mirabuf/MirabufSceneObject.ts
@@ -466,14 +466,13 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
 
     public RemoveScoringZoneObject(zone: ScoringZonePreferences) {
         const index = this._fieldPreferences?.scoringZones?.indexOf(zone) ?? -1
-        if (index != -1) {
-            const zoneObject = this._scoringZones[index]
+        if (index == -1) return
 
-            if (zoneObject != null) {
-                World.SceneRenderer.RemoveSceneObject(zoneObject.id)
-                zoneObject.id = -1
-            }
-        }
+        const zoneObject = this._scoringZones[index]
+        if (zoneObject == null) return
+
+        World.SceneRenderer.RemoveSceneObject(zoneObject.id)
+        zoneObject.id = -1
     }
 
     /**
--- a/fission/src/mirabuf/MirabufSceneObject.ts
+++ b/fission/src/mirabuf/MirabufSceneObject.ts
@@ -465,12 +465,13 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
     }
 
     public RemoveScoringZoneObject(zone: ScoringZonePreferences) {
-        const index = this._fieldPreferences?.scoringZones?.indexOf(zone)
-        if (index != null) {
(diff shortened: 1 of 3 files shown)


fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -447,6 +452,17 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
 
+    public RemoveScoringZoneObject(zone: ScoringZonePreferences) {
+        const index = this._fieldPreferences?.scoringZones?.indexOf(zone)
+        if (index != null) {
+            const zone = this._scoringZones[index]
+            if (zone != null) {
+                World.SceneRenderer.RemoveSceneObject(zone.id)
+                zone.id = -1
+            }
+        }
```
azaleacolburn:
```suggestion
        if (index === -1) return
        
        const zone = this._scoringZones[index]
        if (zone === undefined) return
        
        World.SceneRenderer.RemoveSceneObject(zone.id)
        zone.id = -1
```

azaleacolburn: I'm pretty sure that `indexOf` returns `-1` when the item isn't found.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

azaleacolburn (changes requested):

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -459,6 +464,18 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+    public RemoveScoringZoneObject(zone: ScoringZonePreferences) {
+        const index = this._fieldPreferences?.scoringZones?.indexOf(zone) ?? -1
+        if (index != -1) {
+            const zoneObject = this._scoringZones[index]
+
+            if (zoneObject != null) {
+                World.SceneRenderer.RemoveSceneObject(zoneObject.id)
+                zoneObject.id = -1
+            }
+        }
```
azaleacolburn: In my opinion, unnecessary nesting and type-agnostic checks are both anti-patterns.

https://youtu.be/CFRhGnuXG-4

rutmanz: I saw this in your suggestion earlier and looked at it both ways. I think that this five line function looks much clearer with nested `if`s than guard clauses. If you want me to change it though I can

azaleacolburn (approved):
lgtm :D

fission/src/mirabuf/MirabufSceneObject.ts
BrandonPacewic: I tend to agree with this, for things this small I do tend think its up to the discretion of the author to determine what looks the best. Not going to have a hard rule on this except for more extreme cases.

BrandonPacewic (approved):
🥅

PR #1150, opened 2025-06-20
Improve Preferences Interface `[AARD-1906]`

rutmanz:
### Description
Propagates preference changes instantly instead of waiting for the panel to close. Also replaces the type assertions in the preference system with more robust generics for type checking

AARD-1906


fission/src/systems/simulation/driver/HingeDriver.ts
```diff
@@ -84,20 +84,18 @@ class HingeDriver extends Driver {
...
-                }
+        this._gravityChange = (event: PreferenceEvent<"SubsystemGravity">) => {
+            const motorSettings = this._constraint.GetMotorSettings()
+            if (event.prefValue) {
+                motorSettings.set_mMaxTorqueLimit(this._maxTorqueWithGrav)
+                motorSettings.set_mMinTorqueLimit(-this._maxTorqueWithGrav)
+            } else {
+                motorSettings.set_mMaxTorqueLimit(MAX_TORQUE_WITHOUT_GRAV)
+                motorSettings.set_mMinTorqueLimit(-MAX_TORQUE_WITHOUT_GRAV)
             }
```
azaleacolburn, line 93: If you're calling the same functions in both branches, it's probably better to just create max and min torque limit variables and assign them with a ternary. I know you didn't write this code, but I think it's best to conform it to our standard while we're at it.

```suggestion
            const [maxTorque, minTorque] = event.prefValue
                ? [this._maxTorqueWithGrav, -this._maxTorqueWithGray]
                : [MAX_TORQUE_WITHOUT_GRAV, -MAX_TORQUE_WITHOUT_GRAV]
            motorSettings.set_mMaxTorqueLimit(maxTorque)
            motorSettings.set_mMinTorqueLimit(minTorque)
```

fission/src/test/PreferencesSystem.test.ts
```diff
@@ -12,9 +12,9 @@ describe("Preferences System", () => {
         expect(PreferencesSystem.getGlobalPreference("RenderScoreboard")).toBe(false)
     })
     test("Reset to default if undefined", () => {
-        PreferencesSystem.setGlobalPreference("ZoomSensitivity", undefined)
-        PreferencesSystem.setGlobalPreference("RenderSceneTags", undefined)
-        PreferencesSystem.setGlobalPreference("RenderScoreboard", undefined)
+        PreferencesSystem.setGlobalPreference("ZoomSensitivity", undefined as unknown as number)
+        PreferencesSystem.setGlobalPreference("RenderSceneTags", undefined as unknown as boolean)
+        PreferencesSystem.setGlobalPreference("RenderScoreboard", undefined as unknown as boolean)
```
azaleacolburn, line 17: Can we not just modify the signature of `setGlobalPreference` to make value potentially undefined?

Dhruv-0-Arora (commented):
This might be something to merge after #1070 since it will cause a lot of merge conflicts if we merge it before 😓

BrandonPacewic (commented):
#1135 first.

azaleacolburn (changes requested):
Some small nitpicks

fission/src/test/PreferencesSystem.test.ts
rutmanz: Could, but this is only used in a test file to check that values are guaranteed, not used in the actual program

fission/src/systems/simulation/driver/HingeDriver.ts
rutmanz: This code wasn't actually modified by this PR, just the function around it. Do you want me to change it anyway?

PepperLola: I think this is a situation where I prefer the if statement over the ternary, since it makes it a lot more clear at a glance what the code is doing. I generally agree with the if -> ternary conversion when assigning as long as the condition and values aren't too long, but when assigning multiple values I think the regular if makes more sense.

EDIT: I suppose two separate ternaries could work as well, but I think the ternary with the unpacking takes longer to interpret than either of the other ways.

azaleacolburn: I think the reason ternaries would work here is because the values intrinsically go together and are obviously bound by the same variable (min/max torque with or without gravity).

BrandonPacewic (commented):
Still some persistent issues here, will take over this PR since @rutmanz is out for the week.

fission/src/test/PreferencesSystem.test.ts
azaleacolburn: If a function can properly handle an undefined arg, that should be reflected in the type signature.

rutmanz: But providing an undefined type is not properly handled in the context of the rest of the program (an event is dispatched to notify other parts of the new value). This has no negative consequences in the testing environment because there are no listeners but passing undefined isn't just unused it would cause issues.

azaleacolburn (approved):
lgtm

BrandonPacewic (changes requested):
Double check and make sure that the changes made while the panel is open get reverted upon pressing cancel. For me this manifested in disabling/enabling the scoreboard, pressing cancel, and having the scoreboard not revert to whatever previous state it was in (and should still be in).

AlexD717 (approved):
Looks good. Personally I don't think that the console.log you added is necessary in the ScoreboardPanel.tsx but I will leave that up to you

BrandonPacewic (approved):
Awesome looks good.

> Also replaces the type assertions in the preference system with more robust generics for type checking

As I briefly mentioned before, in the future please include small refactors like this in separate PRs. I don't want to discourage these types of changes at all (in fact these are my favorite types to review, so please do more of them) just separate them out of feature pushes.

Nice work ⚙️

PR #1151, opened 2025-06-20
Match Mode Infrastructure `[AARD-1892]`

AlexD717:
### Objectives
Adds basic infrastructure for future match mode support. Originally planned as a Main Menu button, but integrating it could not be done cleanly requiring major changes to panel flow. Instead, it's added to the MainHUD, allowing full reuse of existing spawn/config panels. (Brandon said it was fine for me to try and do it this way; we can revisit if needed.)

### Description
To play Match Mode the user first spawns a field, robot(s) and configures them as normal. 
They then press the Start Match Mode button that is on the MainHUD
StartMatchMode
The scoreboard now shows the time left and a 15 second autonomous period, where player input isn't received starts
Screenshot 2025-06-20 094048
After those 15 seconds are over, a 135 second teleop period starts when the user can now control the robot. After the 135 seconds are over a end game screen is shown displaying the match results with a button to go back to sandbox (original) mode.
Screenshot 2025-06-20 094602

### Testing Done
- [ ] Robots/Fields can be spawned and configured like normal
- [ ] Match Mode can be started from the MainHUD
- [ ] Match Mode starts with a 15 second autonomous period where the robot does not receive input from the user
- [ ] Match Mode then has a 135 second teleop period where the robot receives input from the user
- [ ] Match Mode ends with a match results screen summarizing each teams score

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1892)
Screenshot 2025-06-20 094524


azaleacolburn:
Would it be possible for the remaining time (or for that matter, the entire top display) to only appear in match mode, rather than always being active?

fission/src/systems/MatchMode.ts
```diff
@@ -0,0 +1,120 @@
...
+    private initialTime: number = 0
+    private timeLeft: number = 0
+    private intervalId: number | null = null
+
+    private constructor() {}
+
+    static getInstance(): MatchMode {
+        if (!MatchMode.instance) {
+            MatchMode.instance = new MatchMode()
+        }
```
azaleacolburn: You should use the nullish coalescing assignment operator here (`??=`), which assigns a variable only if it was initially nullish (ie. undefined or null).

fission/src/systems/MatchMode.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        // Dispatch an event to update the time left in the UI
+        new UpdateTimeLeft(this.initialTime).Dispatch()
+
+        this.intervalId = window.setInterval(() => {
+            this.timeLeft--
+
+            if (this.timeLeft >= 0) {
+                new UpdateTimeLeft(this.timeLeft).Dispatch()
+            }
```
azaleacolburn: Correct me if I'm mistaken, but I think this should be `>` instead of `>=`

azaleacolburn (changes requested):

azaleacolburn (approved):
lgtm

BrandonPacewic (approved):
This looks good to me. This is the core starting task of reimplementing match mode for the summer so a lot of this is subject to change. Thanks for getting the ground work on this started! ⛹️

PR #1153, opened 2025-06-20
feat: add center of mass indicators and analysis tool panel

rutmanz:
### Description
Adds an analysis panel (per the Jira ticket) and a center of mass indicator for robot bodies that can be enabled. Currently the center of mass sphere floats in front of the rest of the scene so it can be seen even when inside a mesh (like Dozer).

image

### Testing Done
- Turning on and off with existing robots
- Adding robots while enabled
- Closing panel while enabled

AARD-1817


itspvty (dismissed):
It's preferable that one is able to continue viewing the CoM's of robots even after the analysis panel is closed. Further, it should be stored alongside other data in browser cache to preserve the setting. lmk if you disagree

Fix unit test **

rutmanz:
I'm inclined to agree that it would make sense as a setting, but the ticket specifies that it should be the start of a separate analysis mode/panel. cc @BrandonPacewic

BrandonPacewic (commented):
> >It's preferable that one is able to continue viewing the CoM's of robots even after the analysis panel is closed. Further, it should be stored alongside other data in browser cache to preserve the setting. lmk if you disagree
>>
>> Fix unit test **
>
>I'm inclined to agree that it would make sense as a setting, but the ticket specifies that it should be the start of a separate analysis mode/panel. cc @BrandonPacewic

I think I agree with these suggestions. I'll take over this PR since @rutmanz is out for the week.

AlexD717 (approved):
lgtm

BrandonPacewic (approved):
I do think that at some point this should be moved into a new section full of user facing tools at some point. Will make this a goal once we re-do the ui a bit.

Nice job ⚫

PR #1154, opened 2025-06-20
Fix toast crashes

rutmanz:
### Description
Dev branch was previously exhibiting intermittent page crashing with toasts (`type` was undefined). I don't see a definitive cause but I imagine it's related to the use of `type` as the parameter name. Renaming seems to have resolved the issue.


AlexD717 (approved):
This fixed the issue for me

Dhruv-0-Arora:
How do I reproduce this in `dev` since I don't see a problem when I run the test toasts from the debug panel

BrandonPacewic (approved):
> How do I reproduce this in `dev` since I don't see a problem when I run the test toasts from the debug panel

@Dhruv-0-Arora I believe this is just to fix the prop name collision with `type`.

Seems reasonable to me 🍞

PR #1155, opened 2025-06-20
Fix overlapping spawn and configure panels

rutmanz:
### Description
Closes the mirabuf import panel when the configure panel is opened. Prevents this behavior
image

AARD-1992

### DIFF
--- a/fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
+++ b/fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
@@ -327,6 +327,7 @@ const ConfigurePanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
 
         closePanel("choose-scheme")
         closePanel("import-mirabuf")
+        closePanel("initial-config")
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [])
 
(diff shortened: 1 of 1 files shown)


AlexD717 (changes requested):
There is still an issue where if you are in the Assembly Setup (where you choose the input scheme) the configure panel can be opened above it.

BrandonPacewic:
> There is still an issue where if you are in the Assembly Setup (where you choose the input scheme) the configure panel can be opened above it.

@AlexD717 Can you provide a screenshot of this issue?

AlexD717:
> > There is still an issue where if you are in the Assembly Setup (where you choose the input scheme) the configure panel can be opened above it.
> 
> @AlexD717 Can you provide a screenshot of this issue?

Screenshot 2025-06-23 120200
The issue occurs if you press spawn asset then select any robot and then press configure assets

BrandonPacewic (commented):
Upon further investigation this issue is best solved by adding some sort of exclusive panel option. The fixes presented here are really just bandaid fixes.

So this should be done after the UI refactor. I'll leave this PR open for now but hopefully the UI refactor will solve these problems.

AlexD717 (changes requested):
There is another issue now where if you are in the initial-config panel and its closed due to you opening the configuration panel, the robot never fully configures which results in you not being able to control it.

BrandonPacewic:
> There is another issue now where if you are in the initial-config panel and its closed due to you opening the configuration panel, the robot never fully configures which results in you not being able to control it.

Yes this is also something that I am aware of at this point. Thinking that UI refactor first and foremost is the best possible plan.

rutmanz:
Made obsolete by UI refactor

PR #1157, opened 2025-06-23
Intake Multiple Gamepieces

ryanzhangofficial:
Adds support for buffered gamepiece intake and configurable ejection order. Robots can now:

- Intake up to `maxPieces` simultaneously (from `intakePreferences`)
- Eject in either FIFO or LIFO order (based on `ejectorPreferences.ejectOrder`)

Behavior is integrated into existing intake/eject flow and scoring logic.
Screenshot 2025-06-23 135551
Screenshot 2025-06-23 135625

### DIFF
--- a/fission/src/systems/preferences/PreferenceTypes.ts
+++ b/fission/src/systems/preferences/PreferenceTypes.ts
@@ -46,6 +46,8 @@ export type IntakePreferences = {
     zoneDiameter: number
     parentNode: string | undefined
     showZoneAlways: boolean
+    maxPieces: number
+    ejectOrder: 'FIFO' | 'LIFO'
 }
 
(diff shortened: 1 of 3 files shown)


Dhruv-0-Arora (changes requested):
Build and formatting

BrandonPacewic (changes requested):
I'm having some issues where I can't actually intake gamepieces past 1 regardless of what the intake count is set to. I'm also getting this weird issue where not all of the gamepieces are intake-able and have physics disabled causing the robot to be able to drive through gamepieces sitting on the ground.

Dhruv-0-Arora (changes requested):
Unit tests

PR #1159, opened 2025-06-24
Configure Assets Save Cancel

ryanzhangofficial:
Currently, every panel only offers the user a single Close action in the bottom-right, which makes it impossible to clearly distinguish between “save my changes” vs “abort/close without saving.” This PR updates the Configure Assets panel to render two buttons—Save and Cancel.

Screenshot 2025-06-24 143018

> [!NOTE]  
> Save cancel does not currently work for Robots -> Brain, Configure Joints (Drivetrain)


fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
```diff
@@ -55,57 +59,43 @@ const AssemblySelection: React.FC<ConfigurationSelectionProps> = ({
...
     const robots = useMemo(() => {
-        const assemblies = [...World.SceneRenderer.sceneObjects.values()].filter(x => {
-            if (x instanceof MirabufSceneObject) {
-                return x.miraType === MiraType.ROBOT
-            }
-            return false
-        }) as MirabufSceneObject[]
-
-        return assemblies
-        // eslint-disable-next-line react-hooks/exhaustive-deps
```
azaleacolburn, line 72: We probably still want the eslint comment (above the new line 73)

fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
```diff
@@ -55,57 +59,43 @@ const AssemblySelection: React.FC<ConfigurationSelectionProps> = ({
...
     const fields = useMemo(() => {
-        const assemblies = [...World.SceneRenderer.sceneObjects.values()].filter(x => {
-            if (x instanceof MirabufSceneObject) {
-                return x.miraType === MiraType.FIELD
-            }
-            return false
-        }) as MirabufSceneObject[]
-
-        return assemblies
-        // eslint-disable-next-line react-hooks/exhaustive-deps
```
azaleacolburn, line 84: We probably still want the eslint comment (above the new line 79)

fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
```diff
@@ -336,15 +353,54 @@ const ConfigurePanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
...
+                    const name = selectedAssembly.assemblyName
+                    if (originalRobotPrefs.current) {
+                        PreferencesSystem.setRobotPreferences(name, originalRobotPrefs.current)
+                    }
+                    if (originalFieldPrefs.current) {
+                        PreferencesSystem.setFieldPreferences(name, originalFieldPrefs.current)
+                    }
+                    if (originalMotorPrefs.current) {
+                        PreferencesSystem.setMotorPreferences(name, originalMotorPrefs.current)
+                    }
```
azaleacolburn:
```suggestion
                    if (originalRobotPrefs.current)
                        PreferencesSystem.setRobotPreferences(name, originalRobotPrefs.current)
                    if (originalFieldPrefs.current)
                        PreferencesSystem.setFieldPreferences(name, originalFieldPrefs.current)
                    if (originalMotorPrefs.current) 
                        PreferencesSystem.setMotorPreferences(name, originalMotorPrefs.current)
```

fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -136,6 +151,34 @@ class PreferencesSystem {
...
+     * @returns {MotorPreferences} Motor preferences found for the given motor, or default motor preferences if none are found.
+     */
+    public static getMotorPreferences(miraName: string): MotorPreferences {
+        const allMotorPrefs = this.getAllMotorPreferences()
+
+        if (allMotorPrefs[miraName] == undefined) {
+            const defaultPrefs = DefaultMotorPreferences(miraName)
+            allMotorPrefs[miraName] = defaultPrefs
+            return defaultPrefs
+        }
```
azaleacolburn: A nullish coalescing assignment would be the better pattern here, since you're returning it anyway.
```suggestion
        allMotorPrefs[miraName] ??= DefaultMotorPreferences(miraName)
```

fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
```diff
@@ -315,16 +305,43 @@ const ConfigInterface: React.FC<ConfigInterfaceProps> = ({ configMode, assembly,
...
+
+                if (robotPrefs) {
+                    originalRobotPrefs.current = structuredClone(robotPrefs)
+                }
+                if (fieldPrefs) {
+                    originalFieldPrefs.current = structuredClone(fieldPrefs)
+                }
+                if (motorPrefs) {
+                    originalMotorPrefs.current = structuredClone(motorPrefs)
+                }
```
azaleacolburn:
```suggestion
                if (robotPrefs)
                    originalRobotPrefs.current = structuredClone(robotPrefs)
                if (fieldPrefs)
                    originalFieldPrefs.current = structuredClone(fieldPrefs)
                if (motorPrefs)
                    originalMotorPrefs.current = structuredClone(motorPrefs)
```

fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -136,6 +151,34 @@ class PreferencesSystem {
...
+    /** @returns Preferences for every motor that was found in local storage. */
+    public static getAllMotorPreferences(): { [key: string]: MotorPreferences } {
+        let allMotorPrefs = this.getPreference<{ [key: string]: MotorPreferences }>(MotorPreferencesKey)
+
+        if (allMotorPrefs == undefined) {
+            allMotorPrefs = {}
+            this._preferences[MotorPreferencesKey] = allMotorPrefs
+        }
+
+        return allMotorPrefs
```
azaleacolburn: It seems like `this._preferences[MotorPreferencesKey]` should equal `allMoterPrefs` no matter what, so we can remove this `if` statement and just have an unnecessary assignment some of the time. In my opinion, this is cleaner and doesn't actually incur any performance penalty because branching is slow.

```suggestion
        return (this._preferences[MotorPreferencesKey] =
            this.getPreference<{ [key: string]: MotorPreferences }>(MotorPreferencesKey) ?? {})
```

azaleacolburn (changes requested):
There are a few menus where this doesn't work (we already talked about that and I don't think it's your problem), so I think you're good when it comes to functionality. However, I have a ton of nitpicks about cleanliness that I'm happy to discuss.

fission/src/systems/preferences/PreferencesSystem.ts
ryanzhangofficial: The original code followed this format (e.g. getRobotPreferences), so I believe it's better to maintain consistency here.

azaleacolburn: I mean if we're changing it on line 174, we might as well change it here too, if we really want to maintain consistency. And on that note, we should change the the get preferences functions too (if they follow the same format).

azaleacolburn: This is still holding up my re-review. I think we should refactor these functions and we might as well do it now.

azaleacolburn (approved):
lgtm

AlexD717 (approved):
lgtm

BrandonPacewic (approved):
🔐

PR #1161, opened 2025-06-24
vite.config.ts Updated `[AARD-1953]`

AlexD717:
### Description
Update vite.config.ts to not use browser.name which is depreciated since Vite 3. All the unit tests also now run in both the Chromium and Firefox environment
Screenshot 2025-06-24 at 4 06 39 PM

### Testing Done
- Unit tests run and pass in Chromium
- Unit tests run and pass in Firefox

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1953)

### DIFF
--- a/fission/vite.config.ts
+++ b/fission/vite.config.ts
@@ -26,7 +26,7 @@ const plugins = [
         exclude: undefined, // Glob pattern, or array of glob patterns to ignore
         warnDuplicatedImports: true, // Warn if the same chunk was imported multiple times
         defaultExtension: "glsl", // Shader suffix when no extension is specified
-        compress: false, // Compress output shader code
+        minify: false, // Minify/optimize output shader code
         watch: true, // Recompile shader on change
         root: "/", // Directory for root imports
     }),
--- a/fission/vite.config.ts
+++ b/fission/vite.config.ts
@@ -1,8 +1,8 @@
-import { defineConfig } from "vitest/config"
-import * as path from "path"
-import react from "@vitejs/plugin-react-swc"
-import basicSsl from "@vitejs/plugin-basic-ssl"
-import glsl from "vite-plugin-glsl"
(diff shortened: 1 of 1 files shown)


fission/vite.config.ts
```diff
@@ -43,14 +43,24 @@ export default defineConfig({
         ],
     },
     test: {
-        testTimeout: 10000,
+        testTimeout: 60000,
```
Dhruv-0-Arora: This was changed to 30000 in another one of your PRs. Are we sure which timeout we want?

Dhruv-0-Arora (dismissed):

fission/vite.config.ts
AlexD717: I don't know what timeout we want. I think one minute is better, that way we won't have to change it again, but 30 seconds should also be fine.

azaleacolburn: How long do all the tests take to run on you machine @AlexD717.

AlexD717: The only test that takes long is the Mirabuf Parser test, which greatly varies. It usually takes ~26 seconds, but I think I had it run over 30 seconds a couple times. It caches the robot files for the test which takes different amounts of time.

Dhruv-0-Arora: Interesting. Usually it doesn't take that long for me. I would double check with @BrandonPacewic and @PepperLola on what we should have as the timeout

AlexD717: A lot of the new pr requests are timing out and failing the unit test, so we should probably increase the timeout

BrandonPacewic: This issue has now been "fixed" in #1174.

BrandonPacewic (approved):
✏️

azaleacolburn (approved):
lgtm

PR #1166, opened 2025-06-25
Sensitivity Sliders

itspvty:
This PR adds sensitivity sliders to general settings that allow the user to modify the DPI required to rotate the scene, and to rotate the scene using the Viewcube. The bounds on the sensitivity are higher on the Viewcube, because it's typically used for faster rotation (giving a "zoomed out" feel).

image


AlexD717 (approved):
Looks good, only thing I would change is placing the camera settings below the preferences as I think this would be used less than the preferences.

Dhruv-0-Arora (approved):
looks good. One extra feature for the view cube that could be its own ticket that I wanted to note was that on fusion, when you drag the view cube, your cursor is stuck in the same spot & disappears. To be consistent, if we did that it would be good.

Also is branch on your fork?

azaleacolburn:
> Also is branch on your fork?

It is on his fork :)

PR #1167, opened 2025-06-25
Clamp Function Unit Test `[AARD-1932]`

AlexD717:
### Description
Unit test for the clamp function in Utilities

### Testing Done
- Clamp function passes all test cases

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1932)

### DIFF
--- a/fission/vite.config.ts
+++ b/fission/vite.config.ts
@@ -43,7 +43,7 @@ export default defineConfig({
         ],
     },
     test: {
-        testTimeout: 30000,
+        testTimeout: 10000,
         globals: true,
         environment: "jsdom",
         browser: {
--- a/fission/vite.config.ts
+++ b/fission/vite.config.ts
@@ -1,8 +1,8 @@
-import { defineConfig } from 'vitest/config'
-import * as path from 'path'
-import react from '@vitejs/plugin-react-swc'
-import basicSsl from '@vitejs/plugin-basic-ssl'
-import glsl from 'vite-plugin-glsl'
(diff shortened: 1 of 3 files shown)


fission/vite.config.ts
Dhruv-0-Arora, line 1: Double check with @BrandonPacewic but im not sure we want to format this file since it looks like your IDE ran an autoformatter. If we do want to, we can make that another PR probably

Dhruv-0-Arora (changes requested):

fission/vite.config.ts
```diff
@@ -43,7 +43,7 @@ export default defineConfig({
         ],
     },
     test: {
-        testTimeout: 10000,
+        testTimeout: 30000,
```
Dhruv-0-Arora: See my review in #1161 about figuring out which test timeout we want

Dhruv-0-Arora: I believe it should be changed (if at all) in the other PR

Dhruv-0-Arora (changes requested):

Dhruv-0-Arora:
Otherwise this PR looks good

fission/vite.config.ts
AlexD717: I think the vite.config.ts pr (#1161) should be merge first and then I will merge dev into this discarding the vite.config.ts changes

Dhruv-0-Arora (approved):

BrandonPacewic (approved):
🧪

PR #1169, opened 2025-06-25
Configure Ejectable Toast

ryanzhangofficial:
AARD-1956

Small PR that adds on to the hidden console.log in `SetEjectable` with an on-screen info toast. Now, if a user tries to intake before configuring the ejector, they won’t be left blindly testing everything—not realizing the message was buried in the console (speaking from a little too much experience).


AlexD717 (approved):
Looks good. Only thing I would consider changing is allowing the toast to show up multiple times. That way if the user was distracted and missed the toast it will still show up the second time they try and intake.

Dhruv-0-Arora (approved):
Looks good!

I'm assuming the this._ejectableToastShown is to prevent the toasts from being spammed onto the screen since when I removed it, that's what it did.

Dhruv-0-Arora:
> s good. Only thing I would consider changing is allowing the toast to show up multiple times. That way if the user was distracted and missed the toast it will still show up the

It might be worth adding a cooldown period rather than using the this._ejectableToastShown

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -87,6 +87,8 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
     private _intakeActive = false
     private _ejectorActive = false
 
+    private _ejectableToastShown = false
```
AlexD717:
```suggestion
    private _timeLastShownToast = 0
```

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -438,7 +440,11 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
         }
 
         if (!this._ejectorPreferences || !this._ejectorPreferences.parentNode || !bodyId) {
-            console.log(`Configure an ejectable first.`)
+            if (!this._ejectableToastShown) {
```
AlexD717:
```suggestion
            if (Date.now() - this._timeLastShownToast > 1000) {
```

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -438,7 +440,11 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+                console.log(`Configure an ejectable first.`)
+                Global_AddToast?.("info", "Configure Ejectable", `Configure an ejectable first.`)
+                this._ejectableToastShown = true
```
AlexD717:
```suggestion
                this._timeLastShownToast = Date.now()
```

ryanzhangofficial: Yeah I've been working on the same changes, but problem is when you actually set the ejector and click close, for some reason the same toast pops up which is odd.

AlexD717 (changes requested):
The issue with the toast appearing when you closing a menu hasn't been fixed. I think that will confuse a lot of people and its better to just show it once instead, if this can't be fixed.

ryanzhangofficial:
Yeah I think it's a more global issue. When I re-tried the original code, it seems like the console.log message appears when you hit cancel as well. I also believe this causes more confusion. @Dhruv-0-Arora thoughts?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -438,7 +441,12 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
         }
 
         if (!this._ejectorPreferences || !this._ejectorPreferences.parentNode || !bodyId) {
-            console.log(`Configure an ejectable first.`)
+            const now = Date.now()
+            if (now - this._lastEjectableToastTime > MirabufSceneObject.EJECTABLE_TOAST_COOLDOWN_MS) {
```
Dhruv-0-Arora, line 525: Im not sure this is the best way to create a cooldown. I would look into doing something like this

```ts
export const useCooldown = (duration: number): UseCooldownReturn => {
  const [isOnCooldown, setIsOnCooldown] = useState(false);
  const timeoutRef = useRef<NodeJS.Timeout | null>(null);

  const startCooldown = useCallback(() => {
    if (!isOnCooldown) {
      setIsOnCooldown(true);
      timeoutRef.current = setTimeout(() => {
        setIsOnCooldown(false);
      }, duration);
    }
  }, [isOnCooldown, duration]);

  useEffect(() => {
    // Clear the timeout when the component unmounts
    return () => {
      if (timeoutRef.current) {
        clearTimeout(timeoutRef.current);
      }
    };
  }, []);
```

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -87,6 +87,9 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
     private _intakeActive = false
     private _ejectorActive = false
 
+    private _lastEjectableToastTime = 0
+    private static readonly EJECTABLE_TOAST_COOLDOWN_MS = 500
```
Dhruv-0-Arora, line 98: Does this have to be static?

Dhruv-0-Arora (changes requested):

Dhruv-0-Arora:
> Yeah I think it's a more global issue. When I re-tried the original code, it seems like the console.log message appears when you hit cancel as well. I also believe this causes more confusion. @Dhruv-0-Arora thoughts?

I can't seem to reproduce the toasts appearing when you close a menu. How do you see it?

AlexD717:
> > Yeah I think it's a more global issue. When I re-tried the original code, it seems like the console.log message appears when you hit cancel as well. I also believe this causes more confusion. @Dhruv-0-Arora thoughts?
> 
> I can't seem to reproduce the toasts appearing when you close a menu. How do you see it?


https://github.com/user-attachments/assets/95e248fb-6c0c-46eb-8b12-34593140ceae

BrandonPacewic:
@ryanzhangofficial What's the current status of this PR? I didn't see any notes about it from standup.

fission/src/mirabuf/MirabufSceneObject.ts
rutmanz: static makes sense for class-specific constants imo

ryanzhangofficial:
Not too sure. It's either one single toast or the ability for multiple toasts with the caveat discussed above. I tried multiple different cooldowns but they all result in the toast appearing when the ejector menu is configured (which causes even more confusing. I think rebasing back to `f4a711e` having a single toast is still helpful though.

AlexD717:
The purpose of this pr was to make the user experience better by adding a toast if they didn't configure a ejector and tried to intake. Currently it sends a toast that they need to configure a ejector right after they configure an ejector, which I think will just cause more confusion.

Dhruv-0-Arora (approved):

AlexD717 (approved):
Looks Good

BrandonPacewic (approved):

PR #1170, opened 2025-06-25
Exporter Materials Tagging `[AARD-1880]`

Dhruv-0-Arora:
## Task

The Exporter Materials Tagging creates a *new* Exporter Tab called **Tagging Options** which allows users to:
1. select a Fusion Body
2. select a tag from a dropdown list
which assigns that body the exporter tag selected.

Screenshot 2025-07-10 at 4 08 37 PM

AARD-1880

## Solution

Features of this PR include
- [x] the "add" button will spawn the body selection UI
- [x] a "cancel" button will appear if currently selecting a body
- [x] Only 1 Fusion body can be selected at a time (+ only bodies are selectable).

## Verification

- Check if the .mira file that downloads on export contains the keyword "tag" -- see Components.py to see how the tags are exported

---

- [ ] User-facing polish:


exporter/SynthesisFusionAddin/src/UI/TaggingConfigTab.py
```diff
@@ -0,0 +1,148 @@
...
+
+from src.Logging import getLogger, logFailure
+from src.UI.CreateCommandInputsHelper import createTableInput, createTextBoxInput
+
+logger = getLogger()
+
+
+class TaggingConfigTab:
+    # stores the types of tags available for selection
+    tagTypes = ["Softbody", "Rigid", "Chain", "Spring", "Rope"]
```
Dhruv-0-Arora, line 13: Need to finalize which tags we would like

rutmanz (changes requested):
I think it would make sense to show both the body and the parent component's name as well

Image

BrandonPacewic (changes requested):
It appears that removing tags is not working correctly at the moment.
image

Dhruv-0-Arora:
> It appears that removing tags is not working correctly at the moment. image

@BrandonPacewic To remove a tag you need to select a row first (which the Fusion UI is really bad at for doing since you need to click in between the columns.) I will update the UI message though so it is clearer 👍

BrandonPacewic (changes requested):
Are you able to select a tag and delete it? Because I'm not able to.

Dhruv-0-Arora:
> Are you able to select a tag and delete it? Because I'm not able to.

Yes; With Zachs UI refactor it will be a lot easier to select and delete a tag but right now you can select a row:
Screenshot 2025-07-10 at 10 52 10 AM

and click remove

rutmanz (approved):

BrandonPacewic (commented):
I can't select a tag to remove.


https://github.com/user-attachments/assets/9af78609-4190-40f3-9471-8ff173c1b260

exporter/SynthesisFusionAddin/src/UI/TaggingConfigTab.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self.taggingListTable.addToolbarCommandInput(addTagInputButton)
+        self.taggingListTable.addToolbarCommandInput(removeTagInputButton)
+        self.taggingListTable.addToolbarCommandInput(cancelInputButton)
+
+        self.toggleSelecting(False)
+
+    @property
+    def isVisible(self) -> bool:
+        return self.taggingConfigTab.isVisible or False
```
azaleacolburn, line 73: Is this `or False` really necessary if `isVisible` is already a boolean?

azaleacolburn (changes requested):

exporter/SynthesisFusionAddin/src/UI/TaggingConfigTab.py
BrandonPacewic: This mirrors what I did last year with `JointConfigTab.py`. I also did the same thing, can't quite remember why I did such a thing but I'm sure there was a reason for it:

https://github.com/Autodesk/synthesis/blob/d765c84d3fc9ddd240bb4312673b131b7ecb7d9e/exporter/SynthesisFusionAddin/src/UI/JointConfigTab.py#L122-L124

azaleacolburn (approved):
lgtm

Dhruv-0-Arora:
> [!Note]
If you are trying to test out whether the tags are properly exported / access the tags themselves, you can do so by adding this line in MirabufSceneObject.ts (line 187)
```ts
console.log(this._mirabufInstance.parser.assembly.data?.parts?.userData)
```

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Parser.py
```diff
@@ -23,6 +23,16 @@
...
+def reload() -> None:
+    """Reloads the Parser module"""
+    import importlib
+
+    importlib.reload(Components)
+    importlib.reload(Joints)
+    importlib.reload(Materials)
+    importlib.reload(PDMessage)
+    importlib.reload(assembly_pb2)
+    importlib.reload(types_pb2)
```
Dhruv-0-Arora: Note that I added this importlib.reload() in the Parser to properly allow testing with Component.py
I think we should merge it into dev in this PR vs opening another since it is mainly use-case based but there is no harm in keeping it.

BrandonPacewic (approved):
A Note for AARD-1883 and #1194, once a rigid body is selected it should be added to the table and material tag selection should be a dropdown within the table itself with `rigid` being the default.

PR #1172, opened 2025-06-26
Refactor Parser Error Handling

azaleacolburn:
### Description
The synthesis parser in the exporter now properly handles the errors it was catching, instead of always printing the stacktrace to the user, it now displays clean errors in most cases. This PR also removes the logging wrapper that was previously being used essentially to wrap everything in a `try`-`catch`. Errors now propagate through the parser as values with a new `Result` class which is inspired by the Rust enum of the same name. Each error contains a message and a severity (`Fatal` or `Warning`), the latter refers to the severity on the entire parsing process, so an error might cause a function to completely fail, but the error can be recovered by a parent function. On instantiation, each `Err` automatically logs itself based on it's severity. For errors that are not even fatal for the originating function, an `Err` should be instantiated but not returned, as to not break control flow.

While some were obvious, for several functions, I wasn't entirely sure how/why they would fail (despite them being marked as fallible with `logFailure`), so for those ones, we'll need to switch to the Result-based error handling system as we find errors.

### Objectives
- [x] `Result` implementation
- [x]  Document `Result`, `Err`, and `Ok` classes
- [x] Replace old error handling with nicer alternative (both for the user and developer)
  - [x] `Components.py`
  - [x] `Materials.py`
  - [x] `JointHierarchy.py`
  - [x] `Joints.py`
  - [x] `PDMessage.py`
  - [x] `PhysicalProperties.py`
  - [x] `RigidGroup.py`(unsure)
  - [x] `Utilities.py`
- [ ] Put checks after every opaque fusion function. This should probably be it's own PR.

### Testing Done
- [x] Successfully exported Dozer

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1858)


exporter/SynthesisFusionAddin/src/ErrorHandling.py
```diff
@@ -0,0 +1,115 @@
...
+        raise Exception(f"Called unwrap on Err: {self.message}")  # type: ignore
+
+    def unwrap_err(self) -> tuple[str, ErrorSeverity]:
+        """
+        Returns the error message and severity contained in the Err variant of the result, or raises an exception if unwrapping an Ok variant. Be sure to check first.
+        """
+
+        if self.is_err():
+            return (self.message, self.severity)  # type: ignore
+        raise Exception(f"Called unwrap_err on Ok: {self.value}")  # type: ignore
```
Dhruv-0-Arora: i personally dont think that the comments here are necessary

exporter/SynthesisFusionAddin/Synthesis.py
```diff
@@ -11,6 +11,7 @@
 from src.Logging import logFailure, setupLogger
 
 logger = setupLogger()
+from src.ErrorHandling import Err, ErrorSeverity
```
Dhruv-0-Arora: Is this a necessary import?

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
```diff
@@ -144,19 +160,24 @@ def populateJoints(
                         # else:
                         #     signals.signal_map.remove(guid)
 
-                _addJointInstance(joint, joint_instance, joint_definition, signals, options)
+                joint_result = _addJointInstance(joint, joint_instance, joint_definition, signals, options)
+                if joint_result.is_err() and joint_result.unwrap_err()[1] == ErrorSeverity.Fatal:
```
Dhruv-0-Arora: are you removing all prefacing "__"?

Dhruv-0-Arora (changes requested):

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
azaleacolburn: I'm removing the underscore for public functions that were marked as private

Dhruv-0-Arora (changes requested):
When I export a robot from the exporter, download it, and import it into synthesis, I receive the error:

Screenshot 2025-07-01 at 11 49 42 AM

Zach had this issue previously in dev however I was unable to reproduce it in dev; only this PR

azaleacolburn:
> When I export a robot from the exporter, download it, and import it into synthesis, I receive the error:
> 
> Screenshot 2025-07-01 at 11 49 42 AM
> 
> Zach had this issue previously in dev however I was unable to reproduce it in dev; only this PR

I'm quite certain this isn't an issue caused by this PR, but I'll investigate it to be sure.

exporter/SynthesisFusionAddin/src/Resources/PWM_icon/16x16-disabled.png
Dhruv-0-Arora, line 1: Note: What even are these files? I also have them in my git.

Dhruv-0-Arora (approved):

exporter/SynthesisFusionAddin/src/Resources/PWM_icon/16x16-disabled.png
azaleacolburn: They're the icons that are displayed in the exporter in the "Signal Type" column of the joint configuration panel when the Signal Type is PWM (which is the default).

exporter/SynthesisFusionAddin/Synthesis.py
```diff
@@ -70,6 +70,7 @@ def stop(_context: dict[str, Any]) -> None:
     Arguments:
         **context** *context* -- Fusion Data.
     """
+
```
BrandonPacewic: According to PEP 257 there should be no blank line between the docstring and code.

Not 100% clear if an auto formatter made this change or not.

.github/workflows/FusionTyping.yml
```diff
@@ -3,9 +3,9 @@ name: Fusion - mypy Typing Validation
 on:
   workflow_dispatch: {}
   push:
-    branches: [ prod, dev ]
+    branches: [prod, dev]
   pull_request:
-    branches: [ prod, dev ]
+    branches: [prod, dev]
```
BrandonPacewic: Rather small nitpick but we have included spaces in our other workflows.

If you feel like this should change thats fine but we should make that decision and apply it to all files rather than just randomly changing one in a PR.

exporter/SynthesisFusionAddin/src/ErrorHandling.py
```diff
@@ -0,0 +1,102 @@
...
+    Warning = 30  # Warning
+
+
+T = TypeVar("T")
+
+
+class Result(Generic[T]):
+    """
+    Result class for error handling, similar to the Result enum in Rust. The `Err` and `Ok` variants are child types, rather than enum variants though. Another difference is that the error variant is necessarily packaged with a message and a severity, rather than being arbitrary.
+    Since python3 has no match statements, use the `is_ok()` or `is_err()` function to check the variant, then `unwrap()` or `unwrap_err()` to get the value or error message and severity.
```
BrandonPacewic: This needs to be broken up into multiple lines. Try to not exceed 120 characters if possible.

This should really be enforced by one of our auto formatters.

exporter/SynthesisFusionAddin/src/ErrorHandling.py
```diff
@@ -0,0 +1,102 @@
...
+    def __init__(self, value: T):
+        self.value = value
+
+    def __repr__(self) -> str:
+        return f"Ok({self.value})"
+
+
+class Err(Result[T]):
+    """
+    The error variant of the Result class. Contains an error message and severity. Severity is the `ErrorSeverity` enum and is either Fatal, Error, or Warning, each corresponding to a logger severity level, Critical Error (50) and Warning (30) respectively. When an `Err` is instantiated, it is automatically logged in the current synthesis logfile.
```
BrandonPacewic: Also needs to be broken up into multiple lines.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
```diff
@@ -14,16 +19,16 @@
 from src.Proto import assembly_pb2, joint_pb2, material_pb2, types_pb2
 from src.Types import ExportMode
 
-# TODO: Impelement Material overrides
-
 
-def _MapAllComponents(
+# TODO: Impelement Material overrides
+def MapAllComponents(
```
BrandonPacewic: We might as well change all instances of this to camel case as well.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
```diff
@@ -65,28 +81,37 @@ def processBody(body: adsk.fusion.BRepBody | adsk.fusion.MeshBody) -> None:
...
-            processBody(body)
+            process_result = processBody(body)
+            if process_result.is_err() and process_result.unwrap_err()[1] == ErrorSeverity.Fatal:
+                return process_result
+
+    return Ok(None)
 
 
-def _ParseComponentRoot(
+def ParseComponentRoot(
```
BrandonPacewic: We might as well change all instances of this to camel case as well.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
```diff
@@ -32,31 +37,42 @@ def _MapAllComponents(
 
         comp_ref = guid_component(component)
 
-        fill_info(partsData, None)
+        fill_info_result = fill_info(partsData, None)
+        if fill_info_result.is_err() and fill_info_result.unwrap_err()[1] == ErrorSeverity.Fatal:
+            return fill_info_result
```
BrandonPacewic: These two lines are duplicated a lot. Should simplify it to something like:

```python
if result.is_fatal:
    return result
```

There might also be a more pythonic way of getting this into 1 or 2 lines. Which would be ideal with how often this pattern is used. But I'm not 100% sure what that would look like.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
```diff
@@ -134,11 +168,15 @@ def __parseChildOccurrence(
         try:
             part.appearance = "{}_{}".format(occurrence.appearance.name, occurrence.appearance.id)
         except:
+            _: Err[None] = Err("Failed to format part appearance", ErrorSeverity.Warning)
```
BrandonPacewic, line 176: Is the full syntax of `_: Err[None] = ...` really needed here? I'm guessing that this is to silence a flake warning. This gets quite verbose when we have longer error messages in other files within this PR.

https://github.com/Autodesk/synthesis/blob/456a853d430ffc2246acfad51ec972d7809bc7d0/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/JointHierarchy.py#L278-L280

Not too big of a deal but it would be nice if we can keep stuff like this to one line where possible.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
```diff
@@ -180,42 +225,52 @@ def GetMatrixWorld(occurrence: adsk.fusion.Occurrence) -> adsk.core.Matrix3D:
     return matrix
 
 
-@logFailure
-def _ParseBRep(
+def ParseBRep(
```
BrandonPacewic: We might as well change all instances of this to camel case as well.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
```diff
@@ -180,42 +225,52 @@ def GetMatrixWorld(occurrence: adsk.fusion.Occurrence) -> adsk.core.Matrix3D:
...
     plainmesh_out.normals.extend(mesh.normalVectorsAsFloat)
     plainmesh_out.indices.extend(mesh.nodeIndices)
     plainmesh_out.uv.extend(mesh.textureCoordinatesAsFloat)
 
+    return Ok(None)
+
 
-@logFailure
-def _ParseMesh(
+def ParseMesh(
```
BrandonPacewic: We might as well change all instances of this to camel case as well.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
```diff
@@ -225,8 +280,10 @@ def _ParseMesh(
     plainmesh_out.indices.extend(mesh.nodeIndices)
     plainmesh_out.uv.extend(mesh.textureCoordinatesAsFloat)
 
+    return Ok(None)
+
 
-def _MapRigidGroups(rootComponent: adsk.fusion.Component, joints: joint_pb2.Joints) -> None:
+def MapRigidGroups(rootComponent: adsk.fusion.Component, joints: joint_pb2.Joints) -> None:
```
BrandonPacewic: We might as well change all instances of this to camel case as well.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/JointHierarchy.py
```diff
@@ -188,26 +191,27 @@ class SimulationEdge(GraphEdge): ...
 class JointParser:
     grounded: adsk.fusion.Occurrence
 
+    # NOTE This function cannot under the value-based error handling system, since it's an  __init__ function
```
BrandonPacewic, line 194: Might be worth revisiting this at some point, to be honest I don't think this should be a class at all and we should instead try and make this more functional. But I don't think its really worth it at the moment.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
```diff
@@ -144,19 +160,24 @@ def populateJoints(
...
+                    return joint_result
 
                 # adds information for joint motion and limits
                 _motionFromJoint(joint.jointMotion, joint_definition)
 
             except:
-                logger.error("Failed:\n{}".format(traceback.format_exc()))
+                # TODO: Figure out how to construct and return this (ie, what actually breaks in this try block)
+                _: Err[None] = Err("Failed:\n{}".format(traceback.format_exc()), ErrorSeverity.Fatal)
```
BrandonPacewic, line 173: This appears to me to be something I missed when I redid the logging for the exporter. Previously, quite actually every function was wrapped in a try-except to catch any errors that the Fusion API could possibly throw and log and exit accordingly.

Without proper documentation I'm not 100% certain what the potential errors are here. It may be best to just wrap this function in a `@logFailure(messageBox=True)` decorator and leave it at that.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
```diff
@@ -542,11 +574,16 @@ def createJointGraph(
         elif nodeMap[suppliedJoint.parent.value] is not None and nodeMap[suppliedJoint.jointToken] is not None:
             nodeMap[str(suppliedJoint.parent)].children.append(nodeMap[suppliedJoint.jointToken])
         else:
-            logger.error(f"Cannot construct hierarhcy because of detached tree at : {suppliedJoint.jointToken}")
+            # TODO: This might not need to be fatal
+            return Err(
+                f"Cannot construct hierarchy because of detached tree at : {suppliedJoint.jointToken}",
+                ErrorSeverity.Fatal,
+            )
```
BrandonPacewic, line 583: I'm slightly concerned about what the requirements are to reach this point. Based on the description of the error there was a floating solid body not connected to the rest of the design. In which case part of your robot is simply not connected to the rest of it and would simply fall off when importing it.

I would be fine with the above behaviour being fatal but I'm not 100% sure what happens with fields and game pieces.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
```diff
@@ -26,26 +26,36 @@
 }
 
 
-def _MapAllPhysicalMaterials(
+def MapAllPhysicalMaterials(
```
BrandonPacewic: We might as well change all instances of this to camel case as well.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
```diff
@@ -127,49 +140,75 @@ def getPhysicalMaterialData(
...
+    if fill_info_result.is_err():
+        return fill_info_result
 
     for appearance in appearances:
         progressDialog.addAppearance(appearance.name)
 
+        # NOTE I'm not sure if this should be integrated with the error handling system or not, since it's fully intentional and immediantly aborts, which is the desired behavior
+        # TODO Talk to Brandon about this
         if progressDialog.wasCancelled():
             raise RuntimeError("User canceled export")
```
BrandonPacewic: Yes this should be part of the error handler. The reason that this behaviour was being properly handled before was that there was a top level try-except or `@logFaliure` decorator that would display this to a message box. We should mimic the same behaviour with the new system.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
```diff
@@ -206,12 +249,15 @@ def getMaterialAppearance(
...
     properties = fusionAppearance.appearanceProperties
+    if properties is None:
+        return Err("Apperarance Properties were None", ErrorSeverity.Fatal)
 
     roughnessProp = properties.itemById("surface_roughness")
     if roughnessProp:
         appearance.roughness = roughnessProp.value
 
     # Thank Liam for this.
+    # TODO Test if this is should be an error that we're just ignoring, or if it's actually just something we can skip over
```
BrandonPacewic: What would the error here be?

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Parser.py
```diff
@@ -249,3 +267,11 @@ def export(self) -> None:
...
 
         logger.debug(debug_output.strip())
+
+
+def handle_err_top[T](err: Result[T]) -> None:
+    if err.is_err():
+        message, severity = err.unwrap_err()
+        if severity == ErrorSeverity.Fatal:
+            app = adsk.core.Application.get()
+            app.userInterface.messageBox(f"Fatal Error Encountered: {message}")
```
BrandonPacewic: Thinking that we should turn this into some sort of decorator. Not the biggest fan of how this function looks in the following sections:

https://github.com/Autodesk/synthesis/blob/456a853d430ffc2246acfad51ec972d7809bc7d0/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Parser.py#L82-L121

BrandonPacewic: Additionally we should add a way to message box a given warning or error, even if its not necessarily fatal.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/RigidGroup.py
BrandonPacewic, line 1: For changes like this, I’d prefer that they be left out of larger PRs in the future. While this particular case is small, these types of unrelated refactors and typing improvements can add up across a full PR and make the review process more difficult.

Each change forces the reviewer to stop and ask, “Is this related to the task, or is it just a cleanup?” which adds unnecessary overhead. In this case, it's clear to me what's going on, but just something to keep in mind moving forward.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Utilities.py
BrandonPacewic, line 1: I'm curious about your decision to remove some, but not all, of the unused code in this file. Was there a specific reason certain parts were left untouched?

If we're going to clean up this file, it might be worth doing a full pass and moving these changes into a separate PR with a clear title, so it's easy to find later if we need it.

exporter/SynthesisFusionAddin/src/Resources/PWM_icon/16x16-normal.png
BrandonPacewic, line 1: While the other assets should I think are fine to include the visual diff of this file does not look quite correct.

fission/src/mirabuf/MirabufParser.ts
```diff
@@ -301,7 +301,7 @@ class MirabufParser {
 
             const mat = partInstance.transform
                 ? MirabufTransform_ThreeMatrix4(partInstance.transform)
-                : def.baseTransform
+                : def?.baseTransform
```
BrandonPacewic: Kind of an odd spot to include this change imo since it's unrelated to the rest of the PR.

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1554,7 +1554,7 @@ function filterNonPhysicsNodes(nodes: RigidNodeReadOnly[], mira: mirabuf.Assembl
         for (const part of x.parts) {
             const inst = mira.data!.parts!.partInstances![part]!
             const def = mira.data!.parts!.partDefinitions![inst.partDefinitionReference!]!
-            if (def.bodies && def.bodies.length > 0) {
+            if (def.bodies && def?.bodies?.length > 0) {
```
BrandonPacewic: Kind of an odd spot to include this change imo since it's unrelated to the rest of the PR.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
```diff
@@ -168,19 +189,25 @@ def _addJoint(joint: adsk.fusion.Joint, joint_definition: joint_pb2.Joint) -> No
         joint_definition.origin.y = 0.0
         joint_definition.origin.z = 0.0
 
-        logger.error(f"Cannot find joint origin on joint {joint.name}")
+        # TODO: We definitely could make this fatal, figure out if we should
+        _: Err[None] = Err(f"Cannot find joint origin on joint {joint.name}", ErrorSeverity.Warning)
```
BrandonPacewic, line 194: I'm assuming that this occurs when a joint exists and is selected but the solid body its connected to cannot be found? (We really need to document all this stuff).

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
```diff
@@ -106,7 +113,8 @@ def populateJoints(
 
         if joint.jointMotion.jointType in AcceptedJointTypes:
             try:
-                #  Fusion has no instances of joints but lets roll with it anyway
+                # Fusion has no instances of joints but lets roll with it anyway
+                # ^^^ This majorly confuses me ^^^
```
BrandonPacewic, line 118: Yea... Hunter left this note when he implemented ball joints.

Essentially, Fusion doesn't distinguish between joint definitions and joint instances, but our exporter code does. I'm assuming that this is just how the mira format works? So that it can reuse some definitions? I'm not 100% sure but this fits our data model for robots even though it's different than what Fusion does.

Also, fun fact-this is a good example of why formatting your code properly before merging matters. The git blame on this line got thrown off because Hunter originally used two spaces, and the auto-formatter removed one. I had to dig through a different commit to figure out who originally wrote the comment.

exporter/SynthesisFusionAddin/logs/.gitkeep
BrandonPacewic, line 1: This .gitkeep file exists to ensure the logs/ directory is present for writing log files. If we're removing it, we should double-check that we handle the case where its absent correctly.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
```diff
@@ -134,11 +168,15 @@ def __parseChildOccurrence(
...
         except:
+            _: Err[None] = Err("Failed to format part appearance", ErrorSeverity.Warning)
             part.appearance = "default"
         # TODO: Add phyical_material parser
 
+    # TODO: I'm fairly sure that this should be a fatal error
     if occurrence.component.material:
         part.physical_material = occurrence.component.material.id
+    else:
+        __: Err[None] = Err(f"Component Material is None", ErrorSeverity.Warning)
```
BrandonPacewic, line 184: Two questions about this: 1) Is it even possible to have a physical component is Fusion *without* a material and 2) do we need to abort the whole export process if this happens?

BrandonPacewic (changes requested):

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: That was almost certainly unintentional

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
azaleacolburn: Yes, joint definitions and instances are separate.

exporter/SynthesisFusionAddin/logs/.gitkeep
azaleacolburn: I'll add it back, it got removed when I deleted and recreated the logs directory at one point (because a logfile wasn't being created and I was confused).

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
azaleacolburn: It's something that was checked for before, so I assume it's possible, as for whether we should abort, I really don't know. I'll try manually setting it to none and seeing if it breaks.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
azaleacolburn: We definitely should document all the error-cases, should that be in this PR, or another one?

.github/workflows/FusionTyping.yml
azaleacolburn: I believe this was autoformatted, I didn't edit this file, I only opened it.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
azaleacolburn: It is fatal.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
BrandonPacewic: Should just be in another one, morso just making a note of this. I would really like to spend some significant time with the exporter parser and fully document everything in detail.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
BrandonPacewic, line 176: Is the full syntax of `_: Err[None] = ...` fully needed here? I'm guessing that this is to silence a flake warning. This gets quite verbose when we have longer error messages in other files within this PR.

https://github.com/Autodesk/synthesis/blob/456a853d430ffc2246acfad51ec972d7809bc7d0/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/JointHierarchy.py#L278-L280

Not too big of a deal but it would be nice if we can keep stuff like this to one line where possible.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
BrandonPacewic: These two lines are duplicated a lot. Should look for a way to simplify it to something like:

```python
if result.is_fatal:
    return result
```

There might also be a more pythonic way of getting this into 1 or 2 lines. Which would be ideal with how often this pattern is used. But I'm not 100% sure what that would look like.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
BrandonPacewic, line 583: I'm slightly concerned about what the requirements are to reach this point. Based on the description of the error there was a floating solid body not connected to the rest of the design. In which case part of your robot is simply not connected to the rest of it and would simply fall off when importing it.

I would be fine with the above behaviour being fatal but I'm not 100% sure what happens with fields and game pieces.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
```diff
@@ -127,49 +140,75 @@ def getPhysicalMaterialData(
...
+    if fill_info_result.is_err():
+        return fill_info_result
 
     for appearance in appearances:
         progressDialog.addAppearance(appearance.name)
 
+        # NOTE I'm not sure if this should be integrated with the error handling system or not, since it's fully intentional and immediantly aborts, which is the desired behavior
+        # TODO Talk to Brandon about this
         if progressDialog.wasCancelled():
             raise RuntimeError("User canceled export")
```
BrandonPacewic: Yes this should be part of the error handler. The reason that this behaviour was being properly handled before was that there was a top level try-except or `@logFaliure` decorator that would display this to a message box. We should mimic the same behaviour with the new system.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
BrandonPacewic: What would the error here be?

fission/src/mirabuf/MirabufParser.ts
BrandonPacewic: Kind of an odd spot to include this change imo since it's unrelated to the rest of the PR.

BrandonPacewic (commented):
There is still a handful of `TODOs` as well. We can either investigate them here or we can just ensure that the behaviour of this new system is the same as the old one, make a note of it, and move on.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
azaleacolburn: Ok cool 👍

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
azaleacolburn: I'll test this with exporting a field.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
azaleacolburn: I'm fairly certain it's impossible to make that one line, as that would require a function returning from its parent, unless there's some python feature to do it.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
azaleacolburn: It's just there to silence a warning, it's not required. Would it be preferable to just live with the warning?

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
azaleacolburn: Should all fatal errors generate alerts?

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
azaleacolburn: I think I meant `modelItem` being `None`, since we just essentially early return from the function, it might have been a situation that silently broke the export before.

fission/src/mirabuf/MirabufParser.ts
azaleacolburn: Ok, I'm not sure if this was introduced since the last review, but I cannot remember doing this, it was probably me being dumb with git or smth.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Components.py
BrandonPacewic: I guess its fine to leave. Just curious since nothing I use throws these warnings and they aren't enforced for the whole of the exporter to my knowledge.

Just to confirm, the full typing of underscore followed by colon and an explicit type is required?

azaleacolburn: Yes, otherwise I get warnings about an unused value (mypy) and an unknown type (basedpyright) respectively.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
BrandonPacewic: I want to say yes, if the export cannot be completed (which is what a fatal error should signify) then we need to tell the user something about why we were unable to export their design.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
BrandonPacewic: In that case what's the plan for this todo?

exporter/SynthesisFusionAddin/src/ErrorHandling.py
```diff
@@ -0,0 +1,113 @@
...
+        self.message = message
+        self.severity = severity
+
+        self.write_error()
+
+    def __repr__(self) -> str:
+        return f"Err({self.message})"
+
+    def write_error(self) -> None:
+        logger.log(self.severity.value, self.message)
```
BrandonPacewic, line 125: Thoughts on adding some sort of system to show the traceback of the error?

This way we can see what file the error occurred in and the exact line number it was created on. The general `@logFailure` decorator supports this.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Parser.py
BrandonPacewic: Thoughts on making this some sort of function decorator? Feels a little verbose currently but I'm open to suggestions.

https://github.com/Autodesk/synthesis/blob/456a853d430ffc2246acfad51ec972d7809bc7d0/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Parser.py#L82-L121

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
azaleacolburn: Ok, this is already how the `handle_err_top` decorator works, so that's good.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Materials.py
azaleacolburn: I think it can be removed.

exporter/SynthesisFusionAddin/src/ErrorHandling.py
azaleacolburn: I'm not sure that full trackbacks are helpful, but the originating line number definitely would be.

azaleacolburn: The issue with this, is with the `@logFailure` decorator, the traceback, line numbers, etc, are all handled by the python interpreter itself, while this isn't necessarily information we have access to without passing in the literal line number as an argument. There may be a workaround though.

exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/Joints.py
azaleacolburn: It isn't an issue when exporting fields.

BrandonPacewic (commented):
Looking good.

In `JointHierarchy.py` there are still some exceptions that are following the old pattern:

https://github.com/Autodesk/synthesis/blob/1121ac676952e947702a10c227277b2941d77cc5/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/JointHierarchy.py#L224-L226

This is causing the message box to display with the error AND the full traceback being showed to the user which was the previous behaviour of our logger. Not really sure how we should update this since it's within an `init` but we maybe could avoid showing the full traceback to the user and confirm that the export is actually being stopped as described by the messagebox.

azaleacolburn:
> Looking good.
> 
> In `JointHierarchy.py` there are still some exceptions that are following the old pattern:
> 
> https://github.com/Autodesk/synthesis/blob/1121ac676952e947702a10c227277b2941d77cc5/exporter/SynthesisFusionAddin/src/Parser/SynthesisParser/JointHierarchy.py#L224-L226
> 
> This is causing the message box to display with the error AND the full traceback being showed to the user which was the previous behaviour of our logger. Not really sure how we should update this since it's within an `init` but we maybe could avoid showing the full traceback to the user and confirm that the export is actually being stopped as described by the messagebox.

Ya this is a tricky one. I think the way to go is to log the message (with the `Err` class), display a message box, then just remove the logging decorator and close the program quietly without logging the full traceback.

BrandonPacewic (changes requested):
I pushed a change that reverted a PWM icon update as the file was being replaced with a question mark. I also merged dev and fixed some merge conflicts.

1. It still looks like we are showing the full traceback upon a rigid component not being detected.
2. I'm getting an issue where the file name of an exported design is starting with a `v`.

image

azaleacolburn:
> It still looks like we are showing the full traceback upon a rigid component not being detected.
This is caused by the `logFailure` decorator around `ConfigureCommandExecuteHandler.notify` method, which initializes and calls the `Parser`. To fix this, I just put a `try` block around the `Parser` usage with an `except` block that just passes.

BrandonPacewic (approved):
Think this is in a good place. I'm still learning lots of things about the exporter and I'm honestly a bit surprised that it even works. Python, as much as I love it, hides many errors very well. I have found many silent bugs in the exporter and I expect that there are many many more that I don't know about. This should help us quite a bit. 

Love the rust error system, python is definitely not built for it but its really what we should be doing here. Nice work.

PR #1175, opened 2025-06-26
Bootstrap Unit Test Suite

itspvty:
Pretty self explanatory, not much to screenshot. Tests pass on dev. Many parts are checked, I skimped on some parts but only because it's a suite we can improve more on in the future; I think this is a good starting point.

### DIFF
--- a/fission/.gitignore
+++ b/fission/.gitignore
@@ -16,6 +16,8 @@ dist
 dist-ssr
 *.local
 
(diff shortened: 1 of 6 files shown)


itspvty:
Someone will need to merge dev and format for me. Sorry ❤️

azaleacolburn:
I'm closing this for #1179

PR #1177, opened 2025-06-26
Intake Multiple Gamepieces

ryanzhangofficial:
AARD-1821

Updates the intake and ejection system to support multiple game pieces by replacing the single `_ejectable` field with an `_ejectables` array. It enforces the `maxPieces` limit, prevents duplicates, and adds support for FIFO or LIFO ejection order.

https://github.com/user-attachments/assets/c744b57b-b752-4212-b1fa-f705c6164edd


AlexD717:
Is it intended behavior that when you press eject it ejects all the game pieces, or is it supposed to eject them one at a time?

ryanzhangofficial:
Honestly, I wasn't really sure how to handle that. I think it really depends on the design of the robot. I choose to eject all because on the flip side you could intake multiple simultaneously.

AlexD717:
What is the point of the LIFO and FIFO order if it just ejects all the pieces?

ryanzhangofficial:
That's true, I'll make the change.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -78,8 +78,9 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
 
     private _fieldPreferences: FieldPreferences | undefined
 
+    private _ejectables: EjectableSceneObject[] = []
     private _intakeSensor?: IntakeSensorSceneObject
-    private _ejectable?: EjectableSceneObject
+    // private _ejectable?: EjectableSceneObject
```
AlexD717:
```suggestion
```

AlexD717: I might be wrong, but should this just be removed?

AlexD717 (approved):
lgtm

fission/src/mirabuf/MirabufSceneObject.ts
ryanzhangofficial: Oh yeah I forgot to remove that.

Dhruv-0-Arora (dismissed):
Could you add a tooltip that explains the FIFO and LIFO eject order?

Also, the configure assets panel looks a but cluttered with how every thing is placed. This will probably be fixed with the UI refactor.
image

This isn't necessarily for this PR but I think it would probably be beneficial to remove the "reset orientation" button since the reset button does pretty much the exact same thing except resetting velocity.

I also think it might be helpful to put the max pieces into the ejector configuration since the pieces themselves are held at the ejector. A potential ticket might also be to display a gamepiece when configuring the ejector to help visualize (might not be helpful though since the arrow mesh is just fine.)
Also making the max pieces an input instead of a slider might be something worth considering.

ryanzhangofficial:
Added the tooltips and did some refactoring on the ejector panel. I agree that the merging of reset and reset orientation could be another PR. I think having a self input option in addition to the sliders for different settings in configure assets could warrant its own ticket, as it is sometimes frustrating to get exact values (e.g. velocity).

BrandonPacewic (approved):
Looks good to me. Nice job tracking down those weird issues as they seemed less than trivial. 

> Also, the configure assets panel looks a but cluttered with how every thing is placed. This will probably be fixed with the UI refactor.
image
> 
> This isn't necessarily for this PR but I think it would probably be beneficial to remove the "reset orientation" button since the reset button does pretty much the exact same thing except resetting velocity.
> 
> I also think it might be helpful to put the max pieces into the ejector configuration since the pieces themselves are held at the ejector. A potential ticket might also be to display a gamepiece when configuring the ejector to help visualize (might not be helpful though since the arrow mesh is just fine.)
Also making the max pieces an input instead of a slider might be something worth considering.

Just waiting for @Dhruv-0-Arora to re-review, I think this current version is more than fine. We can always re-evaluate before we do any sort of release.

PR #1180, opened 2025-06-30
Match Mode Sounds `[AARD-1929]`

AlexD717:
### Description
Adds FIRST sound effects for auto start, teleop start, endgame, and end of game. 

### Testing Done
- [x] FRC match sound is played when autonomous starts
- [x] FRC match sound plays when autonomous ends
- [x] FRC match sound plays when teleop starts
- [x] FRC match sound plays at the start of endgame (20 seconds before end of teleop)
- [x] FRC match sound plays when teleop ends

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1929)

_Note: could someone delete Jira issue AARD-1954. I created that issue before I realized that AARD-1929 existed and addressed the same thing._


ryanzhangofficial (commented):
They sound great! Just to be on the safe side, are you sure the match sounds aren't copyrighted?

rutmanz:
Yeah, the sounds all seem to play at the right times and aren't too jarring, but I'm concerned about the attribution/legal requirements for this

Dhruv-0-Arora (commented):
Definitely worth double checking whether it is copyrighted. The best way would probably be to reach out to FRC themselves since through a quick google search I wasn't able to find the copyright guidelines.

Dhruv-0-Arora:
So unfortunately, reading through their copyright guidelines, they state

> Commercial entities are not permitted to use FIRST and LEGO Intellectual Property unless they have
a prior written agreement with FIRST ...

Meaning that we will need to reach out to them first. Might be better to wait with this PR anyways so we can get the UI refactor in and prevent any further merge conflicts.

AlexD717:
Do we have a Synthesis team email to contact FRC? If so, who will do that? Also, this PR doesn't have any relation to the UI.

BrandonPacewic (changes requested):
Blocked: Waiting on licensing information.

Dhruv-0-Arora:
Emailed FRC and it looks like the "Standard Arena Sounds" are royalty free. However, it we are to use a year specific match sounds, we would need to purchase a license. "(this happened in the 2024 season with the unique endgame sound for example)" @BrandonPacewic @AlexD717

Dhruv-0-Arora (changes requested):
Conflicts

ryanzhangofficial (approved):
It's great that the legal stuff has been sorted out. In terms of user experience, I don't think there is any problem and the timing seems to be on point.

BrandonPacewic:
> Emailed FRC and it looks like the "Standard Arena Sounds" are royalty free. However, it we are to use a year specific match sounds, we would need to purchase a license. "(this happened in the 2024 season with the unique endgame sound for example)" @BrandonPacewic @AlexD717

I don't think I have the full picture here but I think using whatever sounds were last available for free is the best option.

Dhruv-0-Arora (approved):
lgtm

BrandonPacewic (approved):

PR #1181, opened 2025-06-30
Preference System Unit Tests `[AARD-1936]`

AlexD717:
### Description
Implements and Expands Unit Tests for the Preference System.

### Testing Done
- [x] All unit tests pass

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1936)


Dhruv-0-Arora (dismissed):
Can you remove the console.log("Cleared all preferences") in the PreferenceSystem.ts file

```ts
    /** Removes all preferences from local storage. */
    public static clearPreferences() {
        window.localStorage.removeItem(this._localStorageKey)
        this._preferences = {}
        console.log("Cleared all preferences")
    }
```
Makes sense to remove in this PR since it console logs when you are running the test.

Dhruv-0-Arora:
> Can you remove the console.log("Cleared all preferences") in the PreferenceSystem.ts file
> 
> ```ts
>     /** Removes all preferences from local storage. */
>     public static clearPreferences() {
>         window.localStorage.removeItem(this._localStorageKey)
>         this._preferences = {}
>         console.log("Cleared all preferences")
>     }
> ```
> 
> Makes sense to remove in this PR since it console logs when you are running the test.

Might be worth creating like a UI popup later that notifies the user that they cleared their preferences.

AlexD717:
> Might be worth creating like a UI popup later that notifies the user that they cleared their preferences.

If you want, I could add a toast that appears, it would be fairly easy

BrandonPacewic (approved):
> If you want, I could add a toast that appears, it would be fairly easy.

Not really worth it for this PR I don't think. This covers what we want, I think removing the console log is just fine considering we would want to remove all these logs anyways at a later date.

@Dhruv-0-Arora Can you re-review?

ryanzhangofficial (approved):

BrandonPacewic (approved):

PR #1182, opened 2025-06-30
Rising and Falling Edge Sounds `[AARD-1930]`

rutmanz:
### Description
Improves audio caching (there was a noticeable delay in sound on first click before) and splits rising and falling edge sounds. Volumes are a little mismatched and I will play with those more if we decide that we prefer this system.

Currently set up to not play release sounds if the press sound is still active (it sounded rushed to play both on regular button clicks)

AARD-1930


rutmanz:
https://github.com/user-attachments/assets/77152eaf-9ab4-4cb0-8cdb-7ed49204923d

azaleacolburn (approved):
I like the addition, but I could image it being a bit annoying after a while. Would an option to toggle rising and falling edge sounds (vs just rising or just falling ig) be a good idea?

rutmanz:
You only hear the falling edge if you're lingering on the button presses, so it won't be happening enough to hear often if you're pressing buttons normally imo. Or am I misunderstanding what you mean?

ryanzhangofficial (approved):
Lgtm. I think if the user finds it annoying after a while, there's already functionality to mute all sound in settings.

AlexD717 (changes requested):
I can't hear the checkbox press up sound on my PC at all, and its also very quite on the MacBook compared to the other sounds. Could you make it louder?

azaleacolburn:
> You only hear the falling edge if you're lingering on the button presses, so it won't be happening enough to hear often if you're pressing buttons normally imo. Or am I misunderstanding what you mean?

No, you understood, I just lingered too much in my testing ig.

AlexD717 (approved):
Looks good. I don't know how hard this would be, but if it isn't to hard could you make the button press down sound effect like 0.2 seconds longer but have no sound there. Currently if I press a button normally sometimes it plays one sound but if I press the button for a bit to long, even though I would consider that a normal click, it plays the press up sound. This results in a inconsistent and annoying behavior.

Dhruv-0-Arora (approved):

Dhruv-0-Arora (dismissed):
Would you be able to reduce the sound of the second sound (button release)? It seems a bit sharp right now.

AlexD717 (changes requested):
Sounds a lot better, just a couple of small nitpicks.
- NOTICE.txt lists the DullClick.wav file path to be src/asset while it should be src/assets
- NOTICE.txt doesn't follow the best format
(Assuming that every sound file is under CC0 1.0, I think that this is more clear) 
```
This product bundles the following third-party sound assets:

* fission/src/assets/sound-files/checkdown.mp3  
* fission/src/assets/sound-files/checkup.wav  
* fission/src/assets/sound-files/clickdown.mp3  
* fission/src/assets/sound-files/clickup.mp3  
* fission/src/assets/sound-files/DullClick.wav  

All of the above files are licensed under the Creative Commons CC0 1.0 Universal Public Domain Dedication.  
License URL: https://creativecommons.org/publicdomain/zero/1.0/
```
- Currently the dropdownSoundEffect function is never called and SoundPlayer.play(dropdownMenuSound) is called in Dropdown.tsx (line 109)

AlexD717 (approved):
lgtm

BrandonPacewic (changes requested):
Checkbox sounds are very loud compared to button press sounds (as you might have just heard...)

rutmanz:
@BrandonPacewic they kind of sounded the same volume to me already (I might just not have sensitive enough ears), but I've adjusted them to all have the same peak volume, does this sound better?

BrandonPacewic (approved):
Sounds good to me 🔉

PR #1184, opened 2025-06-30
World.ts Unit Tests Created `[AARD-1933]`

AlexD717:
### Description
Creates Unit Tests for World.ts

### Testing Done
- All unit tests pass

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1933)

### DIFF
--- a/fission/src/test/World.test.ts
+++ b/fission/src/test/World.test.ts
@@ -1,18 +1,313 @@
+import { describe, test, expect, beforeEach, vi, afterEach } from "vitest"
+
+// Mock all the system dependencies before importing World
+vi.mock("@/systems/physics/PhysicsSystem", () => ({
+    default: vi.fn(() => ({
+        Update: vi.fn(),
+        Destroy: vi.fn(),
+    })),
+}))
+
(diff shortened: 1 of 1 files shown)


azaleacolburn (commented):
This is pretty cool, but I'm not sure how necessary it is, nearly everything tested here will self-evidently work based on very straightforward, high-level, infallible code. I don't see any technical issues, so if this' something we definitely want, I'll approve.

AlexD717:
There is a Jira issue for this task, which is why I made this. You are going to have to ask @BrandonPacewic

BrandonPacewic (approved):
> This is pretty cool, but I'm not sure how necessary it is, nearly everything tested here will self-evidently work based on very straightforward, high-level, infallible code. I don't see any technical issues, so if this' something we definitely want, I'll approve.

I agree this code is unlikely to break, but I think it’s still worth including these tests. They’re lightweight, low-maintenance, and help ensure the basics keep working as expected. Plus, if we were building something more complex, this stage of testing would be expected, so it makes sense to include it here too, even if the utility is a bit lower given the current scale of the project.

Dhruv-0-Arora (approved):

PR #1185, opened 2025-06-30
Exporter Reloading

Dhruv-0-Arora:
## Description
Allow for reloading the exporter code by disabling and re-enabling the exporter plugin -> instead of restarting fusion.

### Testing
Make a change to one of the tabs then disable and re-run the Synthesis addin to see if your changes were loaded.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1997)

### DIFF
--- a/exporter/SynthesisFusionAddin/Synthesis.py
+++ b/exporter/SynthesisFusionAddin/Synthesis.py
@@ -141,6 +141,8 @@ def register_ui() -> None:
     gm.elements.append(websiteButton)
 
 
+# Transition AARD-1997: Should revisit this once we start working on a re-write.
+# Jank solution for now - Brandon
 @logFailure
 def reload() -> None:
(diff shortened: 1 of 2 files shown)


azaleacolburn (approved):
nice fix!

BrandonPacewic (commented):
Ok this is indeed more jank than I was even expecting to be honest. I'll double check that this works and make some sort of note about this in the exporter rewrite. This dose *look* like a good solution but it just feels... not right...

BrandonPacewic (approved):
I'm not super happy about this but it certainly does appear to work. Need to double check how C++ Fusion extensions are treated. If they are compiled into some sort of shared library then referenced its possible that this issue is worse in C++ compared to Python. I have no way of knowing that till I run some tests but this will do for now.

🧇

PR #1192, opened 2025-07-02
Simulator Field Data Dev Tool `[AARD-1914]`

ryanzhangofficial:
### Task

AARD-1914

Implement an in-sim "Developer Tool" to allow live editing of Mira field files directly in the simulator UI.

### Symptoms
Currently there is no easy way to add or modify field metadata (scoring zones, robot spawns, camera rigs) without the exporter. Some of this mira editing should be moved to Synthesis.

### Solution
We store potential extra field data (e.g. scoring zones and spawn points) in the existing `mirabuf.UserData`([docs](https://www.mirabuf.dev/#mirabuf.UserData)) map using keys like `devtool:scoring_zones`. The user can add/edit in the `DeveloperToolPanel`. The purpose for this PR is to only make a way to edit and update the mira data, NOT make the implementations.

### Verification

- [x] Check if keys are added on console
- [x] `FieldMiraEditor.test.ts` (basic unit tests) passes

Screenshot 2025-07-02 131008
Screenshot 2025-07-02 131504


fission/src/mirabuf/FieldMiraEditor.ts
```diff
@@ -0,0 +1,53 @@
...
+        }
+        if (!this.parts.userData.data) {
+            this.parts.userData.data = {}
+        }
+    }
+
+    /**
+     * Get parsed data for a devtool key (e.g., 'devtool:scoring_zones').
+     */
+    getUserData<T = any>(key: string): T | undefined {
```
rutmanz: what's going on with this typing

rutmanz (changes requested):
It's not explicit in the ticket but I think an important part of having the mira editor is actually saving the updated mirabuf data back to the browser cache / making it exportable in some way. Otherwise there's no reason to make it part of the mirabuf structure

fission/src/mirabuf/FieldMiraEditor.ts
ryanzhangofficial: Do you mean the "any"?

azaleacolburn: I think he's confused about the point of this generic if it's just going to be `any`.

rutmanz: Ah okay I see the point in your test files. I think it makes sense for this to be an actual generic  It doesn't look like actually defining the structures is part of this PR but it would make sense to have the generic interfaces set up to make it more standard (and remove all the `any`s so the tests pass) when those are implemented
```ts
interface DevtoolMiraData {
    devtool:scoring_zones: unknown
	devtool:camera_locations: unknown
}
...
    getUserData<K extends keyof DevtoolMiraData>(key: K): DevtoolMiraData[K] | undefined {
```

ryanzhangofficial: Good point!

ryanzhangofficial:
I think export/cache could be addressed in a follow-up ticket after `AARD-1894`, which is the next step for this PR. The focus of this ticket is enabling users to add data to the Mirabuf, and that data is already accessible within the session.

ryanzhangofficial:
Something odd is that if I use prettier, I get the ES Lint error `Error:   61:13  error  Unnecessary semicolon  no-extra-semi`, but when I delete the semicolon, the ESLint tells me to use prettier again.

fission/src/mirabuf/FieldMiraEditor.ts
```diff
@@ -0,0 +1,67 @@
+import { mirabuf } from "../proto/mirabuf"
+
+interface DevtoolMiraData {
```
itspvty, line 3: I like this a lot, seems very useful. However, not sure why keys are hardcoded. The PR should provide the framework, maybe come with some preset keys (like scoring_zones, etc.), but then also allow us to add more without having to change code. LMK what you think.

itspvty (changes requested):

fission/src/mirabuf/FieldMiraEditor.ts
ryanzhangofficial: I think the challenge with allowing user-created keys is how actionable they would actually be. The goal of this dev tool is to let users input data that the simulator can do something with—not just store arbitrary metadata. So we need to enforce a specific format (like with devtool:scoring_zones) so we can reliably parse and act on the data. In the near future, the JSON formatting will definitely be standardized and enforced on the user side.

ryanzhangofficial: But if we were to actually pursue the AI chatbot idea, I think customizable keys would be quite useful...

rutmanz (approved):
If saving the modified mira structure is not in scope for this, lgtm

fission/src/mirabuf/FieldMiraEditor.ts
itspvty: That's less a challenge and more a "user friendly enhancement", despite it being called "Developer Tool"...

You can leave it if you want, but I think you should reconsider or ask someone smarter than me. Given that these keys essentially ARE metadata, I think users should have the freedom to edit them as they please.

ryanzhangofficial: That's true, but there should be at least a basic level of structure right?

ryanzhangofficial: I think the `Developer Tool` aspect lies in the editing of available keys.

itspvty: > That's true, but there should be at least a basic level of structure right?

I mean that's why I suggested there would still be the 3 necessary keys by default. Regardless, the PR is good so yeah

itspvty (approved):
LGTM

BrandonPacewic (approved):
> I think export/cache could be addressed in a follow-up ticket after AARD-1894, which is the next step for this PR. The focus of this ticket is enabling users to add data to the Mirabuf, and that data is already accessible within the session.

Will make a note of this in AARD-1894 as a part of the deliverables for said ticket. This will mean that AARD-1915 will have to come after AARD-1894.

Looking forward to getting the rest of this stuff in.

PR #1195, opened 2025-07-04
Fix Three.js Versioning Errors `[AARD-1982]`

rutmanz:
## Task
AARD-1982. Fix the seemingly inconsistent error when importing mirabuf assemblies with local files.

## Symptom
When using three.js versions greater than `0.165.0`, an exception is thrown by by line `137` in `MirabufInstance.ts` preventing any models from being imported or rendered in the scene.
https://github.com/Autodesk/synthesis/blob/155b97bfa6c591b35805d8d1db34c4fe6209893f/fission/src/mirabuf/MirabufInstance.ts#L230-L249

If you got past that point, the transform gizmo code was also broken and wouldn't render in `three@0.169.0`. 

Having outdated types (`@types/three@0.160.0`) also meant that some of these errors weren't being found by linters. 

## Solution

I went through the [migration guides](https://github.com/mrdoob/three.js/wiki/Migration-Guide) for three.js and the relevant PRs and updated our rendering code as appropriate.

## Verification

Run `bun install` or `npm install` to update your local `node_modules`, then observe that you
- can import mirabuf files
- can see the transform gizmo
- can see colors

---

- [x] User-facing polish:

### DIFF
--- a/fission/package.json
+++ b/fission/package.json
@@ -41,7 +41,7 @@
     "react-colorful": "^5.6.1",
     "react-dom": "^18.3.1",
     "react-icons": "^4.12.0",
-    "three": "0.165.0",
+    "three": "^0.178.0",
     "typescript-cookie": "^1.0.6"
   },
   "devDependencies": {
@@ -55,7 +55,7 @@
     "@types/pako": "^2.0.3",
     "@types/react": "^18.3.23",
     "@types/react-dom": "^18.3.7",
-    "@types/three": "^0.160.0",
+    "@types/three": "^0.178.0",
     "@typescript-eslint/eslint-plugin": "^7.18.0",
     "@typescript-eslint/parser": "^7.18.0",
     "@vitejs/plugin-basic-ssl": "^1.2.0",
--- a/fission/package.json
+++ b/fission/package.json
@@ -41,7 +41,7 @@
     "react-colorful": "^5.6.1",
     "react-dom": "^18.3.1",
     "react-icons": "^4.12.0",
-    "three": "^0.177.0",
+    "three": "0.165.0",
     "typescript-cookie": "^1.0.6"
   },
   "devDependencies": {
--- a/fission/src/mirabuf/MirabufInstance.ts
+++ b/fission/src/mirabuf/MirabufInstance.ts
@@ -235,8 +235,8 @@ class MirabufInstance {
                     const geometry = new THREE.BufferGeometry()
                     transformGeometry(geometry, body.triangleMesh!.mesh!)
                     const geoId = batchedMesh.addGeometry(geometry)
-
-                    batchedMesh.setMatrixAt(geoId, mat)
+                    const instanceId = batchedMesh.addInstance(geoId)
+                    batchedMesh.setMatrixAt(instanceId, mat)
 
                     let bodies = this._meshes.get(instance.info!.GUID!)
                     if (!bodies) {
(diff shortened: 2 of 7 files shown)


rutmanz:
~Actually hold off on this I found the migration guide for this change and am working on updating to match~

rutmanz:
Ok, now ready for review. Until this is merged, people experiencing the importing issue can downgrade to the working threejs version with  `bun install three@0.165.0`

fission/src/mirabuf/MirabufInstance.ts
```diff
@@ -235,8 +235,8 @@ class MirabufInstance {
                     const geometry = new THREE.BufferGeometry()
                     transformGeometry(geometry, body.triangleMesh!.mesh!)
                     const geoId = batchedMesh.addGeometry(geometry)
-
-                    batchedMesh.setMatrixAt(geoId, mat)
+                    const instanceId = batchedMesh.addInstance(geoId)
+                    batchedMesh.setMatrixAt(instanceId, mat)
```
Dhruv-0-Arora, line 239:
```suggestion
                    batchedMesh.setMatrixAt(batchedMesh.addInstance(geoId), mat)
```

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -83,7 +83,7 @@ class GizmoSceneObject extends SceneObject {
     public Setup(): void {
         // adding the mesh and gizmo to the scene
         World.SceneRenderer.AddObject(this._obj)
-        World.SceneRenderer.AddObject(this._gizmo)
+        World.SceneRenderer.AddObject(this._gizmo.getHelper())
```
Dhruv-0-Arora, line 86: Im not a fan of this. I personally believe threeJS should revert this ☹️

Dhruv-0-Arora (changes requested):
Im assuming that because in the diff you upgraded the version of threeJS, that the description is outdated and since you changed the implementation, it is infact fixed.

I will say, however, that upgrading the three version might cause other issues and we should definitely do a lot more testing and maybe even make the upgrade its own PR.

fission/src/mirabuf/MirabufInstance.ts
rutmanz: I'm personally opposed to having functions with side effects as inlined parameters. I think it makes more sense to have addInstance on its own line to make it clear that it's not just about the value

fission/src/systems/scene/GizmoSceneObject.ts
rutmanz: um yeah it's a little ugly but :shrug:

fission/package.json
```diff
@@ -41,7 +41,7 @@
     "react-colorful": "^5.6.1",
     "react-dom": "^18.3.1",
     "react-icons": "^4.12.0",
-    "three": "^0.177.0",
+    "three": "^0.178.0",
```
rutmanz, line 44: @Dhruv-0-Arora the bun.lock file was actually out of date, the version (since #1164) has been `0.177.0`, I checked the 177 -> 178 changelog and there's nothing relevant for us in it. This PR is fixing issues caused by the security-motivated version bump that already happened

AlexD717 (approved):
Fixed the error I was getting

BrandonPacewic (approved):
Managed to reproduce this error with a fresh fork of the sim. Will take a closer look at automated version bumps to avoid future issues.

🐰

Dhruv-0-Arora (approved):

PR #1199, opened 2025-07-08
Match Mode Rules `[AARD-1893]`

AlexD717:
## Task
Match Mode Rules and Configuration

AARD-1893

Create a way to allow the user to select a ruleset when starting match mode. This ruleset then affects the game parameters in specified ways to account for different rules in different years. 

## Symptom

Currently when you start match mode certain parameters, such as how long the match lasts, are hardcoded variables that can't be modified by the user. This means that if you want to test the game for a specific year where the autonomous time was longer than 15 seconds there is no way to do so.

## Solution

When you press match mode a configuration panel opens up allowing you to select which years ruleset to use, or upload your own custom json.
Screenshot 2025-07-08 at 11 40 08 AM
Currently the only data the json stores is the length of teleop, autonomous, and endgame, but this can be easily expanded to support more parameters. Once the user selects what ruleset to use the data is parsed and the corresponding variables that control match duration in MatchMode.ts are updated to the new values.

There are preset rulesets for the years 2025, 2024, and 2018 (power up). However, because we can't predict the game for next year, the user can upload their own json following our format to set the match length to whatever they want. Currently there is no tutorial explaining what format the data has to follow. Ideally their should be a help button that links to a page on our website explaining the format.

## Verification

To test these changes I selected in pre-set match ruleset and made sure that the time the match lasted, corresponded to the data in the json. I also tried uploading multiple valid and invalid custom json files, making sure that the data was accepted if it followed the format and rejected if it had invalid data. If only some data was invalid (such as missing an optional field like autonomous time) a warning is printed but the rest of the data is accepted. 

Unit tests were also written to render this panel making sure that nothing crashes, and to test uploading valid and invalid custom json rulesets. 

---

- [x] User-facing polish:


azaleacolburn:
We should probably just the format in the UI, or have a UI element to create custom rulesets, but that could be another PR.

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,286 @@
...
+            try {
+                const indexRes = await fetch("match-mode-config/index.json")
+                const fileNames: string[] = await indexRes.json()
+
+                const defaultConfigs = await Promise.all(
+                    fileNames.map(async fileName => {
+                        const res = await fetch(`match-mode-config/${fileName}`)
+                        return res.json()
+                    })
+                )
```
azaleacolburn: Why bother with an `index.json` file, why not just load all the cached files and all the files in `match-mode-config`?

AlexD717:
> We should probably just the format in the UI, or have a UI element to create custom rulesets, but that could be another PR.

I agree that a UI menu would be a better way to allow custom match rulesets, but I do think that it should be its own PR as this was mostly about the underlying framework.

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+    const fileUploadRef = useRef<HTMLInputElement>(null)
+
+    const uploadClicked = () => {
+        if (fileUploadRef.current) {
+            fileUploadRef.current.click()
+        }
+    }
+
+    const validateMatchModeConfig = (config: unknown): config is MatchModeConfig => {
```
azaleacolburn: Can we not just make `MatchModeConfig` the type of `config` directly? Like we can parse the object to a specific type and just catch that, bubbling the parse error back to the user. This feels too complicated for what we need. Additionally, if we're doing this whole validation function, we can assume `config` is of type `any` at least, since we can default an empty file to an empty object.

There're some things I really like about the way you handled this, and some things that could be done better. I would love to sit down with you some time and go over some of the nuances of types and json parsing in TS.

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,286 @@
...
+        if (typeof config !== "object" || config === null) {
+            return false
+        }
+
+        const configObj = config as Record<string, unknown>
+
+        // Check required fields
+        if (typeof configObj.id !== "string") {
+            console.error("Match mode config validation failed: 'id' field is required and must be a string")
+            Global_AddToast?.("error", "Invalid Match Mode Config", "The 'id' field is required and must be a string")
```
azaleacolburn: I do like these errors though

azaleacolburn (approved):
lgtm

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,272 @@
...
+    const fileUploadRef = useRef<HTMLInputElement>(null)
+
+    const uploadClicked = () => {
+        if (fileUploadRef.current) {
+            fileUploadRef.current.click()
+        }
+    }
+
+    // eslint-disable-next-line
+    const validateMatchModeConfig = (config: any): config is MatchModeConfig => {
```
rutmanz: Because `MatchModeConfig` has all of these properties as required and not optional this type guard is not accurately representing whether `config` is an instance of `MatchModeConfig`. I think it makes most sense to deal with assigning the defaults in here rather than outside of the function

fission/src/systems/MatchMode.ts
```diff
@@ -17,13 +18,24 @@ class MatchMode {
...
     static getInstance(): MatchMode {
         MatchMode.instance ??= new MatchMode()
         return MatchMode.instance
     }
 
+    setMatchModeConfig(config: MatchModeConfig) {
+        this.autonomousTime = config.autonomousTime
+        this.teleopTime = config.teleopTime
+        this.endgameTime = config.endgameTime
+    }
```
rutmanz: I feel like it makes more sense to have there be a `MatchModeConfig` field on `MatchMode` rather than transferring individual properties. That way it's simpler to extend when more properties are added

fission/src/systems/MatchMode.ts
```diff
@@ -53,13 +65,13 @@ class MatchMode {
...
         this.matchModeType = MatchModeType.Autonomous
-        this.startTimer(15, () => this.teleopModeStart(openModal))
+        this.startTimer(this.autonomousTime, () => this.teleopModeStart(openModal))
     }
 
     teleopModeStart(openModal: (modalName: string) => void) {
         // TODO play the teleop start sound
         this.matchModeType = MatchModeType.Teleop
-        this.startTimer(135, () => this.matchEnded(openModal)) // 2 minutes and 15 seconds
+        this.startTimer(this.teleopTime, () => this.matchEnded(openModal)) // 2 minutes and 15 seconds
```
BrandonPacewic: Should consider removing the comment here all together.

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,272 @@
...
+            <Box
+                component={"div"}
+                display={"flex"}
+                key={`button-box-${id}`}
+                flexDirection={"row-reverse"}
+                gap={"0.25rem"}
+                justifyContent={"center"}
+                alignItems={"center"}
+            >
+                <NegativeButton value={SynthesisIcons.DeleteLarge} onClick={secondaryOnClick} />
```
rutmanz: I don't think it makes sense for there to be a delete button for the built in modes (if they reappear when you reload the page)

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,272 @@
...
+                        window.localStorage.setItem(
+                            "match-mode-configs",
+                            JSON.stringify(matchModeConfigs.filter(c => c.id !== config.id))
+                        )
+                        matchModeConfigs.filter(c => c.id !== config.id)
+                    },
+                })
+            ),
+        [matchModeConfigs, openModal, closePanel]
+    )
```
BrandonPacewic: Correct me if I'm wrong but I believe that this should be rendered as JSX.

rutmanz (changes requested):

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,272 @@
...
+                    },
+                    secondaryOnClick: () => {
+                        // Delete the config from the local storage
+                        setMatchModeConfigs(prev => prev.filter(c => c.id !== config.id))
+                        window.localStorage.setItem(
+                            "match-mode-configs",
+                            JSON.stringify(matchModeConfigs.filter(c => c.id !== config.id))
+                        )
+                        matchModeConfigs.filter(c => c.id !== config.id)
+                    },
```
BrandonPacewic: I think there might be a race condition here since we are reading `matchModeConfigs` when `setMatchModeConfigs` has a chance of not being done updating yet.

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
BrandonPacewic: Should use `unknown` here over `any`.

BrandonPacewic (changes requested):

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -184,7 +200,21 @@ const MatchModeConfigPanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
...
+        }
+
+        // If validation passes, normalize the config with defaults for missing fields
+        const normalizedConfig: MatchModeConfig = {
+            id: configObj.id as string,
+            name: configObj.name as string,
+            isDefault: false, // User-uploaded configs are not default configs
+            autonomousTime: typeof configObj.autonomousTime === "number" ? configObj.autonomousTime : 15,
+            teleopTime: typeof configObj.teleopTime === "number" ? configObj.teleopTime : 135,
+            endgameTime: typeof configObj.endgameTime === "number" ? configObj.endgameTime : 20,
```
rutmanz: I think it makes sense to use the already-defined default config for these defaults, rather than defining them twice

https://github.com/Autodesk/synthesis/blob/11c1000a848a1a2f2167bdfb55b47e28b132c014/fission/src/systems/MatchMode.ts#L22-L29

rutmanz (changes requested):

rutmanz (commented):
Not sure why I didn't think about this before, but what's the intent behind making the default configs stored in json files rather than as an array in some TS file like the input schemes?

AlexD717:
> Not sure why I didn't think about this before, but what's the intent behind making the default configs stored in json files rather than as an array in some TS file like the input schemes?

I guess I never thought of that. The Jira issues said that I could choose the way to store the data and listed json as an example so that what I did. I guess I will go through and change it to being a typescript file

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+
+    const validateAndNormalizeMatchModeConfig = (config: unknown): MatchModeConfig | null => {
+        let valid = true
+
+        // Type guard to check if config is an object
+        if (typeof config !== "object" || config === null) {
+            console.error("Match mode config validation failed: config must be an object")
+            globalAddToast?.("error", "Invalid Match Mode Config", "Configuration must be an object")
```
rutmanz: since #1204 the optional chaining is not necessary (not just here but whenever you're doing toasts)

```suggestion
            globalAddToast("error", "Invalid Match Mode Config", "Configuration must be an object")
```

rutmanz (commented):
Other than that, lgtm

rutmanz (approved):

fission/src/systems/DefaultMatchModeConfigs.ts
Dhruv-0-Arora, line 1: Im not a huge fan of the name "DefaultMatchModeConfigs"

Also, what exactly is the endgameTime for?

Dhruv-0-Arora (changes requested):
Can you make the scoreboard reset after the Match is done?

Also, not sure if everyone would find this helpful, but maybe make a flag or notify the user in some way when the autonomous time or teleop time is enabled (this is moreso for the "Match Test" config. This change might not be useful so its really up to you.

fission/src/systems/DefaultMatchModeConfigs.ts
AlexD717: Currently endgameTime determines how long endgame lasts, and plays a sound when endgame starts, but it will later be expanded on to provide more features. For example, some penalties only count if it is/isn't endgame.

AlexD717: What would you suggest as the name instead of "DefaultMatchModeConfigs"

AlexD717:
> Can you make the scoreboard reset after the Match is done?

The scoreboard is reset when you start match mode again, but I can have it reset after match mode also

> Also, not sure if everyone would find this helpful, but maybe make a flag or notify the user in some way when the autonomous time or teleop time is enabled (this is moreso for the "Match Test" config. This change might not be useful so its really up to you.

I think that the game sounds do a good enough job of this, and that this should be implemented in a different PR (if we decide that we need it).

fission/src/systems/DefaultMatchModeConfigs.ts
Dhruv-0-Arora: No idea, if you can find a better name then I would encourage you to use that instead 👍

Dhruv-0-Arora (approved):
I think for another task it might be nice to hide the scoreboard unless you are currently in a match (since there isn't much utility outside of that). However, this lgtm

AlexD717:
> I think for another task it might be nice to hide the scoreboard unless you are currently in a match (since there isn't much utility outside of that). However, this lgtm

It would be better if this were a toggle option, so that they could choose to keep the scoreboard. If you are practicing a match in single-player mode, you might want to have a score tracker but not be limited by the match mode time.

BrandonPacewic:
> I think for another task it might be nice to hide the scoreboard unless you are currently in a match (since there isn't much utility outside of that). However, this lgtm

@Dhruv-0-Arora There’s actually already a way to hide the scoreboard — it’s available as a toggle in the General Settings panel.

In the future, if you have suggestions like this, I’d support you taking a quick look through the codebase or UI to see if the feature already exists. And if it doesn’t, to either create or suggest a ticket for it — that’d be really helpful.

BrandonPacewic (approved):

PR #1205, opened 2025-07-10
fix(ci): avoid storing empty cache

BrandonPacewic:
## Task

Fix the playwright tests.

## Symptom

~~The cache failed causing the workflow to cache empty contents over and over again without downloading them.~~

Cache was broken.

## Solution

~~Download browsers first on cache-miss, then snapshot and save them.~~

No longer using the cache.

## Verification

Workflow now runs.

---

- [x] User-facing polish:

### DIFF
--- a/.github/workflows/FissionUnitTest.yml
+++ b/.github/workflows/FissionUnitTest.yml
@@ -38,14 +38,6 @@ jobs:
         id: playwright-version
         run: echo "PLAYWRIGHT_VERSION=$(node -e "console.log(require('./package-lock.json').dependencies['@playwright/test'].version)")" >> $GITHUB_ENV
 
-      - name: Cache downloaded browsers
-        id: cache-browsers
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cache/ms-playwright/
-          key: ${{ runner.os }}-assets-playwright-${{ env.PLAYWRIGHT_VERSION }}-v3
-
       - name: Install Dependencies
         run: |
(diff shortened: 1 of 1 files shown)


AlexD717 (approved):

.github/workflows/FissionUnitTest.yml
```diff
@@ -58,6 +50,14 @@ jobs:
           npx playwright install --with-deps
           npx playwright install-deps
 
+      - name: Cache downloaded browsers
+        id: cache-browsers
```
rutmanz: Doesn't this cause the previous step to always run and defeat the purpose of the cache?

BrandonPacewic: Correct, trying to find out how to make it so that we verify that the cache actually has contents before saying that they are there and skipping the install.

rutmanz: It looks like it exits early if the required files are already installed, so if you keep them in the old order and just remove the conditional it seems like it should work

BrandonPacewic: 🤔

BrandonPacewic: Yea um, think this does not work unless I applied your diff wrong.

AlexD717 (approved):

rutmanz (approved):

BrandonPacewic:
Made a ticket to re-enable the cache as a part of AARD-1989.

PR #1206, opened 2025-07-10
APS Testing `[AARD-1941]`

AlexD717:
## Task
Create Tests for APS.ts

AARD-1941

## Symptom

No tests for the APS system

## Solution
Add tests for the APS system

## Verification

All Unit tests pass

---

- [x] User-facing polish:


fission/src/test/APSTesting.test.ts
```diff
@@ -0,0 +1,694 @@
...
+    getItem: vi.fn(),
+    setItem: vi.fn(),
+    removeItem: vi.fn(),
+    clear: vi.fn(),
+    length: 0,
+    key: vi.fn(),
+}
+
+// Override both window.localStorage and global localStorage
+Object.defineProperty(window, "localStorage", {
```
rutmanz: Why are you mocking localStorage? I feel like it would make more sense (and be less work, and be more helpful of a test) to check that the values in localStorage are what you expect rather than mocking the function calls and checking what they're called with

rutmanz: Mocking IMO is for when the mocked function won't behave correctly in the testing environment or isn't useful in the way it normally behaves, but it seems to me like you would just want to have a working localStorage implementation here

rutmanz (changes requested):

fission/src/test/APSTesting.test.ts
```diff
@@ -0,0 +1,635 @@
...
+        test("isSignedIn returns true when valid auth data exists", async () => {
+            localStorage.setItem("aps_auth", JSON.stringify(mockAuth))
+            expect(await APS.isSignedIn()).toBe(true)
+        })
+
+        test("handles corrupted auth data gracefully", async () => {
+            localStorage.setItem("aps_auth", "invalid json")
+            expect(await APS.isSignedIn()).toBe(false)
+        })
+    })
```
ryanzhangofficial, line 587: These tests are already covered in the "End-to-End User Journey" tests.

fission/src/test/APSTesting.test.ts
```diff
@@ -0,0 +1,635 @@
...
+            expect(mockFetch).toHaveBeenCalled()
+        })
+
+        test("handles user info load failure", async () => {
+            mockFetch.mockResolvedValueOnce(
+                createMockResponse(
+                    {
+                        error: "unauthorized",
+                        userMessage: "Invalid access token",
+                    },
```
ryanzhangofficial, line 621: Same comment for the error handling tests.

ryanzhangofficial (changes requested):

fission/src/test/APSTesting.test.ts
AlexD717: They mostly are, but I thought it was better to have "unit" tests and "end to end" tests. The end to end verifies that the general process works, while the unit tests verify the function works as expected. If a unit test fails we know exactly in which function it failed, vs that not being the case in the end to end tests.

rutmanz (approved):
lgtm

fission/src/test/APSTesting.test.ts
ryanzhangofficial: I guess it's fine to have an end to end instead of functionals here because it's testing the APS. Looks good then.

ryanzhangofficial (approved):

fission/src/ui/panels/DebugPanel.tsx
```diff
@@ -85,14 +85,14 @@ const DebugPanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
                     <Button
                         value={"Refresh APS Token"}
                         onClick={async () =>
-                            APS.isSignedIn() && APS.refreshAuthToken((await APS.getAuth())!.refresh_token, true)
+                            (await APS.isSignedIn()) && APS.refreshAuthToken((await APS.getAuth())!.refresh_token, true)
```
BrandonPacewic: Nice catch for changing `isSingedIn()` to be a async function.

I do believe there is a potential race condition in this line since `APS.getAuth()` is called twice in this line and state can technically change within those two calls. Should save this state to a variable rather than calling these functions inline.

BrandonPacewic (changes requested):
My apologies for merging dev before suggesting these changes. Should have not merged dev till I was ready to approve.

fission/src/ui/panels/DebugPanel.tsx
```diff
@@ -84,15 +84,16 @@ const DebugPanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
                     <LabelStyled>Autodesk Platform Services</LabelStyled>
                     <Button
                         value={"Refresh APS Token"}
-                        onClick={async () =>
-                            APS.isSignedIn() && APS.refreshAuthToken((await APS.getAuth())!.refresh_token, true)
-                        }
+                        onClick={async () => {
+                            const auth = await APS.getAuth()
+                            auth && APS.refreshAuthToken(auth.refresh_token, true)
+                        }}
```
BrandonPacewic, line 90: I think the syntax is a little messed up here. I believe since `auth` is a promise it always is evaluated as truthly. For this reason I think an if statement is required but correct me if I'm wrong.

rutmanz: `auth` isn't a promise here, it's `APSAuth | undefined` because `APS.getAuth()` is `await`ed. It's a bit unconventional to use short-circuiting in JS like this (more common in react components) but this does work

Dhruv-0-Arora (approved):

BrandonPacewic (approved):

PR #1187, opened 2025-07-01
Definition of Done: Update GitHub Pull Request Template

BrandonPacewic:
After talking with Nick and reviewing our Jira workflow, I decided to update our pull request template.

The new version is designed to force you to split up your description of what you are working on and what you did into a few different sections: 
1. Task – What you were assigned to do.
2. Symptom – How the issue is currently reflected in the product (what prompted the change).
3. Solution – What you did to fix the issue or implement the task.
4. Verification – How you verified that your solution works correctly.

At the bottom, I've also added a section that defines our **"Definition of Done"** - what needs to be in place before we can merge a PR and close the related issue. This includes some new additions:
- Tests:
  - From now on, if you can write a test for what you worked on, where applicable and it makes sense, you should do it.
- Safe disable path:
  - While we have been attempting to split up our larger tasks into smaller components this leaves many unfinished features in `dev` that should not be visible when pushing to `prod` and making a release.
  - For such features ensure there is a safe way to disable it from the developer side.
    - This can be as simple as a constant that can be toggled.
    - Just make sure that it's not 15 different sections that would need to be commented out if we wanted to temporarily disable a feature.
- Polish:
  - Now that we have the web based version of Synthesis is maturing, we're shifting more attention toward polish.
  - After your done with your task take some time to look it over, check your code and check how the product looks.
  - Take a bit of time to make any small tweaks you can to improve the overall polish of the user experience and our codebase.

Going forward, Jira issues will include more explicit **acceptance criteria** (I'll do my best), to make the review process smoother and more objective.

PR #1207, opened 2025-07-10
Added station selection

itspvty:
## Task
AARD-1988

The task is to add a UI selection for a robots' alliance **station**.

## Symptom
As we approach having a functional match mode, part of having fixed camera locations for players involves identifying unique positions for each player in the match. To give users more fine-grained control over their simulation, this feature allows users to pick each location instead of assigning them randomly.

## Solution
Alongside alliance color selection, the user can now select the station number (from 1 to 3) of the focused robot, both in the initial config panel and the regular config panel.

## Verification
I validated that the value updates and persists, through logging and visually in the UI. It follows the same hook structure as the Alliance selection.

Initial selection panel
image

Config sections (rephrased from "Alliance" to "Alliance / Station"
image

Inside the Alliance / Station menu
image


azaleacolburn:
Could we put this into this project's standard format for pull requests (#1187).

fission/src/ui/panels/configuring/assembly-config/interfaces/AllianceSelectionInterface.tsx
```diff
@@ -11,18 +12,57 @@ const saveSetAlliance = (alliance: Alliance, assembly: MirabufSceneObject) => {
     assembly.alliance = alliance
 }
 
+const saveSetStation = (station: Station, assembly: MirabufSceneObject) => {
+    assembly.station = station
+}
```
azaleacolburn, line 17: Is there a reason this isn't a setter on `MirabufSceneObject`?

itspvty: I just mirrored how the alliance setting is implemented; if you need me to change it, can you elaborate a little?

AlexD717 (approved):
Looks super nice. I especially like the fact that the color of the selected button changes with the alliance color. I'm assuming that more functionality will later be added (ex, spawning at driver station positions)

itspvty:
> Looks super nice. I especially like the fact that the color of the selected button changes with the alliance color. I'm assuming that more functionality will later be added (ex, spawning at driver station positions)

Thanks for the feedback! Yes, my current WIP is creating camera presets.

Dhruv-0-Arora (approved):

BrandonPacewic (commented):
I can't quite decide if this should go in at this point since we are still waiting for the UI refactor. Wondering what @PepperLola's preference is to updating the refactor with additional UI changes.

fission/src/ui/panels/configuring/assembly-config/interfaces/AllianceSelectionInterface.tsx
azaleacolburn: Like you have an external function that modifies an internal property of an object, usually that's the job of a setter method on the object itself. Setters make code cleaner and make it clear which properties the class expects to be modified in a very explicit way.

PR #1208, opened 2025-07-10
Add NPM Dependency Caching `[AARD-1991]`

rutmanz:
## Task

AARD-1991

Adds a cache step on fission/node_modules for the four workflows that include it

## Symptom

Previously the dependency installation step of the workflows could be up to around 50s

## Solution

I added a npm cache step that is shared across the four fission workflows, based on the hash of package.json.

This does mean that if a new patch version (or version otherwise compatible with the restrictions in `package.json`) is released, the cache won't be updated because the cache key won't change. This won't break anything, but it means that all workflows will keep downloading the new patch version independently instead of getting it from the cache until package.json changes

It would be ideal to hash a lockfile instead, but we don't have package-lock.json committed, we could switch the gh actions workflows to bun (probably resulting in additional speedups), however, if we don't see a problem with that.

cache download + dependency installation now take around 10s when there's a cache hit

## Verification

Observe that the github action workflow runtimes are shorter by ~30-50s

---

- [ ] User-facing polish:


azaleacolburn (approved):
very nice :D

AlexD717 (approved):
lgtm

BrandonPacewic (approved):
Great optimization, we should have a vote at some point to determine what package manager we want to use. I feel like I know which one is going to win... 🍞 (AARD-1992)

Dhruv-0-Arora (approved):

PR #1211, opened 2025-07-11
Support Fields Having Pre Select Scoring Zones `[AARD-1894]`

ryanzhangofficial:
## Task

Allow users to choose and cache/export where scoring zones are positioned.
AARD-1894

## Symptom

Benefits of having direct user editing: bulk edits and reuse, version control and diffs, schema validation

## Solution

- Include pre select scoring zones in-simulator as part of the new `Developer Tool`. 
- Create a framework for caching devtool keys
- Allow users to export mira files with edited `UserData` and import them for use cases such as cleared cache

Screenshot 2025-07-11 150237
Screenshot 2025-07-11 104009

## Verification

- [x] Added new tests in `FieldMiraEditor.test.ts` and all passed
- [x] Cached `ScoringZonePreferences` are loaded after refresh


fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -369,6 +403,55 @@ class MirabufCachingService {
...
+                assemblyName: assembly.info?.name ?? "unknown",
+                fileSize: updatedBuffer.byteLength,
+            })
+
+            return true
+        } catch (e) {
+            console.error("Failed to persist devtool changes", e)
+            World.analyticsSystem?.exception("Failed to persist devtool changes to cache")
+            return false
+        }
```
azaleacolburn, line 470: We should try to put the `try` block around as small a section of fallible code as possible. It seems like the fallible function here is the `event` method call (obviously any file I/O is subject to failure as well, but since we're creating the directory, it shouldn't regularly fail), unless I'm missing something. So maybe we could more the try block to be just around the `event` method class.

fission/src/ui/panels/DeveloperToolPanel.tsx
```diff
@@ -172,7 +250,17 @@ const DeveloperToolPanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
                             {selectedKey ? (
                                 <>
                                     {/* strip off the prefix here */}
-                                    <div className="font-bold text-sm mb-2">{selectedKey.replace(/^devtool:/, "")}</div>
+                                    {selectedKey === "devtool:scoring_zones" ? (
+                                        LabelWithTooltip(
+                                            "scoring_zones",
+                                            `Add and cache scoring zones. \n Example:\n[\n  {\n    \"name\": \"Red Zone\",\n    \"alliance\": \"red\",\n    \"parentNode\": \"root\",\n    \"points\": 5,\n    \"destroyGamepiece\": false,\n    \"persistentPoints\": true,\n    \"deltaTransformation\": [1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1]\n  }\n]`,
```
azaleacolburn: You don't need the `\` to escape the`"`, since the it's already a format string which uses the grave character as a delimiters instead. However, you don't need a format string for this (since you're not formatting any variables) and should probably switch to single quotes (`'`) instead.

azaleacolburn (dismissed):

rutmanz (changes requested):
I'm trying to test this but it doesn't actually seem to persist the changes? These are the steps I'm taking:

1. Import field
2. Add scoring zone
3. Go to devtool editor and click "save" (I see the toast that says they've been persisted)
4. Clear preferences (because scoring zone data is already scored here; I think part of this PR would also be ending the storage of scoring zones in the preferences cache so it can stay with the mirabuf file)
5. Reload
6. Import same field
7. No scoring zones are present

I'm not seeing any console errors

AlexD717:
> 4. Clear preferences (because scoring zone data is already scored here; I think part of this PR would also be ending the storage of scoring zones in the preferences cache so it can stay with the mirabuf file)

Wouldn't saving data back to the mira file require us to store the mira files for each individual users? The way I understood the PR (I might be wrong) the goal was to allow fields to come with pre-selected scoring zones. Meaning that in the exporter a new tab would be added saving this data to the mira file, that way when the user comes and tries to load a field all the scoring zones are already pre-configured.

ryanzhangofficial:
> I'm trying to test this but it doesn't actually seem to persist the changes? These are the steps I'm taking:
> 
> 1. Import field
> 2. Add scoring zone
> 3. Go to devtool editor and click "save" (I see the toast that says they've been persisted)
> 4. Clear preferences (because scoring zone data is already scored here; I think part of this PR would also be ending the storage of scoring zones in the preferences cache so it can stay with the mirabuf file)
> 5. Reload
> 6. Import same field
> 7. No scoring zones are present
> 
> I'm not seeing any console errors

Ah I didn't incorporate the "clear preferences" step on my interpretation of the caching deliverable of this ticket. The changes do persist after reloading if the preferences aren't cleared. However, I don't think it makes sense for the changes to still persist after preferences are cleared especially for default fields. I think your suggestion in `AARD-1914` on making the edited mira files exportable makes a lot of sense though (currently working on it). The other route would to be make it similar to the custom input panel, but I think that's out of the scope for this ticket.

rutmanz:
As the current behavior on `dev` is to save and restore scoring zones automatically, I don't know what feature this pull request would be adding if it doesn't load the scoring zones from the saved cached mirabuf files

fission/src/mirabuf/MirabufLoader.ts
ryanzhangofficial: I would argue that most of the lines in the current `try` block is fallible. For example, `const updatedBuffer = mirabuf.Assembly.encode(assembly).finish()` if the assembly is malformed.

ryanzhangofficial:
The issues should all be addressed.

rutmanz (changes requested):
it looks like you didn't update the type for DevtoolMiraData (it's still unknown), but other than that looks good

fission/src/mirabuf/MirabufLoader.ts
azaleacolburn: Would it be possible to reorder some of the statements such that all fallible statements are contiguous? Or at the very least move some of the infallible statements out of the `try` block.

rutmanz: What's the problem with having a large try block here?

ryanzhangofficial: I think it's more of a preference change rather than something that would benefit the code structure. I look at the try catch block and see that it does exactly what I want it to do.

rutmanz (approved):
lgtm

fission/src/mirabuf/MirabufLoader.ts
BrandonPacewic: I do agree that we should generally avoid having try blocks this large. At least for me coming from system programming this is usually a red flag that program design went wrong. But in this case I think it's totally fine, even if it may feel uncomfortable.

BrandonPacewic (approved):
Well done getting the integration of everything all lined up. Nice work.

fission/src/mirabuf/MirabufLoader.ts
ryanzhangofficial: Ok, I’ll keep that in mind for the future. Coming from machine learning/data science, I’m used to just throwing everything into Jupyter cells😂

PR #1212, opened 2025-07-14
Save Settings Logic [`AARD-1990`]

ryanzhangofficial:
## Task
Only save and show the "Settings Saved" toast when the save button is clicked, not when clicking outside of the modal.

[Jira Ticket](https://jira.autodesk.com/secure/RapidBoard.jspa?rapidView=13402&projectKey=AARD&view=detail&selectedIssue=AARD-1990#)

## Symptom
Before, changes to the General Settings tab saved when the user clicked outside of the `SettingsModal`. 

## Solution
Set `allowClickAway` to false.

## Verification
- SettingsModal doesn't close on `clickAway` anymore


fission/src/ui/components/Modal.tsx
```diff
@@ -24,7 +24,8 @@ type ModalProps = {
     cancelBlocked?: boolean
     middleBlocked?: boolean
     acceptBlocked?: boolean
-    allowClickAway?: boolean
+    disableClickAway?: boolean
+    allowClickAway?: boolean
```
rutmanz: why not just set `allowClickAway` to false? Why do we need two properties controlling the same thing?

rutmanz (changes requested):
LGTM from a UX perspective

fission/src/ui/components/Modal.tsx
ryanzhangofficial: yeah i was definitely overthinking it

rutmanz (approved):

BrandonPacewic (approved):

PR #1194, opened 2025-07-03
Exporter UI Refactor `[AARD-1883]`

rutmanz:
## Task

AARD-1883

Uses [Fusion's palette system](https://help.autodesk.com/view/fusion360/ENU/?guid=GUID-6C0C8148-98D0-4DBC-A4EC-D8E03A8A3B5B) to make the UI for the exporter a React webpage. Goal is to make the UI cleaner, easier to build on, and not such a pain to update. 

## Symptom
When you click the exporter button in fusion, you previously saw the old UI. The old UI was difficult to update and very inflexible in its layout.

## Solution

- Old `ConfigureCommand` GUI replaced with a palette, which has all of the same inputs re-implemented using MUI and React
- Uses the Fusion-Web communication protocol to transfer data at key moments
  - When selecting joints and game pieces (Web UI prompts Fusion to allow the user to select an appropriate item in the scene, Fusion returns the selected items)
  - When first opening the panel (Fusion sends calculated robot weight)
  - When exporting (Web UI sends configuration to Fusion and initiates the export process)

### Photos
image

### Limitations
- Uses Fusion's internal units (rad/s, cm/s, kg, N) instead of user's unit preferences
	- This isn't a technical limitation, could update to account for that if we think it's important
- 
## Verification

- Confirmed that `ExporterOptions` objects were the same with both the old configuration UI and the new web UI
- Confirmed that saving and loading of config in the document metadata worked
- Confirmed that joint selector and gamepiece selectors allow you to select joints and gamepieces
- Confirmed that models can be imported

---

Before merging, ensure the following criteria are met:

- [ ] All acceptance criteria outlined in the ticket are met.
- [ ] Necessary test cases have been added and updated.
- [ ] A feature toggle or safe disable path has been added (if applicable).
- [ ] User-facing polish:
  - Ask: *"Is this ready-looking?"*
- [ ] Cross-linking between Jira and GitHub:
  - PR links to the relevant Jira issue.
  - Jira ticket has a comment referencing this PR.

PR #1214, opened 2025-07-14
Exporter DRC UI

Dhruv-0-Arora:
## Task

Adding the ability for users to view whether their robot design meets the guidelines outlined in the [FRC inspection checklist](https://firstfrc.blob.core.windows.net/frc2025/Manual/2025FRCInspectionChecklist.pdf).

AARD-1882

## Solution

Created an Exporter tab that shows the user the characteristics of their design and cross check them with the FRC Inspection Checklist.

Screenshot 2025-07-18 at 3 08 05 PM

---

- [ ] User-facing polish:


rutmanz (changes requested):
ImageI like the font on these

The UI here doesn't seem as extensible/modifiable in the future as it could be. Ideally adding another rule would be as simple as adding the calculation function, a name, and the max value to an array, rather than require adding new inputs manually

This also isn't necessarily part of this PR because it's mostly for the UI, but I think these calculations are just not correct. My calculated height for this "robot" is about 30cm but it's reporting over 150cm. It also appears to use the bounding box of the robot (not the length of the outermost surface) which causes it to underreport non-rectangles.

Image

Would also probably be nice to allow max values to change from year to year, as they often do

rutmanz:
This appears to have made it worse (?). Synthesis and fusion swap the meaning of Y and Z (fusion has XY as the floor, but fission has XZ), and so the meaning of height for a robot that is later imported into synthesis is kind of weird (also the robot levitation appears to be based on the "down" as defined by fusion -Y, so rotating doesn't just fix it)
image

rutmanz (approved):
I think it still has the problem with concave frame perimeters (and also bumpers I suppose) but that's probably out of scope (and also maybe something we just don't deal with). lgtm

BrandonPacewic (changes requested):
- Needs updating towards #1194.

exporter/SynthesisFusionAddin/src/UI/ConfigCommand.py
```diff
@@ -299,6 +300,9 @@ def notify(self, html_args: adsk.core.HTMLEventArgs) -> None:
         elif html_args.action == "cancelSelection":
             gm.ui.terminateActiveCommand()
             html_args.returnData = "{}"
+
+        elif html_args.action == "designRules":
+            html_args.returnData = DesignRuleChecks().getDesignRules
```
rutmanz: This needs to be dumped as json

rutmanz (changes requested):
Nothing shows up in the DRC tab when I run this

rutmanz (changes requested):
This is what happens when I load the page on an existing assembly in fusion

Image

rutmanz (commented):
So at this point I think the UI is okay (could do with some colored chips around valid/invalid) but the DRC requirements change every year, and while making rulesets year-based might be out of scope for this ticket, it's so close to the end of the summer that idk if it makes sense to include the DRC ui at all because without that it's kind of a half-finished feature IMO.

BrandonPacewic:
Will not be finished as a part of the v7.2 release

PR #1179, opened 2025-06-27
Bootstrap Unit Testing

azaleacolburn:
### Description
This is a branch off of #1175 which I can actually edit to make fixes, since @ariesninjadev was working from a personal fork.

The tests are not my work, I have just fixed the formatting and typing issues.

[JIRA Issue](https://jira.autodesk.com/browse/AARD-1942)

PR #1217, opened 2025-07-15
Scene System Testing `[AARD-1938]`

AlexD717:
## Task
Create unit tests for each element found in `fission/src/systems/scene/`. 

AARD-1938

## Symptom

No unit tests currently exist for elements in `fission/src/systems/scene/`.

## Solution
Create unit test for each element in `fission/src/systems/scene/`.
_Note: No unit tests were created for SceneObject.ts due to it being very simple and other tests already testing its functionality._

## Verification
All unit tests pass

---

- [x] User-facing polish:


fission/src/test/scene/CameraControls.test.ts
```diff
@@ -0,0 +1,235 @@
...
+            expect(controls.enabled).toBe(true)
+            expect(controls.controlsType).toBe("Orbit")
+            expect(controls.locked).toBe(false)
+            expect(controls.focusProvider).toBeUndefined()
+            expect(camera.fov).toBe(75)
+            expect(camera.aspect).toBe(1)
+            expect(camera.near).toBe(0.1)
+            expect(camera.far).toBe(1000)
+            expect(camera.position.equals(new THREE.Vector3(0, 0, 5))).toBe(true)
+        })
```
rutmanz: This doesn't feel like something that actually should be unit tested IMO. If this fails it'd be because someone changed the defaults and they shouldn't also have to change a unit test when they change what the default camera position is

fission/src/test/scene/CameraControls.test.ts
```diff
@@ -0,0 +1,235 @@
...
+                controls.update(1 / 60)
+            }
+
+            // Camera should have moved from initial position
+            expect(camera.position.equals(initialPosition)).toBe(false)
+
+            // Camera should be moving towards the expected distance (within reasonable range)
+            const distance = camera.position.distanceTo(new THREE.Vector3(0, 0, 0))
+            expect(distance).toBeGreaterThan(3.0) // Should be somewhere between 3 and 6
+            expect(distance).toBeLessThan(6.0)
```
rutmanz: shouldn't it be exactly 5? (within rounding)

fission/src/test/scene/DragModeSystem.test.ts
```diff
@@ -0,0 +1,303 @@
...
+
+            // Mock the physics system to return our association
+            const originalGetBodyAssociation = physicsSystem.getBodyAssociation
+            physicsSystem.getBodyAssociation = vi.fn().mockReturnValue(mockAssociation)
+
+            // Mock the raycasting to return our cube when clicked
+            const originalRayCast = physicsSystem.rayCast
+            physicsSystem.rayCast = vi.fn().mockReturnValue({
+                data: { mBodyID: bodyId },
+                point: { GetX: () => 0, GetY: () => 0, GetZ: () => 0 },
```
rutmanz, line 174: probably want to suppress eslint here

fission/src/test/scene/DragModeSystem.test.ts
```diff
@@ -0,0 +1,303 @@
...
+            const wheelEvent = new WheelEvent("wheel", {
+                deltaY: 100,
+                bubbles: true,
+                cancelable: true,
+            })
+
+            // This should not throw even without active drag
+            expect(() => {
+                window.dispatchEvent(wheelEvent)
+            }).not.toThrow()
```
rutmanz: when would window.dispatchEvent ever possibly throw?

fission/src/test/scene/GizmoSceneObject.test.ts
```diff
@@ -0,0 +1,280 @@
...
+
+            gizmoSceneObject.setTransform(transform)
+
+            // Test that position is applied correctly
+            expect(gizmoSceneObject.obj.position.x).toBeCloseTo(targetPosition.x)
+            expect(gizmoSceneObject.obj.position.y).toBeCloseTo(targetPosition.y)
+            expect(gizmoSceneObject.obj.position.z).toBeCloseTo(targetPosition.z)
+
+            // Test that the position vector matches
+            expect(gizmoSceneObject.obj.position.equals(targetPosition)).toBe(true)
```
rutmanz: ?

fission/src/test/scene/GizmoSceneObject.test.ts
```diff
@@ -0,0 +1,280 @@
...
+        test("should apply rotation transformation correctly", () => {
+            const rotationAngle = Math.PI / 4 // 45 degrees
+            const rotationAxis = new THREE.Vector3(0, 1, 0) // Y-axis
+            const transform = new THREE.Matrix4().makeRotationAxis(rotationAxis, rotationAngle)
+
+            gizmoSceneObject.setTransform(transform)
+
+            // Test that rotation is applied correctly
+            expect(gizmoSceneObject.obj.rotation.y).toBeCloseTo(rotationAngle)
+            expect(Math.abs(gizmoSceneObject.obj.rotation.x)).toBeLessThan(0.001)
```
rutmanz: What is with the inconsistent use of .toBeCloseTo vs .toBeLessthan for checking that things are almost exactly the same?

fission/src/test/scene/GizmoSceneObject.test.ts
```diff
@@ -0,0 +1,280 @@
...
+
+            gizmoSceneObject.setTransform(identityTransform)
+
+            // Should set to origin position and zero rotation
+            expect(gizmoSceneObject.obj.position.x).toBeCloseTo(0)
+            expect(gizmoSceneObject.obj.position.y).toBeCloseTo(0)
+            expect(gizmoSceneObject.obj.position.z).toBeCloseTo(0)
+            expect(gizmoSceneObject.obj.rotation.x).toBeCloseTo(0)
+            expect(gizmoSceneObject.obj.rotation.y).toBeCloseTo(0)
+            expect(gizmoSceneObject.obj.rotation.z).toBeCloseTo(0)
```
rutmanz: why would this test fail when the rest succeed? I guess I'm not seeing what kinds of errors a lot of these tests are meant to catch.

rutmanz (changes requested):
I'm only about halfway done reading through this but I'm concerned that a lot of the tests here aren't really checking for things that matter. Like #1179, it kind of feels like these might be AI generated, and I'm not sure what a lot of them are doing

AlexD717:
> I'm only about halfway done reading through this but I'm concerned that a lot of the tests here aren't really checking for things that matter. Like #1179, it kind of feels like these might be AI generated, and I'm not sure what a lot of them are doing

I did use AI to help make these unit tests, and I thought that I looked through them and fixed all the mistakes. I have clearly missed some, so I will go through again and double check that everything is good.

fission/src/systems/scene/CameraControls.ts
```diff
@@ -203,12 +203,6 @@ export class CustomOrbitControls extends CameraControls {
         return { ...this._coords }
     }
 
-    public setTargetCoordinates(coords: Partial<SphericalCoords>) {
-        if (coords.theta !== undefined) this._nextCoords.theta = coords.theta
-        if (coords.phi !== undefined) this._nextCoords.phi = coords.phi
-        if (coords.r !== undefined) this._nextCoords.r = coords.r
-    }
-
     public setImmediateCoordinates(coords: Partial<SphericalCoords>) {
```
Dhruv-0-Arora, line 206: Why are we removing this function?

AlexD717: This function was added in the view cube pr, but it isn't actually used anywhere in the code. It also only moves the camera towards the target coordinates for only one frame and then stops.

fission/src/test/scene/SceneInteractionHandler.test.ts
```diff
@@ -0,0 +1,340 @@
...
+
+        expect(mockElement.removeEventListener).toHaveBeenCalledWith("pointermove", expect.any(Function))
+        expect(mockElement.removeEventListener).toHaveBeenCalledWith("wheel", expect.any(Function))
+        expect(mockElement.removeEventListener).toHaveBeenCalledWith("contextmenu", expect.any(Function))
+        expect(mockElement.removeEventListener).toHaveBeenCalledWith("pointerdown", expect.any(Function))
+        expect(mockElement.removeEventListener).toHaveBeenCalledWith("pointerup", expect.any(Function))
+        expect(mockElement.removeEventListener).toHaveBeenCalledWith("pointercancel", expect.any(Function))
+        expect(mockElement.removeEventListener).toHaveBeenCalledWith("pointerleave", expect.any(Function))
+        expect(mockElement.removeEventListener).toHaveBeenCalledWith("touchmove", expect.any(Function))
+    })
```
ryanzhangofficial, line 74: I don't really see the need to test all event listeners. Picking a representative subset (e.g. one pointer event and one wheel event) is enough; the rest seems redundant.

fission/src/test/scene/DragModeSystem.test.ts
```diff
@@ -0,0 +1,272 @@
...
+
+            expect(dragModeSystem.enabled).toBe(false)
+            expect(removeEventListenerSpy).toHaveBeenCalledWith("disableDragMode", expect.any(Function))
+
+            removeEventListenerSpy.mockRestore()
+        })
+    })
+
+    describe("Physics Integration", () => {
+        test("should actually drag and move a cube and respond to wheel scroll", () => {
```
ryanzhangofficial: Can this be split into different tests rather than one single bulky one so it's easier to maintain?

fission/src/test/scene/SceneInteractionHandler.test.ts
AlexD717: This is basically a sanity check that all events where properly set up and then disposed. We want to avoid a scenario where someone accidentally deleted the setup/dispose code for one of the events, or added some logical that made it so that the events weren't being set up.

fission/src/test/scene/DragModeSystem.test.ts
AlexD717: Done

fission/src/test/scene/DragModeSystem.test.ts
```diff
@@ -0,0 +1,306 @@
...
+                position: [400, 300] as [number, number],
+            }
+
+            screenHandler.interactionEnd?.(endInteraction)
+
+            cleanup()
+        })
+
+        test("should move cube away from camera when wheel scrolled during drag", () => {
+            const { physicsBody, cleanup } = setupDraggableCube()
```
rutmanz, line 243: Why isn't this just a beforeEach() and afterEach()

fission/src/test/scene/SceneRenderer.test.ts
```diff
@@ -0,0 +1,509 @@
...
+
+        test("should create sphere with default material", () => {
+            const sphere = sceneRenderer.createSphere(1.0)
+            expect(sphere.material).toBeInstanceOf(THREE.MeshToonMaterial)
+        })
+
+        test("should create box with default material and correct position", () => {
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            const mockVec3: MockVec3 = { GetX: () => 2, GetY: () => 3, GetZ: () => 4 }
+            const box = sceneRenderer.createBox(mockVec3 as unknown as Parameters<typeof sceneRenderer.createBox>[0])
```
rutmanz: Is there a reason this isn't just an actual Vec3?

fission/src/test/scene/SceneRenderer.test.ts
```diff
@@ -0,0 +1,509 @@
...
+            const worldPos = sceneRenderer.pixelToWorldSpace(centerX, centerY)
+
+            expect(worldPos.x).toBe(0)
+            expect(worldPos.y).toBe(0)
+        })
+
+        test("should convert world to pixel space", () => {
+            const pixelPos = sceneRenderer.worldToPixelSpace(new THREE.Vector3(0, 0, 0))
+            expect(pixelPos[0]).toBeCloseTo(1861, 0)
+            expect(pixelPos[1]).toBeCloseTo(1261, 0)
```
rutmanz: Shouldn't these be based off of the window.innerWidth and window.innerHeight property like the previous test?

fission/src/test/scene/SceneRenderer.test.ts
```diff
@@ -0,0 +1,509 @@
...
+    })
+
+    describe("Canvas Management", () => {
+        test("should update camera aspect ratio based on window size", () => {
+            // Windows size is already set to 1920x1080
+            sceneRenderer.updateCanvasSize()
+
+            const aspectRatio = 1920 / 1080
+            expect(sceneRenderer.mainCamera.aspect).toBeCloseTo(aspectRatio)
+            expect(sceneRenderer.mainCamera.fov).toBeCloseTo(110 / aspectRatio)
```
rutmanz: I don't think having magic numbers in test files is a good idea, if this is a constant defined elsewhere it should probably use that definition

fission/src/test/scene/SceneRenderer.test.ts
```diff
@@ -0,0 +1,509 @@
...
+            expect(noDirectionalLight).toBeUndefined()
+
+            sceneRenderer.changeLighting(false)
+            const newDirectionalLight = sceneRenderer.scene.children.find(
+                child => child instanceof THREE.DirectionalLight
+            )
+            expect(newDirectionalLight).toBeInstanceOf(THREE.DirectionalLight)
+        })
+
+        test("should handle null material gracefully", () => {
```
rutmanz, line 287: this exclusively tests the 3JS CSM functions, not our code

fission/src/test/scene/SceneRenderer.test.ts
```diff
@@ -0,0 +1,509 @@
...
+            sceneRenderer.changeLighting(true)
+
+            expect(() => sceneRenderer.setupMaterial(null as unknown as THREE.Material)).not.toThrow()
+        })
+    })
+
+    describe("Scene Management", () => {
```
rutmanz: All of the Scene Management tests appear to exclusively test library code in THREE.Scene, not Synthesis code

rutmanz (changes requested):

fission/src/test/scene/DragModeSystem.test.ts
AlexD717: It would be more lines of code and I think that the current approach is cleaner. It just doesn't make sense to do this only for 2 functions.

fission/src/test/scene/SceneRenderer.test.ts
AlexD717: This verifies that our code works if a null parameters is passed in as a setup material

rutmanz (approved):

ryanzhangofficial (approved):
Looks great!

BrandonPacewic (approved):

PR #1219, opened 2025-07-15
Refactor Input System `[AARD-1997]`

rutmanz:
## Task

AARD-1997

This was something I found myself doing in preparation for handling swerve inputs, but thought it would make more sense as a separate PR. Makes the input system more effectively typed (rather than just using strings), cleans up some of the syntax, adds dynamic conflict checking by used keys, and adds support for tank input schemes (which makes swerve input schemes easier down the road).

## Symptom

Easy to use invalid input names or key codes, the syntax for constructing new axis and button inputs was very verbose and required passing a lot of unused parameters

## Solution

- Tank controls were added in preparation for supporting swerve controls 
- String literal types used for input names and key names
- Static construction methods on AxisInput and ButtonInput
- Conflict identification for active schemes to indicate when buttons will overlap with an existing scheme 

## Verification

- Verify that you can configure robots and assign input schemes
- Verify that you can create new input schemes
- Verify that switching between drivetrain types and schemes works
- Verify that schemes with conflicts (including custom schemes) are indicated when selecting schemes

---

- [ ] User-facing polish:


fission/src/systems/input/InputSchemeManager.ts
```diff
@@ -99,17 +110,52 @@ class InputSchemeManager {
...
+        }
+
+        allSchemes.forEach(scheme => {
+            if (scheme.inputs.some(input => input.keysUsed.some(k => usedKeyMap.has(k)))) {
+                result[scheme.schemeName] ??= { scheme, status: InputSchemeUseType.CONFLICT }
+            } else {
+                result[scheme.schemeName] = { scheme, status: InputSchemeUseType.AVAILABLE }
+            }
+        })
+        console.log(result)
```
AlexD717:
```suggestion
```
Was this left here intentionally or was this supposed to have been removed?

fission/src/systems/input/InputSchemeManager.ts
```diff
@@ -99,17 +110,51 @@ class InputSchemeManager {
...
+        const usedKeyMap = new Set<KeyDescriptor>()
+        const result: Record<string, InputSchemeAvailability> = {}
+        for (const scheme of InputSystem.brainIndexSchemeMap.values()) {
+            result[scheme.schemeName] = { scheme, status: InputSchemeUseType.IN_USE }
+            scheme?.inputs?.forEach(input => {
+                input.keysUsed.forEach(key => {
+                    if (key != null) {
+                        usedKeyMap.add(key)
+                    }
+                })
```
azaleacolburn: This looks like a job for the `filter` function

fission/src/test/InputSystem.test.ts
```diff
@@ -251,12 +260,36 @@ describe("Gamepad Input Check", () => {
...
+        names.forEach(name => {
+            expect.soft(names.filter(other => other == name).length, `Only one schema named ${name}`).toBe(1)
+        })
+    })
+    test("Default schemes internally conflict-free", () => {
+        DefaultInputs.defaultInputCopies.forEach(scheme => {
+            const usedKeys = new Map<KeyDescriptor, number>()
+            scheme.inputs.forEach(input => {
+                input.keysUsed.forEach(key => {
+                    if (key == null) return
```
azaleacolburn: This should probably be filtered instead.

fission/src/ui/panels/configuring/ChooseInputSchemePanel.tsx
```diff
@@ -36,7 +36,9 @@ const ChooseInputSchemePanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
             if (brainIndex == undefined) return
             if (InputSystem.brainIndexSchemeMap.has(brainIndex)) return
 
-            const scheme = InputSchemeManager.availableInputSchemes[0]
+            const scheme = InputSchemeManager.availableInputSchemesByBrain(brainIndex).filter(
+                scheme => scheme.status == InputSchemeUseType.AVAILABLE
+            )[0]?.scheme
```
azaleacolburn: This could probably just be the `find` function instead.

fission/src/ui/panels/configuring/initial-config/InitialConfigPanel.tsx
```diff
@@ -58,7 +58,10 @@ const InitialConfigPanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
             if (brainIndex == undefined) return
             if (InputSystem.brainIndexSchemeMap.has(brainIndex)) return
 
-            const scheme = InputSchemeManager.availableInputSchemes[0]
+            const scheme = InputSchemeManager.availableInputSchemesByBrain(brainIndex).filter(
+                scheme => scheme.status == InputSchemeUseType.AVAILABLE
+            )[0]?.scheme
```
azaleacolburn: Same comment as before about the using `find` instead.

azaleacolburn (changes requested):

PepperLola (changes requested):
Previously when importing a robot and setting the control scheme from that panel, clicking the check would remove that row from the control scheme list, whereas now it stays there. I think it makes more sense for it to be removed because it makes it clearer that the control scheme has been applied.

rutmanz:
@PepperLola I didn't remove them from the list because I always found that kind of unintuitive, but I did add tooltips and make their buttons clearly disabled. Does this seem good?

AlexD717 (approved):
Looks good!

Once this is merged in, i think it would be beneficial to make a ticket to add some sort of indicator if you modify an input scheme causing it to overlap with another scheme (in the configure assets panel).

azaleacolburn (approved):
lgtm

PepperLola:
Are there meant to be two dividers between the available and in-use schemes?

rutmanz:
@PepperLola There are technically three categories (hence two dividers, select Carmella to see), but I've fixed so they only show if the categories aren't empty

PepperLola (approved):
looks good 🥇

PR #1221, opened 2025-07-17
Speed Up Playwright `[AARD-1989]`

rutmanz:
## Task

AARD-1989

Figure out a solution for preventing playwright installation from taking so long

## Symptom

Playwright dependencies can't be cached easily, so unit test actions had to install fresh each time. This meant download times of 1-2 minutes best case, and some runs lasted 10-30 minutes for reasons unknown

## Solution

I switched to the playwright docker image, which is prebuilt and doesn't have such long install times (server for the image appears to also be on Azure, where GH actions run), this makes the playwright install take about 30s instead (very cool).

## Verification

Workflow succeeds (as much as it does elsewhere, currently something weird going on with one of the tests), times are much faster

---

- [x] User-facing polish:

### DIFF
--- a/.github/workflows/FissionUnitTest.yml
+++ b/.github/workflows/FissionUnitTest.yml
@@ -10,8 +10,13 @@ on:
 jobs:
 
   runUnitTests:
-    name: Unit Tests
+    name: Playwright Unit Tests
+    container:
+      image: mcr.microsoft.com/playwright:v1.54.0-noble
     runs-on: ubuntu-latest
+    defaults:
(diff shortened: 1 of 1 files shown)


rutmanz:
~~(Currently not passing due to unknown unit test problems)~~

looks to be fine now

azaleacolburn (approved):
lgtm

AlexD717 (approved):

PepperLola (approved):
Looks good

PR #1224, opened 2025-07-17
Migrate from Prettier and Eslint to Biome `[AARD-1998]`

azaleacolburn:
## Task

AARD-1998

Switch from Prettier and ESLint to Biome, fix issues that Biome recognized by default but ESLint doesn't along the way, and refactor `package.json` scripts.

## Symptom

- Formatting and linting being handled by separate tools is sub-optimal.
- There were several cases of bad practices hidden behind `eslint-disable`.
- The fission package's formatting and linting scripts were nonstandard and confusing.
- Using multiple package managers for the same project can lead to discontinuities in workflows and is generally annoying to manage.

## Solution

- Switched to Biome
- Properly typed `APSDataManagement` and fixed various other bad practices on other files.
- When needed, either removed or converted old `eslint-disable` comments to `biome-ignore` comments.
- Refactored `package.json` scripts to `lint` for linting, `fmt` for formatting, and `style` for both, each of which can be appended with `:fix` to apply changes automatically when possible.
- Switched scripts to exclusively use [Bun](https://bun.sh), since we've already standardized on only committing `bun.lock`

> [!NOTE]
> Linting fixes which may change the semantic value of the statement or expression must be applied explicitly with the flag `--unsafe`.
> 
> Migrating NPM should use `bun add <package>` instead of `bun install <package>`, although both will work in most cases.

## Verification

- `bun run style` outputs no errors.
- The project builds
- All tests pass

---

- [ ] User-facing polish:


itspvty (approved):
LGTM

.github/workflows/FissionBiome.yml
```diff
@@ -0,0 +1,48 @@
+name: Fission - ES Lint Format Validation
```
PepperLola: Should this be changed to Biome?

.github/workflows/FissionBiome.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  workflow_dispatch: {}
+  push:
+    branches: [prod, dev]
+  pull_request:
+    branches: [prod, dev]
+
+jobs:
+  runFormatValidationScript:
+    name: ESLint Format Validation
```
PepperLola: Biome?

.github/workflows/FissionBiome.yml
```diff
@@ -0,0 +1,48 @@
...
+          node-version: latest
+
+      - name: Cache Dependencies
+        uses: actions/cache@v3
+        with:
+          key: "${{runner.os}}-npm-fission-${{hashFiles('fission/package.json')}}"
+          path: "fission/node_modules"
+          restore-keys: |
+            ${{runner.os}}-npm-fission-
+            ${{runner.os}}-npm
```
PepperLola, line 29: Bun instead of NPM? not that it really matters

.github/workflows/FissionBiome.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Linter
+        id: linter-validation
+        if: ${{ always() }}
+        run: |
+          cd fission
+          bun run lint && echo "Biome Linting Validation Passed" || (echo "Biome Linting Validation Failed" && exit 1)
+
+      - name: Prettier
+        id: prettier-validation
```
PepperLola: Format instead of Prettier?

.github/workflows/FissionBuild.yml
```diff
@@ -13,25 +13,25 @@ jobs:
...
 
       - name: Cache Node Dependencies
         uses: actions/cache@v3
         with:
           key: "${{runner.os}}-npm-fission-${{hashFiles('fission/package.json')}}"
-          path: 'fission/node_modules'
+          path: "fission/node_modules"
           restore-keys: |
             ${{runner.os}}-npm-fission-
             ${{runner.os}}-npm
```
PepperLola, line 27: Bun?

.github/workflows/FissionPackage.yml
```diff
@@ -3,62 +3,62 @@ name: Fission - Package
...
-          ${{runner.os}}-npm-fission-
-          ${{runner.os}}-npm
+      - name: Cache Dependencies
+        uses: actions/cache@v3
+        with:
+          key: "${{runner.os}}-npm-fission-${{hashFiles('fission/package.json')}}"
+          path: "fission/node_modules"
+          restore-keys: |
+            ${{runner.os}}-npm-fission-
+            ${{runner.os}}-npm
```
PepperLola, line 32: Bun?

.github/workflows/FissionUnitTest.yml
```diff
@@ -52,17 +52,17 @@ jobs:
...
-      - name: Cache Node Dependencies
+      - name: Cache Dependencies
         uses: actions/cache@v3
         with:
           key: "${{runner.os}}-npm-fission-${{hashFiles('fission/package.json')}}"
-          path: 'fission/node_modules'
+          path: "fission/node_modules"
           restore-keys: |
             ${{runner.os}}-npm-fission-
             ${{runner.os}}-npm
```
PepperLola, line 53: Bun?

fission/bun.lock
```diff
@@ -712,7 +729,7 @@
 
     "dlv": ["dlv@1.1.3", "", {}, "sha512-+HlytyjlPKnIG8XuRG8WvmBP8xs8P71y+SKKS6ZXWoEgLuePxtDoUEiH7WkdePWrQ5JBpE6aoVqfZfJUQkjXwA=="],
 
-    "doctrine": ["doctrine@3.0.0", "", { "dependencies": { "esutils": "^2.0.2" } }, "sha512-yS+Q5i3hBf7GBkd4KG8a7eBNNWNGLTaEwwYWUijIYM7zrlYDM0BFXHjjPWlWZ1Rg7UaddZeIDmi9jF3HmqiQ2w=="],
+    "doctrine": ["doctrine@2.1.0", "", { "dependencies": { "esutils": "^2.0.2" } }, "sha512-35mSku4ZXK0vfCuHEDAwt55dg2jNajHZ1odvF+8SSr82EsZY4QmXfuWso8oEd8zRhVObSN18aM0CjSdoBX7zIw=="],
```
PepperLola, line 734: What happened here? Is this a biome dependency?

fission/src/Synthesis.tsx
```diff
@@ -134,7 +134,7 @@ const Synthesis: React.FC = () => {
         }
 
         // eslint-disable-next-line react-hooks/exhaustive-deps
```
PepperLola: Should we remove the eslint disable directives? They should be unnecessary now and are just cluttering the code

fission/src/aps/APSDataManagement.ts
```diff
@@ -58,15 +57,24 @@ export type DataAttributes = {
...
 
+export type Relationships = {
+    storage: { meta: { link: { href?: string } } }
+    parent: { data: { id?: string } }
+    rootFolder: { data: RawData }
+}
+
+export type RawData = Omit<{ [key in keyof Data]: Data[key] }, "raw" | "href"> & { relationships: Relationships }
+
+// biome-ignore-start lint/suspicious/noExplicitAny: We need to index a generic object
```
PepperLola: Delete now that we aren't using `any`?

fission/src/util/TypeConversions.ts
```diff
@@ -18,10 +18,22 @@ export function convertArrayToThreeMatrix4(arr: number[]) {
     // DO NOT ask me why retrieving and setting the same EXACT data is done is two DIFFERENT majors
     // prettier-ignore
     return new THREE.Matrix4(
-        arr[0], arr[4], arr[8], arr[12],
```
PepperLola, line 21: :( is there a way to have it skip formatting this part? I like being able to visualize the matrix more easily

fission/src/util/dom.ts
```diff
@@ -23,10 +22,11 @@ export const mousePosition = (x: number, y: number) => {
     el?.dispatchEvent(event)
 }
 
-// prettier-ignore
+// biome-ignore-start lint/suspicious/noExplicitAny: We need to index a generic object
 export const addGlobalFunc = <T>(name: string, func: (...args: any[]) => T) => {
-    (window as any)[name] = func
+    ;(window as any)[name] = func
```
PepperLola: Is there any way around this syntax?

.github/workflows/FissionBiome.yml
azaleacolburn: Honestly, I'm not entirely sure if that matters, I think it's just the name of the keys, so I don't see a reason to change them. @rutmanz knows more about workflows than I do and could probably tell me I'm wrong.

.github/workflows/FissionBuild.yml
azaleacolburn: Bun doesn't have its own version of `node_modules`.

.github/workflows/FissionBiome.yml
rutmanz: It's just the cache key and it doesn't really matter, I don't really see a need to change it

.github/workflows/FissionPackage.yml
rutmanz: This is good for bun

fission/src/Synthesis.tsx
rutmanz: I think yes

fission/src/aps/APSDataManagement.ts
rutmanz: Is there a way to make biome report unused ignore rules

fission/src/util/dom.ts
rutmanz: Could update where the properties of window are declared to include these properties and not require casting, which seems like the better solution overall because it doesn't make sense to define arbitrary global functions (either we know what they are and use them and therefore want types or we don't care what they are and they're unused)

.github/workflows/FissionUnitTest.yml
```diff
@@ -20,11 +19,14 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
-      - name: JavaScript Setup
-        uses: actions/setup-node@v4
+      - name: Update and Install Unzip
```
rutmanz: This was intentionally part of the asset downloading step because it's slow and should not run (unzip isn't needed) if the assets are already cached.

.github/workflows/FissionUnitTest.yml
```diff
@@ -52,17 +52,17 @@ jobs:
       #       ~/.cache/ms-playwright/
       #     key: ${{ runner.os }}-assets-playwright-${{ env.PLAYWRIGHT_VERSION }}-v3
 
-      - name: Cache Node Dependencies
+      - name: Cache Dependencies
```
rutmanz: I agree that it's no longer Node from a technical perspective but just "dependencies" is a lot more ambiguous. It's still called node_modules fwiw

fission/src/mirabuf/MirabufInstance.ts
```diff
@@ -115,7 +115,7 @@ class MirabufInstance {
         this._mirabufParser = parser
         this._materials = new Map()
         this._meshes = new Map()
-        this._batches = new Array<THREE.BatchedMesh>()
+        this._batches = [] as THREE.BatchedMesh[]
```
rutmanz: Why are we casting?

rutmanz (changes requested):
It feels like the bun stuff is out of scope for this PR. I recall Brandon saying that he wanted to do a poll/check with the whole team about fully committing to Bun

fission/src/util/dom.ts
azaleacolburn: We could ignore formatting here

.github/workflows/FissionBiome.yml
```diff
@@ -33,15 +33,15 @@ jobs:
...
+        id: lint-validation
         if: ${{ always() }}
         run: |
           cd fission
           bun run lint && echo "Biome Linting Validation Passed" || (echo "Biome Linting Validation Failed" && exit 1)
 
-      - name: Prettier
-        id: prettier-validation
+      - name: Format
+        id: format-validation
```
rutmanz, line 44: IDs not necessary (and kind of confusing imo) unless they're referenced later

rutmanz:
Also I'm pretty sure Bun is making your unit tests fail fwiw

.github/workflows/FissionUnitTest.yml
azaleacolburn: For some reason, because `bun-setup` expects `unzip`. I'm a bit confused about this, since there wasn't an issue on the other actions. [Here's the action](https://github.com/Autodesk/synthesis/actions/runs/16357681832/job/46219934177?pr=1224) where it couldn't find `unzip`.

.github/workflows/FissionUnitTest.yml
azaleacolburn: Ok, I'm happy to revert it.

fission/src/mirabuf/MirabufInstance.ts
azaleacolburn: This was probably an unsafe fix that was applied by biome, I can tweak the configuration to prevent this.

.github/workflows/FissionBiome.yml
azaleacolburn: I'll remove them

.github/workflows/FissionUnitTest.yml
rutmanz: Unit tests need unzip specifically because of the playwright container, it's normally preinstalled. Requiring the installation I think makes the bun version of the workflow about a net 30s slower

azaleacolburn: Wait I'm confused, we were installing `unzip` with `apt` before, did we not just change where we're doing so?

azaleacolburn:
> Also I'm pretty sure Bun is making your unit tests fail fwiw

Ya, it's timing out for some reason

rutmanz:
oh actually I did more looking and it might be unrelated. Working on a fix for the timeout now

.github/workflows/FissionUnitTest.yml
azaleacolburn: Oh, it was only installed when the assets weren't cached, nevermind.

fission/src/mirabuf/MirabufInstance.ts
rutmanz: I only commented this once but it happens like 6 times in this file

Dhruv-0-Arora (commented):
I am getting this error when I try to go to the spawn asset panel

image

rutmanz (approved):

rutmanz:
make sure no eslint comments sneak in when you merge dev into this branch

Dhruv-0-Arora (changes requested):
I still have the error I mentioned before

fission/bun.lock
azaleacolburn: I presume so but am not really sure

fission/src/mirabuf/MirabufInstance.ts
```diff
@@ -240,7 +240,7 @@ class MirabufInstance {
 
                     let bodies = this._meshes.get(instance.info!.GUID!)
                     if (!bodies) {
-                        bodies = new Array<[THREE.BatchedMesh, number]>()
+                        bodies = [] as [THREE.BatchedMesh, number][]
```
rutmanz: oh these are still here

Dhruv-0-Arora (changes requested):
Screenshot 2025-07-21 at 1 53 08 PM

I can open the spawn assets panel but after I spawn a robot, I get this

fission/bun.lock
Dhruv-0-Arora: I couldn't really find any reason for this to be downgraded (specifically for biome) but there isn't many relevant changes for us between the versions.

fission/src/mirabuf/MirabufParser.ts
```diff
@@ -75,7 +75,7 @@ class MirabufParser {
 
     public constructor(assembly: mirabuf.Assembly, progressHandle?: ProgressHandle) {
         this._assembly = assembly
-        this._errors = new Array<ParseError>()
+        this._errors = []
```
Dhruv-0-Arora, line 78: Im assuming theres a reason you are getting rid of the typing for all the Arrays

Dhruv-0-Arora (approved):

fission/src/mirabuf/MirabufParser.ts
rutmanz: it's not getting rid of the typing. It's replacing array constructors with empty arrays (whose type is set by the type of `this._errors` and doesn't need to be explicitly specified)

Dhruv-0-Arora: icic; I didn't notice that `this._errors` already has a specified type

ryanzhangofficial (approved):

BrandonPacewic (changes requested):
Want to highlight @PepperLola's comments on bun vs npm in the workflows. Not sure if they should be resolved or addressed at this time.

Dhruv-0-Arora:
> Want to highlight @PepperLola's comments on bun vs npm in the workflows. Not sure if they should be resolved or addressed at this time.

Since it is a key, it doesn't effect anything to change from npm to bun.
Both Zach and Azalea both believe we should keep it as npm.

azaleacolburn:
> Want to highlight @PepperLola's comments on bun vs npm in the workflows. Not sure if they should be resolved or addressed at this time.

The name of the key doesn't actually matter, since we're still using NPM for some of the workflows and NPM is less ambiguous than Bun (since it could only mean one thing), @rutmanz and I don't think we should change it, unless we change it to something more generic like `node_modules`.

rutmanz:
I believe that calling it `${{runner.os}}-bun-fission-abcd` is less clear than `npm` because npm is specifically a package manager, but bun is also a runtime and test runner and build system. The old npm caches are compatible with the new bun ones (they're both just node_modules)

BrandonPacewic (approved):
Resolved conversation.

PR #1230, opened 2025-07-22
Height Limit Penalty `[AARD-1999]`

AlexD717:
## Task
Implement a penalty tracker that detects if the robot exceeded the maximum allowed height during match mode, in which case a penalty is applied.

AARD-1999

## Symptom

FRC has rules in certain years against extending past a certain height. 
## Solution

Implement a tracker that checks each robot height and verifies that they are under the height limit. If they are not a penalty is applied.
## Verification
To test this I changed on of the match configs to have a value of 1.5. The 2471 robot with elevator fully down is just slightly below that height. Extending the elevator should cause a penalty to be applied (if in match mode).
Also check that
- If ignore rotation is true, rotating the robot doesn't cause any penalties to be triggered
- Penalties still apply with multiple robots
- Penalties are not applied every frame
- Penalty can be configured by uploading a custom match mode config

---

- [x] User-facing polish:


azaleacolburn (approved):
lgtm

fission/src/util/UnitConversions.ts
```diff
@@ -0,0 +1,3 @@
+export function convertFeetToMeters(feet: number): number {
```
rutmanz, line 1: I think it would be a lot better to just define them in meters (or call this function where they're first used) rather than having feet exist in our data structures (with no real indication that they're feet and not meters

fission/src/ui/panels/configuring/MatchModeConfigPanel.tsx
```diff
@@ -23,6 +30,9 @@ export interface MatchModeConfig {
     autonomousTime: number // Optional, defaults to 15
     teleopTime: number // Optional, defaults to 135
     endgameTime: number // Optional, defaults to 20
+    ignoreRotation: boolean // Optional, defaults to true
+    maxHeight: number // Optional, defaults to infinity
+    heightPenalty: number // Optional, defaults to 2
```
rutmanz: I think it would be helpful to use the TS-style docstrings to explain what these mean
```ts
/**
Optional, defaults to true

If true, the height limit will be calculated relative to the base of the robot, rather than the base of the field
*/
ignoreRotation: boolean
```

fission/src/systems/match_mode/RobotDimensionTracker.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        const robots = [...sceneRenderer.sceneObjects.values()].filter(
+            (obj): obj is MirabufSceneObject => obj instanceof MirabufSceneObject && obj.miraType === MiraType.ROBOT
+        )
+
+        robots.forEach(robot => {
+            const dimensions = this._ignoreRotation ? robot.getDimensionsWithoutRotation() : robot.getDimensions()
+
+            if (dimensions.height > this._maxHeight) {
+                if ((this._robotHeightPenalties.get(robot.id) ?? 0) < Date.now() - PENALTY_COOLDOWN) {
```
rutmanz: This is typically not a time based penalty, and so I don't think that having it re-apply every second here is the best behavior. I think it would make more sense to only re-penalize once they go back under the height limit, so it's one penalty per "instance" of overextension

rutmanz (changes requested):

fission/src/systems/match_mode/RobotDimensionTracker.ts
AlexD717: I don't think that this rule allows someone to extend over max height and then stay they for the entire match while only getting one penalty. I am fairly sure that the judges will continue penalizing while the robot is over the height limit.

rutmanz: It doesn't come up often because this is mostly caught on inspection and not by the refs, but anecdotally it's typically a "every time you do something with it" penalty (like every time you contact something while overextended). A robot sitting in the corner of the field extended a foot over max height is not going to keep accruing penalties

AlexD717: I am unable to find anything definitive in the rules, and from what I found it is up to the referee how they want to do it. A robot sitting in the corner of the field isn't affecting gameplay and so the penalties wouldn't be applied. However a robot constantly staying above max height to score something/be useful would be penalized. Personally I think that a time based thing makes more sense, particularly because we are a simulator and so should remind the player constantly (also shows the worst case judge ruling decision compared to best case). 

I would like to here someone else's opinion on this, and if the majority thinks that every time you exceed is better, I will go in and implement that.

fission/src/util/UnitConversions.ts
AlexD717: I think that its better to have the user entered config values to be in feet, because the rules use feet, that way they don't have to spend extra time converting units manually. I will go through and convert to feet during the input processing phase though.

fission/src/systems/match_mode/DefaultMatchModeConfigs.ts
```diff
@@ -39,7 +39,7 @@ class DefaultMatchModeConfigs {
             teleopTime: 135,
             endgameTime: 30,
             ignoreRotation: true,
-            maxHeight: 6.5,
```
rutmanz: You could also do feetToMeters(6.5) right here but this is fine

rutmanz (approved):

fission/src/systems/match_mode/RobotDimensionTracker.ts
PepperLola: I personally think that per-incident makes more sense than periodically. Especially if we show a report of how many penalties a robot got at the end of the match, as a user I would expect that number to represent the number of times I exceeded the height limit (or incurred any other type of penalty). I think that the number is much more valuable if it represents a single thing (how many times we exceeded height limit) rather than multiple (how many times exceeded AND influenced by how long exceeded each time). I suppose we could decide on a more nuanced approach, like tracking contact, movement, or points scored while exceeding the height limit and try to imitate the way a ref would score the match, but I think in the meantime it would make the most sense to take the simplest approach. I would also like input from others before we decide one way or the other though.

azaleacolburn (approved):

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -632,6 +632,101 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+                const vertices = new Float32Array(
+                    JOLT.HEAP32.buffer,
+                    triangleContext.GetVerticesData(),
+                    triangleContext.GetVerticesSize() / Float32Array.BYTES_PER_ELEMENT
+                )
+
+                for (let i = 0; i < vertices.length; i += 3) {
+                    const vertex = new THREE.Vector3(vertices[i], vertices[i + 1], vertices[i + 2])
+
+                    vertex.applyMatrix4(bodyTransform).applyMatrix4(inverseRotation)
```
BrandonPacewic, line 704: What happens here if we rotate the robot with the gizmo before confirming its placement?

This would also break if we export something at an angle but I'm not sure if its worth fixing that or not.

fission/src/systems/match_mode/RobotDimensionTracker.ts
```diff
@@ -0,0 +1,41 @@
...
+        if (!MatchMode.getInstance().isMatchEnabled()) return
+
+        const robots = [...sceneRenderer.sceneObjects.values()].filter(
+            (obj): obj is MirabufSceneObject => obj instanceof MirabufSceneObject && obj.miraType === MiraType.ROBOT
+        )
+
+        robots.forEach(robot => {
+            const dimensions = this._ignoreRotation ? robot.getDimensionsWithoutRotation() : robot.getDimensions()
+
+            if (dimensions.height > this._maxHeight) {
```
BrandonPacewic: Consider adding a small epsilon buffer here. I understand that this much precision isn't really required for our application but its best practice.

fission/src/mirabuf/MirabufSceneObject.ts
AlexD717: Rotating the robot before confirming its placement has no difference.

Exporting something at an angle would break this. Would you like me to record the robot rotation when match mode starts and reset the rotation to that instead of the origin rotation?

rutmanz: exporting at angles also means the wheel levitation goes in the wrong direction fwiw, we might benefit from a more robust way of defining down generally

AlexD717: If robots exported at an angle don't drive properly, I think it would be better to look at that in a separate ticket

BrandonPacewic: > If robots exported at an angle don't drive properly, I think it would be better to look at that in a separate ticket

Agreed, will make note of this. Not relevant for this task. AARD-2019

BrandonPacewic (approved):
I'm running out of ways to say looks good. So. This looks good.

PR #1236, opened 2025-07-24
Camera Focus Fix `[AARD-2017]`

AlexD717:
## Task
Make sure that if an object exists that the camera can focus on, the camera will focus on that object.

AARD-2017

## Symptom
When a robot is already imported, trying to import another one and then canceling the import causes the camera to focus back on the origin rather than on the robot that was already in focus.

## Solution
Add a validate focus provider check that runs during the update loop on the camera. If it detects that the sceneRenderer no longer has the focusProvider or the focusProvider is undefined, it attempts to find a new object to focus on.

## Verification
- Spawning in a robot has the camera focus on it
-  - Spawning in a different robot but then canceling, results in the camera focusing on the original robot
- You can still right click on a robot and select unfocus, causing the camera to not be focused on anything

---

- [x] User-facing polish:


fission/src/systems/scene/CameraControls.ts
```diff
@@ -143,6 +145,31 @@ export class CustomOrbitControls extends CameraControls {
...
+    /**
+     * Validates that the current focus provider still exists in the scene.
+     * If not, automatically finds a suitable replacement.
+     */
+    private validateFocusProvider(): void {
+        if (this._focusProvider) {
+            const fallbackFocus = this.findFallbackFocus()
+            this._focusProvider = fallbackFocus
+        }
+    }
```
BrandonPacewic: Does this not unconditionally overwrite the focus provider even if its valid?

AlexD717: Fixed

BrandonPacewic (approved):
lgtm

rutmanz (approved):
lgtm

azaleacolburn (approved):
lgtm

PR #1190, opened 2025-07-02
Game Piece Asset Support `[AARD-1902]`

azaleacolburn:
### Description
Game pieces are now treated as separate assets from the field, they are able to be spawned, deleted, cached, etc, all on their own.

### Objectives
- [x] Frontend configuration menu for game pieces
- [x] Game pieces are returned as separate assemblies
- [x] Game pieces are pruned from the main assembly
- [x] Game pieces are treated like separate entities, neither fields nor robots and are configurable
- [x] Game pieces still start in the same place on the field as before
- [x] Game pieces cache properly
- [x] Game pieces can be interacted with by robots as before
- [x] Game pieces work in match mode as before

### Testing Done
- [x] Parsing fields (2023 & 2024) still works
- [x] Game pieces can still be manipulated
- [x] Game pieces can be configured
- [x] Configuration and spawning UI works 

> [!Note]
> Fields no longer spawn with a transform gizmo attached

image

AARD-1902

PR #1237, opened 2025-07-25
Analytics Testing & Exploration `[AARD-1937]`

rutmanz:
## Task

AARD-1937

Discover what we're doing with analytics and write unit tests

## Symptom

No unit tests, analytics kind of obscure and Google Analytics is a confusing platform

## Solution

I wrote tests to validate that the `AnalyticsSystem.ts` => `google-analytics.com` pipeline was working and all the intermediate components were present and working

I went into the google analytics page and poked around until I could find where the data was going, and discovered that the user attributes we had been sending were unusable because they had spaces (oops). 

I added typing to the analytics events (at least for keys) and added an event for selecting an input scheme.

In terms of processing analytics, the user attributes will give us more flexibility over what we filter, I found (kind of) how to see event info, and we can filter by developer or not.
image

## Verification

Tests pass, removing important parts of the code makes them not pass. Analytics are sent and can be seen live in the debug page of the GA dashboard (the main pages take a while to update)

---

- [ ] User-facing polish:


azaleacolburn (commented):
Seems like there are some security issues from some third-party app, is that something we care about.

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -277,6 +278,13 @@ class MirabufCachingService {
             }
         }
 
+        World.analyticsSystem?.event("Local Upload", {
+            assemblyName: displayName,
+            fileSize: buffer.byteLength,
+            key,
+            type: miraType == MiraType.ROBOT ? "robot" : "field",
```
azaleacolburn, line 291: Could you put a note in #1190 to fix this if this PR gets merged first.

rutmanz:
@azaleacolburn From an actual "is this a vulnerability" standpoint the answer is "no." The thing it is flagging is not a type of vulnerability that matters to us and that the library is used in the build process, not when serving. However, from a "we don't want Snyk errors" perspective, it might make sense to try and find a different way of accomplishing the git commit hash inclusion (@PepperLola @BrandonPacewic?)

BrandonPacewic (commented):
> @azaleacolburn From an actual "is this a vulnerability" standpoint the answer is "no." The thing it is flagging is not a type of vulnerability that matters to us and that the library is used in the build process, not when serving. However, from a "we don't want Snyk errors" perspective, it might make sense to try and find a different way of accomplishing the git commit hash inclusion (@PepperLola @BrandonPacewic?)

I'm honestly not 100% sure, if its not possible to mark this issue as ignored then we need to find some sort of workaround. At some point we should document all of our internal org checks as some of these things get flagged and sent to Pam.

rutmanz:
did a workaround so the dependency is no longer included

fission/src/test/analytics/Analytics.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        }, 20000)
+    })
+
+    describe("Without gtag Script", () => {
+        beforeEach(() => {
+            window.dataLayer = undefined
+            window.gtag = undefined
+            install()
+        })
+        test("AnalyticsSystem calls gtag appropriately", async () => {
```
AlexD717: This test doesn't look like it tests any actual behavior. Its calling a simple wrapper function, that then calls a function from an imported module.

AlexD717 (changes requested):

AlexD717 (approved):
lgtm

BrandonPacewic (commented):
Think this looks good, again going to have conflicts with #1241, doesn't look too bad. Will revisit merge order once we see how refactor is going.

PepperLola (approved):
There are a couple unrelated formatting changes but I don't really mind that, otherwise looks good

PR #1239, opened 2025-07-28
Unstick Button Input `[AARD-2018]`

AlexD717:
## Task
Add a new input that when pressed applies a slight upward force on the robot to get the robot unstuck when the robot is partially in the ground.

AARD-2018

## Symptom
When playing on certain fields, the robot can sometimes get partially stuck in the ground, making it impossible to move. This is especially common if there is a slight bump or ledge. The current way to fix this is to right click on the robot, drag it up slightly and then let go, but this is tedious and takes too long to be a viable solution in multiplayer.

## Solution
Add a new button that when pressed applies a slight upward force to the robot. This force isn't enough to give some sort of game advantage, but is strong enough to get the robot unstuck from the ground. A slider was also added in drivetrain configuration to allow the unstuck force to be customized for each robot.
Screenshot 2025-07-28 095710

## Verification
- Pressing the unstuck button causes a slight upward force to be applied
- - Robot rotation doesn't affect force direction
- If the robot gets stuck, spam clicking the unstuck button usually gets it unstuck
- Unstick force can be changed with a slider in configure assets

---

- [x] User-facing polish:


rutmanz:
I get an error when I try to go to Configure Joints > Drivetrain on this branch, with both cached and non cached dozer
image
Goes away when I clear preferences, but there should probably be a more graceful migration of old preferences to new preferences than crashing the page

AlexD717:
> I get an error when I try to go to Configure Joints > Drivetrain on this branch, with both cached and non cached dozer image Goes away when I clear preferences, but there should probably be a more graceful migration of old preferences to new preferences than crashing the page

I think this should have fixed it

rutmanz (approved):
lgtm.

I feel like unstick is kind of vague but I don't have any ideas on how to make it clearer without being too verbose

ryanzhangofficial (approved):
On Zach's point, perhaps labelling as "Get Unstuck" would be more clearer? Looks good though!

BrandonPacewic (commented):
This looks good to me, unfortunately its going to have a minor conflict with the UI refactor (#1241), not 100% sure what our plan is for this but just going to wait on it for now.

Dhruv-0-Arora (approved):
lgtm. An idea for another ticket would be to add a keybind (such as '/') that will show a list of commands (similar to discord's implementation)

BrandonPacewic (approved):

PR #1220, opened 2025-07-17
Websocket Prediction Spike

itspvty:
### Client-Side Prediction Spike: Fast Input/Visual with Server Correction

This PR contains multiple elements:

-   A test folder at `/client-prediction-spike` containing a client-side prediction system that meets the spike requirements as specified by the [ticket](https://jira.autodesk.com/browse/AARD-1962)
    - A quick setup guide
- An implementation guide for hooking the server into Fission/Jolt
- Findings regarding plausability and UX

## Test Stack

Included is a client-side prediction system that shows how to improve input responsiveness while keeping server authority. The client immediately applies movement using physics when inputs are received, while the server calculates the same positions and validates and corrects the predictions when they diverge from the server's decision. The client and server are essentially doing the same thing, but now the client's job is to make the game more responsive, and the server's job is to prove and validate the client action.

The spike **server** features:

-   All features from the Client-Server Spike
-   Input sequence tracking for catch-up rules
-   State history buffering
-   Divergence detection
-   Client-trust "Smooth" correction + Client-distrust "Snap" correction
-   Metrics system

The spike **client** features:

-   All features from the Client-Server Spike
-   Immediate input prediction and visualization
-   Visual debugging indicators (ghost robots, correction sphere)
-   Toggleable prediction

This spike **_does not_** feature:

-   Lag compensation (client-side rewind)
-   Rollback networking
-   Proactive cheating prevetion
-   Optimizations for low-bandwidth situations

### Prerequisites:

-   Node.js 18+
-   npm or yarn (more like yarnotcool)

### Installation

```bash
cd client-prediction-spike
npm install
```

### Run

```bash
npm run dev
```

_The server will start on port `3000` by default_

## Independent Tests

To verify the functionality, carry out the following tests:

### Prediction Responsiveness

-   **Test:** Press WASD keys to move the robot with prediction enabled
-   **Verify:** Movement is immediate (<8ms visual delay), _red_ robot moves instantly on keypress

### Correction Behavior

-   **Test:**
    -   Move robot in weird ways (button spam, etc) with prediction enabled
    -   Check corrections and ghost robot
-   **Verify:** Corrections should be small (< 10px average), smooth, and not too often

### Network-Distrust

-   **Test:**
    -   Simulate network delay (throttle connection through Devtools: Performance > Environment Settings)
    -   Test with 2-6 clients, preferably with multiple moving at the same time
-   **Verify:** Predictions will continue to work for the client, and other client's movements will be smooth even if delayed

## Measured Metrics

[...]

PR #1198, opened 2025-07-08
Websocket Client-Server Framework Spike

itspvty:
### Multiplayer Framework Spike: Websocket / Authoritative Server on Lazy Client Implementation

This PR contains multiple elements:

 - A test folder at `/multiplayer-spike` containing a barebones server and client that meet the spike requirements as specified by the [ticket](https://jira.autodesk.com/browse/AARD-1959)
	 - A quick setup guide
 - An implementation guide for hooking the server into Fission/Jolt
 - Findings regarding plausability and UX

## Test Stack

Included is a basic client-server setup that is meant to represent the framework that would be integrated into Synthesis. It pipes client inputs, simulates the world and physics on the server, and broadcasts the authoritative state to all clients on a fixed basis.

The spike **server** features:

 - A fixed game loop
 - Input processing
 - World simulation
 - State broadcasting
 - Performance monitoring
 
 The spike **client** features:
 
 - Real-time rendering
 - Input registry
 - State syncing

This spike does ***not*** feature:

 - Room/session seperation
 - Jolt physics
 - Input validation (*anticheat*)
 - Peripherals (score, timers, etc.)

### Online Test
An instance of the spike is live [here](http://bucket.test.ninjam.us:3010/) until July 15th.

---
Testing the spike locally is very simple.

### Prerequists:
- Node.js 18+
- npm or yarn (weird ahh)

### Installation
```bash
cd multiplayer-spike
npm install
```

### Run
```bash
npm run dev
```

*The server will start on port `3000` by default*

## Independent Tests

To verify the functionality, carry out the following tests:

### Multi-Client Stability
- **Test:** Open 2-6 browser tabs
- **Verify:** Each tab shows a red robot (*your* robot), and n-1 green robots (*other* robots). Robots update visually in sync.

### Input Latency
- **Test:** Press the WASD keys to move the robot
- **Verify:** Movement should feel responsive (<80ms perceived delay), and the ping internally (or server ping when on localhost) should stay below 50ms.

### Connection Health
- **Test:**
	- Kill server during activity
	- Restart/pause server
- **Verify:** Connection status indicates drops, and the client reconnects when the server becomes available.

## Measured Metrics

*All metrics observed on Windows*

Performance
 - **Flat Tick Rate:** 60Hz (16.6ms interval per frame)
 - **Medium Load Tick Rate (4 clients):** 59.95Hz (99.9%)
 - **High Load Tick Rate (10 clients):** 59.1Hz (98.5%)
 
Memory
[...]

PR #1240, opened 2025-07-28
WebRTC Technology and Peer to Peer Architecture Spike

azaleacolburn:
## Task

AARD-1960

The point of this spike was to determine the feasibility of the WebRTC protocol for peer-to-peer communication between clients in the Synthesis multiplayer mode.

This spike has been split up into three sections:

1. Code quality
2. Performance
3. Architecture

## Symptom

We need to decide on a technology to use for multiplayer communication.

### Code Quality

In my opinion, the [PeerJs](https://peerjs.com/) framework makes establishing connections and sending message between peers concise and easy to understand. It is a helpful and robust abstraction on the native WebRTC API.

PeerJs allows you to use this approximate logic for finding and messaging another peer

Setup custom PeerJs server on port 9000

```typescript
const peerServer = PeerServer({ port: PORT, path: "/", allow_discovery: true });
```

Connect to a PeerJs server

```typescript
const clientId = this.generateClientId();
const peer = new Peer(clientId, {
  host: "localhost",
  port: 9000,
  path: "/",
});
```

Get a list of all peers on the server and connect to one of them

```typescript
peer.listAllPeers((peers) => {
  const otherPeers: string[] = [];
  peers
    .filter((peerId) => peerId !== clientId)
    .forEach((peerId) => otherPeers.push(peerId as string));

  const firstPeer = otherPeers.pop();
  const connection = peer.connect(otherPeers);
});
```

Setup event listeners for peer

```typescript
connection.on("open", () => {
  console.log("Connection opened");
});

connection.on("data", (data: any) => {
  // Message handling logic here
});
```

### Performance

Most of this game was retrofitted from the [client prediction model spike](https://github.com/Autodesk/synthesis/pull/1220)

### Architecture

Due to the significant computational and thus financial cost of having an [authoritative server](https://github.com/Autodesk/synthesis/pull/1198) do the physics calculations and handle the I/O for all multiplayer games, with performance for the client being the same, a peer-to-peer solution would be preferable.

There have been two proposals for a peer-to-peer architecture that effectively synchronize the states of the two clients during disruptive events like collisions and knocking over game pieces, the Hybrid and the Authoritative Client Approach.

#### Authoritative Client Approach

In this architecture, one the hosting client essentially acts as the authoritative server, doing all the physics calculations and sending the entire physics state to the guest client every tick. This much like its server-based counterpart, this would be potentially extremely slow for the host, but it could be sped up for the guest with a [prediction model](https://github.com/Autodesk/synthesis/pull/1220).

#### Hybrid Approach

In this architecture, each client handles handles the physics calculations for their own robot which it sends to the other client each tick. The exception to this would be whenever it collided with either another robot or a game piece, in that case, it would send over its entire physics state to the other client(s), who would apply it.

The hosting robot would have priority in the case that each client sent a collision report to the other within a certain small interval of time, to avoid each client taking on the other's state instead of synchronizing.

Pictures of the original plans on whiteboard and shown below:

## Verification

1. `bun server.ts` to start the brokering server on port 9000
2. `bun client.html` to serve the client on port 3000
3. Open two tabs, go to `localhost:3000` on both
4. Move both robots around, observe that collisions are handled properly
5. (To be continued once this project is hosted on Vercel)

On Tuesday we'll have a discussion about how this compares to an authoritative server + client side prediction approach. I believe WebRTC with the hybrid approach is the correct choice because it keeps cost down for our servers, would be the fastest approach I've heard proposed, and have a simplier architecture than the server approach.

---

- [ ] User-facing polish:

```

```


rutmanz:
Just ran this with two devices on the local network and overlayed the videos as best as I could in like 5 minutes. ~~What is kind of odd to me is that the green ghost doesn't trail consistently, but rather one client is consistently behind the other (although thinking harder leads me to believe this could be an issue with my timing alignment)~~


[Media removed because I failed]

azaleacolburn:
> Just ran this with two devices on the local network and overlayed the videos as best as I could in like 5 minutes. What is kind of odd to me is that the green ghost doesn't trail consistently, but rather one client is consistently behind the other (although thinking harder leads me to believe this could be an issue with my timing alignment)

Hm ya that is weird, if you end up thinking that this lag is due latency with WebRTC or the hybrid implementation, we'll discuss it at the meeting. If you think this is the case, I'd be happy to see what optimization I can come up with for this implementation. 

I need to spend a few minutes spinning up a Vercel instance to test this properly though.

rutmanz:
New conclusion is that it looks like this, the small misalignment is probably the video editing and not the program itself given that it's still biased towards one client

Very nice (mostly)


https://github.com/user-attachments/assets/1a42fff3-a914-4f96-a347-3a04b8efeace

azaleacolburn:
> New conclusion is that it looks like this
> Very nice (mostly)

Ya not going to lie, that looks a lot better than I would expect. Thanks so much for the cool overlay videos!

Obviously the actual implementation will use Jolt and probably won't have the same hitbox issues.

BrandonPacewic (commented):
Going to put this into a spike branch and close this PR.

PR #1218, opened 2025-07-15
Download All Save Correct Names

ryanzhangofficial:
## Task
Make the naming for saved robots after "Download All" the same as individually downloading robots. 
[Jira Ticket](https://jira.autodesk.com/secure/RapidBoard.jspa?rapidView=13402&projectKey=AARD&view=detail&selectedIssue=AARD-1996#)

## Symptom
The newly downloaded robots from "Download All" have the full path as their name (`/api/mira/robots/...`) rather than just the robot/file name.

## Solution
Take in the mira name as an argument to `cacheRemote`

## Verification
- Names do not include `/api/mira/robots/` or `/api/mira/fields/` anymore.

PR #1241, opened 2025-07-28
UI Refactor `[AARD-1903]`

PepperLola:
## UI Refactor

AARD-1903

Refactored UI to improve the process of creating and using modals and panels. Rather than registering each possible UI element and then opening them with hardcoded IDs, the UI element itself is now passed into the relevant `open*` functions. In a `tsx` file, this looks something like `openPanel(<ConfigurePanel />)` but in regular `.ts` files it would look like `openPanel(React.createElement(ConfigurePanel))`.

<details>
    <summary>Panels vs Modals</summary>

There are two types of screens: Panels and Modals. I decided to generalize these as UI screens. There can only ever be one modal open at a time, and the modals demand attention from the user, meaning they block interaction to everything else behind them until they're dismissed. There can be any number of panels open and are draggable, so they're intended to be passive and able to remain open on the side while using Synthesis. For example, the main menu is a modal because we need the user to choose a mode before we can load a scene. The configuration panel, on the other hand, doesn't require the attention of the user so it can be made a panel. It's my personal opinion that any screen that doesn't *necessarily* require the attention of the user should be made a panel, as a modal can disrupt a user's focus.

The bulk of the UI management happens in [UIProvider](https://github.com/Autodesk/synthesis/blob/jwrigh/1903/ui-refactor-tmp/fission/src/ui/UIProvider.tsx). The screen type looks as follows:
```ts
interface UIScreen<T> {
    id: string
    parent?: UIScreen<unknown>
    content: ReactElement
    props: ModalProps | PanelProps
    onClose: UICallback<[CloseType], void>
    onCancel: UICallback<[], void>
    onAccept: UICallback<[T], void>
    onBeforeAccept: UICallback<[], T>
}
```

The `id`s are automatically generated when opening a screen and returned from the open function. They're required for closing a panel, which means you can only close a screen that you open (or if the code that opened the screen passes the `id` down elsewhere). This was a byproduct of the way I designed the opening (i.e. passing in a react component rather than an `id` so you don't have to register screens beforehand) but I think also makes sense as a requirement, as our previous code frequently closed panels in random places in a way that made the flow difficult to follow.

The `parent` is optional, but is a screen object referencing the screen that opened the current one, and comes from an optional argument in the open functions specifying the parent. This allows a screen to close its parent, as it has direct access to the `id`.

The `content` is just a `ReactElement` and is generally going to be a functional component that we define. This is what is rendered on the screen.

The `props` are either Modal- or Panel-specific properties. Each `Modal` and `Panel` interface that extends `UIScreen<T>` overwrite the `props` field with one of the relevant type. The properties interfaces are as follows:

```ts
interface UIScreenProps {
    configured: boolean
    title?: string
    hideCancel?: boolean
    hideAccept?: boolean
    cancelText?: string
    acceptText?: string
}

interface ModalProps extends UIScreenProps {
    // required for PanelProps to not satisfy ModalProps
    type: "modal"
    allowClickAway?: boolean
}

interface PanelProps extends UIScreenProps {
    type: "panel"
    position: PanelPosition
}
```

The `configured` property can be ignored, and is only used internally to denote whether the screen has been configured; the screen will only be rendered when once it has been configured. The `title` will render at the top of the screen. The `hideCancel` and `hideAccept` props will disable rendering of the relevant buttons when true, and `cancelText` and `acceptText` will change the text shown in the buttons. Each modal has a clickaway listener that will close it if the user clicks the background, but this behavior can be disabled by setting the `allowClickAway` property to `false`. As panels can exist alongside each other and can be moved, they also have configurable opening positions, which can be configured by setting the `position` property.

As the `ModalProps` interface only contains one optional field, there's nothing stopping a `PanelProps` instance from satisfying `ModalProps`, which is why the `type` fields were added. They can also be ignored and only serve the purpose of allowing the enforcement of the right properties type for each type of screen.
</details>

<details>
<summary>Callbacks</summary>

I created a `UICallback` object that can contain a default function and a user-defined function (or any combination of the two, as they are both optional). The user-defined function always runs first when defined, and the return value of the entire callback will always be that of the user-defined when it's not undefined. The user-defined functions are passed into the screen open functions, and the default functions are provided when calling `configureScreen` from within a screen element. Each screen has the following callbacks:

```ts
interface UIScreenCallbacks<T> {
    onClose?: (closeType: CloseType) => void
    onCancel?: () => void
    onBeforeAccept?: () => T
    onAccept?: (arg: T) => void
}
```

`onClose` is always called before any other callback when a screen is closed. The callback has a parameter of the `CloseType`, which can either be `Accept`, `Cancel`, or `Overwrite`. They are pretty self-explanatory - `Accept` and `Cancel` depend on the button that was pressed, and `Overwrite` is used when one screen takes the place of another. When a modal is opened, any previously open modal will also be closed and its `onClose` function will be called with the `Overwrite` close type. `onCancel` is called when the cancel button is clicked (or whenever a function is called to close a screen with the intent of canceling its purpose). `onBeforeAccept` is called, as the name suggests, before `onAccept`. It is only configurable from within a screen and can thus only contain a default implementation (passed in when calling `configureScreen`). It can return a value of type `T` (defined when creating a screen) that is then passed as an argument to `onAccept`, which can only contain a user-defined value configured in the open function for the screen.
</details>

<details>
    <summary>UIProvider and UIRenderer</summary>
    [UIProvider](https://github.com/Autodesk/synthesis/blob/jwrigh/1903/ui-refactor-tmp/fission/src/ui/UIProvider.tsx) exposes the current modal and panels, along with all the relevant ui-related functions:

```ts
type UIContextProps = {
    modal?: Modal<unknown>
    panels: Panel<unknown>[]
    openModal: OpenModalFn
    openPanel: OpenPanelFn
    closeModal: CloseModalFn
    closePanel: ClosePanelFn
    addToast: AddToastFn
    configureScreen: ConfigureScreenFn
}
```
These can be imported by calling the `useUIContext` hook from within a React component (e.g. within a panel or modal).

[UIRenderer](https://github.com/Autodesk/synthesis/blob/jwrigh/1903/ui-refactor-tmp/fission/src/ui/UIRenderer.tsx) just renders the current modal and panels, passing through the props and parent.
</details>

<details>
    <summary>Opening UI Screens</summary>

UI Screens can be opened using their relevant open functions, with the following type signatures:

```ts
type OpenModalFn = <T>(
    contents: ReactElement,
    parent?: UIScreen<T>,
    props?: Omit<ModalProps, "type" | "configured">
) => string
type OpenPanelFn = <T>(
    contents: ReactElement,
    parent?: UIScreen<T>,
    props?: Omit<PanelProps, "type" | "configured">
) => string
```
To open a screen, you need to call the relevant open function and pass in a `ReactElement`. You can optionally pass in a parent (which is where the parent property from before comes from) and properties you want to configure (which is where the user-defined callbacks come from). The functions will return the randomly-generated `id` for the new screen.
</details>

<details>
    <summary>Closing UI Screens</summary>

To close a screen, you need to first have the `id` of the screen you want to close, which will have previously been returned from the function that opened that screen. To close the screen, you simply call the relevant close function. As there can only ever be one modal open at once, the `closeModal` function only requires a close type to be passed in, which was defined above to be one of `Accept`, `Cancel`, or `Overwrite`. The `closePanel` function requires the `id` of the panel to close along with the close type.
</details>

<details>
    <summary>Rendering Toasts</summary>

I installed the `notistack` package to add a notification stack to the bottom-left corner of the screen like we had previously. UIProvider exposes an `addToast` function that takes in a type (from `notistack`) along with any number of string arguments beyond that, which will be joined by line breaks and rendered in the notification.
</details>

<details>
    <summary>Global Functions</summary>

The [GlobalUIControls](https://github.com/Autodesk/synthesis/blob/jwrigh/1903/ui-refactor-tmp/fission/src/ui/components/GlobalUIControls.ts) file exports `globalAddToast`, `globalOpenPanel` and `globalOpenModal` which are all set to the respective functions from UIProvider. This is so that non-React-component code can perform any of these actions. This is especially important for adding toasts, as there are many places we would want to provide feedback to the user (e.g. in the event of an error) from code that isn't in a React component.
</details>

<details>
    <summary>Creating a UI Screen</summary>

Creating a new screen is as simple as creating a new file with a functional component in the correct directory (i.e. `src/ui/modals/` for modals, `src/ui/panels/` for panels) and exporting it. The type of the component must be either `PanelImplProps<T>` or `ModalImplProps<T>`, where `T` is the type of the data to be returned from `onBeforeAccept` and passed into `onAccept`.

Those types allow the screens to access their object (`Modal` or `Panel`, the interface introduced earlier with the `id`, `parent`, etc.) along with that of their parent as props.

Each screen must also be configured, which must be done via the `configureScreen` function exposed by the UI context within UIProvider. **This is necessary for all panels**, even if you don't plan to override any properties or callbacks. If this isn't done, the UI renderer won't consider the screen to have been configured and as such will not render it.

    The type signature for the configure function is a bit difficult to understand at first:
```ts
type ConfigureScreenFn = <T extends UIScreen<any>>(
    screen: T,
    props: T extends Panel<infer _> ? Partial<Omit<PanelProps, "configured">> : Partial<Omit<ModalProps, "configured">>,
    callbacks: T extends Modal<infer S>
        ? Omit<Partial<UIScreenCallbacks<S>>, "onAccept">
        : T extends Panel<infer S>
          ? Omit<Partial<UIScreenCallbacks<S>>, "onAccept">
          : never
) => void
```

It essentially takes either a panel or a modal as its first argument and, depending on which is provided, will require the props for that type to be passed as the second argument (guarding against configuring a modal with panel props). The callbacks provided here to the `callbacks` argument are then set as the default implementation for the callbacks (as opposed to the user-defined implementation; discussed above).

The APS management modal is a good, small example of creating a modal:
```ts
import { Stack } from "@mui/material"
import type React from "react"
import { useEffect, useState } from "react"
import { HiUser } from "react-icons/hi"
import APS from "@/aps/APS"
import type { ModalImplProps } from "@/ui/components/Modal"
import { useUIContext } from "../helpers/UIProviderHelpers"

const APSManagementModal: React.FC<ModalImplProps<void>> = ({ modal }) => {
    const { configureScreen } = useUIContext()
    const [userInfo, _] = useState(APS.userInfo)
    useEffect(() => {
        const onBeforeAccept = () => {
            APS.logout()
        }

        configureScreen(modal!, { title: userInfo?.name ?? "Not signed in", acceptText: "Logout" }, { onBeforeAccept })
    }, [modal, userInfo?.name])

    return (
        <Stack spacing={10} direction="row">
            {userInfo?.picture ? (
                <img alt={userInfo?.name} src={userInfo?.picture} className="h-10 rounded-full" />
            ) : (
                <HiUser />
            )}
        </Stack>
    )
}

export default APSManagementModal
```

We can see that it imported the `configureScreen` function using the `useUIContext` hook. This is also how to import the `openModal`, `closeModal`, `openPanel`, and `closePanel` functions.

It also called `configureScreen` within a `useEffect`, which is the preferred way of configuring a screen. Even if no configuration were necessary, it would still need to call `configureScreen` with empty property objects, as follows:

```ts
configureScreen(modal!, {}, {})
```

This modal could then be opened in another component like so:
```ts
// first import the openModal function
const { openModal } = useUIContext()

// then open the modal
openModal(<APSManagementModal />)
```

or from a regular TypeScript file, where you can't create a JSX/TSX element:
```ts
// first import React and the globalOpenModal function
import React from "react"
import { globalOpenModal } from "@/ui/components/GlobalUIControls"

// then open the modal
globalOpenModal(React.createElement(APSManagementModal))
```
</details>

<details>
    <summary>MUI</summary>

I migrated (almost?) everything to use [MUI Material UI](https://mui.com/material-ui/) instead of our own built-in components. Some more complex components were kept but internally use MUI components (e.g. SelectMenu), but some were discarded entirely (e.g. removed Dropdown in favor of MUI's Select).

The most important components going forward are `Stack` (flex box), `Box` (general container), `Select` (dropdown menu), `Button` (button), and `TextField` (text input field). A full list of MUI components can be found [here](https://mui.com/material-ui/all-components/).
</details>

---

- [ ] User-facing polish:


Dhruv-0-Arora (changes requested):
1. No way to close the settings modal unless you go to the theme editor and close using the theme editor.

2. When the InputSchemeSelector is visible and you click "Enter", it does not select Ernie as intended and instead throws an error. Same happens if you don't select anything and just click "Finish"
Screenshot 2025-07-29 at 4 27 33 PM

3. Make the Blue team's color when choosing input scheme not the secondary color and instead just a dark blue.

4. Fields disappear after clicking "Finish"

fission/src/events/ConfigurationSavedEvent.ts
```diff
@@ -6,11 +6,11 @@ export class ConfigurationSavedEvent extends Event {
         window.dispatchEvent(this)
     }
 
-    public static listen(func: (e: Event) => void) {
+    public static Listen(func: (e: Event) => void) {
```
rutmanz: bad

rutmanz:
I think onBeforeAccept and onAccept should be switched so the panel can exfiltrate date to its caller

AlexD717 (changes requested):
1) Can select text by dragging with your mouse in text boxes (the panel moves).
2) In Configure Assets tab, clicking on the same tab already selected changes the category
3) Configure Assets input tab, doesn't show all input scheme, instead shows one input scheme to configure
4) Many panels are to high to fit on one screen
5) No scoreboard for match mode
6) Initially spawning in a robot, their name tag doesn't match the alliance (its grey)

fission/src/systems/match_mode/MatchMode.ts
```diff
@@ -110,16 +106,21 @@ class MatchMode {
         this._endgame = true
     }
 
-    start(openModal: (modalName: string) => void) {
+    start(openModal: OpenModalFn) {
```
rutmanz: what if you used the global function here

PepperLola:
> 1. No way to close the settings modal unless you go to the theme editor and close using the theme editor.
> 
>     2. When the InputSchemeSelector is visible and you click "Enter", it does not select Ernie as intended and instead throws an error. Same happens if you don't select anything and just click "Finish"
> 
> 
> Screenshot 2025-07-29 at 4 27 33 PM
> 
>     3. Make the Blue team's color when choosing input scheme not the secondary color and instead just a dark blue.
> 
>     4. Fields disappear after clicking "Finish"

1. Fixed - non-overwritten default properties were being discarded when opening modals and panels
2. This error was related to 4, which was also fixed. I'm not sure how you're having behavior when pressing enter because that doesn't do anything for me
3. As discussed earlier, those buttons were a theme color just to signify that they weren't active, they weren't associated with the blue team. We might want to go in later and adjust the red and blue alliance colors to not be #ff0000 and #0000ff but for now I think it's fine
4. Fixed, the initial config panel was trying to close itself in a callback after the UI provider was closing it

PepperLola:
> I think onBeforeAccept and onAccept should be switched so the panel can exfiltrate date to its caller

I switched them and updated the PR description, let me know if it looks good and makes sense.

PepperLola:
> 1. Can select text by dragging with your mouse in text boxes (the panel moves).
> 
>     2. In Configure Assets tab, clicking on the same tab already selected changes the category
> 
>     3. Configure Assets input tab, doesn't show all input scheme, instead shows one input scheme to configure
> 
>     4. Many panels are to high to fit on one screen
> 
>     5. No scoreboard for match mode
> 
>     6. Initially spawning in a robot, their name tag doesn't match the alliance (its grey)

1. Fixed by adding "input" to selectors that aren't draggable
2. Fixed by not allowing null selections within the toggle button group
3. I don't have this issue. I'm not sure if I changed something that inadvertently fixed it but I see all the input schemes
4. I'm not sure this is within the scope of this PR - a different ticket is to improve the appearance of the UI which I think this makes more sense for; this PR was just to change the way we handle the UI and reimplement all the screens
5. Done
6. Fixed, the extra configuration in the initial config panel didn't make it through the merge :(

fission/src/systems/simulation/SimConfigShared.ts
```diff
@@ -258,14 +253,14 @@ export class SimConfig {
         return config
     }
 
-    public static refreshRobotIO(config: SimConfigData) {
-        SimConfig.addRobotIONode(config)
+    public static RefreshRobotIO(config: SimConfigData) {
```
rutmanz: uh oh

itspvty (dismissed):
Could we conform this PR's description to the project's standard?

fission/src/ui/components/simulation/FlowControls.tsx
```diff
@@ -1,21 +1,15 @@
...
-        <button
-            className="border-[0.0625rem] border-interactive-element-solid bg-background p-1 w-8 h-8 flex flex-col justify-center items-center"
-            onClick={() => onClick?.()}
-        >
-            {children}
-        </button>
-    )
+function FlowControlButton({ onClick, children }: PropsWithChildren<{ onClick?: () => void }>) {
+    return <Button onClick={() => onClick?.()}>{children}</Button>
 }
```
rutmanz: isn't this just actuall a Button?

fission/src/Synthesis.tsx
```diff
@@ -107,40 +44,30 @@ const Synthesis: React.FC = () => {
...
+    // useEffect(() => {
+    // TODO:
+    // const scoreboardExists = false
+    // panelElements.forEach(x => {
+    //     if (x.key == "scoreboard") scoreboardExists = true
+    // })
+    // if (PreferencesSystem.getGlobalPreference("RenderScoreboard") && !scoreboardExists) {
+    //     openPanel("scoreboard")
+    // }
+    // });
```
rutmanz: why no scoreboard?

AlexD717:
To me, it looks like the UI Refactor has gotten to a point, where multiple people can now work on it, and improve it. There are a lot of bugs that have to be fixed before this can be put into production, but I am fine with this being in dev, and a bunch of Jira tickets being made to fix all the issues.

Here are some more issues that I found (I think its fine to merge this first and fix these later)
1) Configure Assets input tab, doesn't show all input scheme, instead shows one input scheme to configure. - This only happens if you have at least one robot spawned
2) Right clicking on a robot and pressing move opens the configure assets panel, but doesn't place you in the correct spot.
3) End of match scoreboard UI is broken, but AARD-1896 already exists
4) Configure ejector tab has a unlabeled button (reset)
5) There appears to be no way to delete custom input schemes - in the assembly setup panel
6) There is no way to cancel creating a input scheme, on the model where you input the input scheme name
7) Drag mode arrow key rotation is extremely slow
8) Starting drag mode previously spawned a button in the bottom left that could be clicked to disable drag mode
9) You can select and highlight the title on the panels/modals
10) The checkbox/toggle-switch for persistent points in scoring zones doesn't change graphically but leaving the menu, and going back in updates it
11) The UI for scoring/protected zone config looks ugly
Screenshot 2025-07-30 at 9 07 07 AM

12) Some dropdowns, when pressed remove part of the border
Screenshot 2025-07-30 at 9 07 21 AM

Personal Preference
- I think its very annoying how the spawn assets and configure assets spawn in the middle of the screen, since the robot you are spawning/configuring is usually in the middle of the screen, and so you can't see it.
- Instead of allowing the user to spawn multiple panels, if one is already open I think we should just bring that one up

azaleacolburn (dismissed):
Clicking an action item in the control panel for an object (E.g., `Move`) just opens the general Configure Assets Panel.

fission/src/ui/components/Modal.tsx
```diff
@@ -1,158 +1,68 @@
...
-                                {...SoundPlayer.buttonSoundEffects()}
-                                className={`${
-                                    acceptBlocked ? "bg-interactive-background" : "bg-accept-button"
-                                } rounded-md cursor-pointer px-4 py-1 font-bold duration-100 hover:brightness-90
-                                transform transition-transform hover:scale-[1.03] active:scale-[1.06]`}
-                                style={{ fontWeight: "bold" }}
-                            />
-                        )}
-                    </div>
+                    </CardActions>
```
rutmanz: I think `=== false` would make a lot more sense for these

fission/src/ui/components/Scoreboard.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    <Stack direction="row" className="w-full justify-center">
+                        <Label size="lg">{time}</Label>
+                    </Stack>
+                )}
+                <Stack direction="row" className={`px-4 ${showTime() ? "pt-1 pb-4" : "py-4"}`} gap={1}>
+                    {/* TODO: change colors? */}
+                    <Stack
+                        direction="column"
+                        className="items-center justify-center w-20 h-20 rounded-lg"
+                        sx={{ background: "#ff0000" }}
```
rutmanz: I feel like alliance red and blue should be part of the theme (in the MUI sense) or at least not defined everywhere they're referenced (and not these colors please my eyes)

fission/src/ui/components/Skybox.tsx
```diff
@@ -1,10 +1,10 @@
 import World from "@/systems/World"
-import { useTheme } from "@/ui/helpers/UseThemeHelpers"
+// import { useTheme } from "@/ui/ThemeContext"
 
 const Skybox = () => {
-    const { currentTheme, themes } = useTheme()
+    // const { currentTheme, themes } = useTheme()
     if (World.sceneRenderer) {
-        World.sceneRenderer.updateSkyboxColors(themes[currentTheme])
+        // World.SceneRenderer.UpdateSkyboxColors(themes[currentTheme])
```
rutmanz: ?

fission/src/ui/components/StatefulSlider.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const StatefulSlider: React.FC<
+    Omit<Parameters<typeof Slider>[0], "value" | "onChange"> & {
+        label: string
+        defaultValue: number
+        onChange: (val: number) => void
+        tooltip?: string
+    }
+> = props => {
+    const [value, setValue] = useState(props.defaultValue)
```
rutmanz, line 12: This is the opposite of what StatefulCheckbox does (having an internal state variable). I called StatefulCheckbox that in contrast to the established Checkbox which held the state internally

fission/src/ui/helpers/StateProviderHelpers.ts
```diff
@@ -0,0 +1,40 @@
...
+    unconfirmedImport: boolean
+    setUnconfirmedImport: (_state: boolean) => void
+    // ConfigureInputs stuff
+    selectedScheme?: InputScheme
+    setSelectedScheme: (_scheme: InputScheme) => void
+    // Configure Panel
+    configurePanelSettings?: ConfigurePanelSettings
+    setConfigurePanelSettings: (_settings?: ConfigurePanelSettings) => void
+    configurationType: ConfigurationType
+    setConfigurationType: (_type: ConfigurationType) => void
```
rutmanz: I kind of thought the point of being able to pass things back and forth between panels and the things that open/close them was to avoid this global state stuff

BrandonPacewic (commented):
Two things:

1) Are we planning on pushing this before we make any more changes to the UI that would cause conflicts?
2) If we are, what's the plan for getting everything into a "polished" state before the end of the summer?

If we choose to wait then we are going to be doing two things of development, on dev and copying the changes over to this branch as development continues.

If we choose to merge this now then it looks like we have a lot of changes to make before we can release it and if so then how is that going to affect the rest of the stuff we have planned for the end of summer.

rutmanz:
These might not all go in this PR but things that I noticed
- Drag mode indicator is gone
- Toasts and graphics settings show up on the left now which overlaps more with main hud
- Can't dismiss toasts
- 
<details>
<summary>Tall things don't scroll</summary>

image

</details>

<details>
<summary>Analytics consent has no background</summary>

image


</details>

<details>
<summary>Scoreboard doesn't respect preference for whether or not to show it and also the colors are blinding me</summary>

image


</details>

<details>
<summary>Can't upload files (no continue button)</summary>

image


</details>

<details>
<summary>Random deprication warning</summary>

image


</details>

<details>
<summary>Why is it so wide?</summary>

image


</details>

<details>
<summary>Dropdown multiselect isn't super clear and also doesn't move with the panel and also has a random missing border section</summary>

image


</details>

<details>
<summary>Nametag text is dark</summary>

image


</details>

rutmanz (changes requested):

AlexD717:
> Nametag text is dark
> image

It looks like the UI changes based on if the user has dark mode or light mode. Is that something that we want to support?

fission/src/ui/components/StatefulSlider.tsx
PepperLola: I renamed StatefulCheckbox to Checkbox since it doesn't hold state internally. I forget if that was what we settled on or if I got it backwards though.

PepperLola:
> 1. Configure Assets input tab, doesn't show all input scheme, instead shows one input scheme to configure. - This only happens if you have at least one robot spawned
> 2. Right clicking on a robot and pressing move opens the configure assets panel, but doesn't place you in the correct spot.
> 3. End of match scoreboard UI is broken, but [AARD-1896](https://jira.autodesk.com/browse/AARD-1896) already exists
> 4. Configure ejector tab has a unlabeled button (reset)
> 5. There appears to be no way to delete custom input schemes - in the assembly setup panel
> 6. There is no way to cancel creating a input scheme, on the model where you input the input scheme name
> 7. Drag mode arrow key rotation is extremely slow
> 8. Starting drag mode previously spawned a button in the bottom left that could be clicked to disable drag mode
> 9. You can select and highlight the title on the panels/modals
> 10. The checkbox/toggle-switch for persistent points in scoring zones doesn't change graphically but leaving the menu, and going back in updates it
> 11. The UI for scoring/protected zone config looks ugly

1. This is an issue in dev and I don't think it's within the scope of this PR
2. In progress
3. I was expecting that task to take care of this
4. Fixed
5. Also an issue in dev - I created a Jira ticket to fix
6. Also an issue in dev - Jira ticket created
7. Also an issue in dev
8. Fixed
9. Fixed
10. Fixed
11. Not within the scope of this PR - another ticket to fix the UI appearance already exists

PepperLola:
> * Drag mode indicator is gone
> * Toasts and graphics settings show up on the left now which overlaps more with main hud
> * Can't dismiss toasts

1. Drag mode indicator is back!!!!
2. Toasts now show up on the right, default panel spawn location is the right and the graphics panel opens in the middle
3. Can now dismiss toasts

> Tall things don't scroll
> Analytics consent has no background
> Scoreboard doesn't respect preference for whether or not to show it and also the colors are blinding me
> Can't upload files (no continue button)
> Random deprication warning
> Why is it so wide?
> Dropdown multiselect isn't super clear and also doesn't move with the panel and also has a random missing border section
> Nametag text is dark

1. I think this is not within the scope of this PR, but rather the one to fix the UI that will happen next
2. Fixed
3. Fixed
4. Fixed
5. I'm not sure why that's showing up on line 18 of Scoreboard.tsx since that just calls `useState`. However, a similar error appears for panels and originates from the react-draggable library.
6. Fixed (made it tall instead)
7. Fixed labels so there's text in the gap, added checkboxes to multi-select dropdown
8. Made nametag text always light

AlexD717 (changes requested):
Just a couple more issues that I found

1) If go into the configure score zone tab in configure assets, and then press the robot tab the app crashes

https://github.com/user-attachments/assets/889e64bf-bcf4-442b-849b-5dc2fac2326f

2) The scoring zone in configure assets panel has two buttons that both navigate back
Screenshot 2025-07-31 at 8 28 50 AM

AlexD717:
Also:

3) In the intake configure tab, show intake zone indicator always, doesn't graphically update
4) Is there a way to have the slider show what value they are on, because this isn't very helpful
Screenshot 2025-07-31 at 9 10 09 AM

ryanzhangofficial:
Some suggestions. These might part of 2002 though.
1. The buttons in debug tools, developer tool, and match mode config seem jammed together and uncomfortable to look at it. Perhaps make the borders the bright blue and the body the same color as the background?
2. Perhaps make the saved robot text the same size as the 6 default robots so it's easy to distinguish between section title (Saved Robot) and the actual saved robots.
3. I agree with Alexey on his point on having slider show what value they are on. Other examples are the ejector velocity,  I think this should be standardized throughout fission.

Bugs:
1. The issue resolved in [AARD-1996](https://github.com/Autodesk/synthesis/pull/1218) have resurfaced.
2. Receiving the error below when you add click a field -> click on scoring zones or protected zones -> click on Robot tab.
Screenshot 2025-07-31 094405

ryanzhangofficial (changes requested):
Some suggestions:
1. The buttons in debug tools, developer tool, and match mode config seem jammed together and uncomfortable to look at it. Perhaps make the borders the bright blue and the body the same color as the background?
2. Perhaps make the saved robot text the same size as the 6 default robots so it's easy to distinguish between section title (Saved Robot) and the actual saved robots.
3. I agree with Alexey on his point on having slider show what the actual value is. Another example is the ejector velocity which had a number before as well. I think this should be standardized throughout fission.

Some bugs:
1. The issue resolved in [AARD-1996](https://github.com/Autodesk/synthesis/pull/1218) has resurfaced.
2. Receiving the errors pictured below when I click on a field -> click scoring or protected zones -> click on robot. Fission crashes.

Image

PepperLola:
> Just a couple more issues that I found
> 
> 1. If go into the configure score zone tab in configure assets, and then press the robot tab the app crashes
> 
> 2. The scoring zone in configure assets panel has two buttons that both navigate back

Both fixed

PepperLola:
> Also:
> 
> 3. In the intake configure tab, show intake zone indicator always, doesn't graphically update
> 4. Is there a way to have the slider show what value they are on, because this isn't very helpful

3. Fixed
4. I don't think that's within the scope of this PR, can be done in the UI appearance improvement one. I agree with the change though

fission/src/ui/helpers/StateProviderHelpers.ts
PepperLola: The configure panel settings were only used in one place so I was able to make those a state just in that component. Otherwise, these are all pretty inherently global state– they're configured and used in screens that don't open each other and are settings that need to persist beyond the lifetime of any one panel. I can't think of a better way to manage it than a context like this, and I think it's a lot better than our solution before of having a shared object between multiple files.

PepperLola:
> Some suggestions:
> 
> 1. The buttons in debug tools, developer tool, and match mode config seem jammed together and uncomfortable to look at it. Perhaps make the borders the bright blue and the body the same color as the background?
> 2. Perhaps make the saved robot text the same size as the 6 default robots so it's easy to distinguish between section title (Saved Robot) and the actual saved robots.
> 3. I agree with Alexey on his point on having slider show what the actual value is. Another example is the ejector velocity which had a number before as well. I think this should be standardized throughout fission.

I don't think these fall within the scope of this PR– I definitely agree with the feedback but I think that will be covered by the UI improvement PR later.

> 
> Some bugs:
> 
> 1. The issue resolved in [AARD-1996](https://github.com/Autodesk/synthesis/pull/1218) has resurfaced.
> 2. Receiving the errors pictured below when I click on a field -> click scoring or protected zones -> click on robot. Fission crashes.

Both have been fixed

AlexD717 (changes requested):
If you have a robot and a field spawned in, and you right click on the field and press configure the app crashes.

Dhruv-0-Arora (changes requested):
Graphics settings panel should be more descriptive. I think its better to just make this another PR though and get the UI Refactor merged.

Screenshot 2025-08-04 at 4 03 19 PM

This is what graphics settings looks like on dev

Screenshot 2025-08-04 at 4 07 07 PM


[I realized that its not to do with th  graphics settings panel but how the sliders are formatted. The value of the slider is not displayed which, in my opinion, would be useful be to add]

Dhruv-0-Arora:
Potential feature to be added in the future: 

Save the positioning of panels in preferences so that they always open in the position set by the user.

rutmanz (approved):
I think all of the core functionality seems good, and it's just style things remaining

ryanzhangofficial (approved):
If the criteria for merging is no crashing and retaining functionality, I think we're good to go.

PepperLola:
> Graphics settings panel should be more descriptive. I think its better to just make this another PR though and get the UI Refactor merged.
> 
> 
> 
> Screenshot 2025-08-04 at 4 03 19 PM
> 
> 
> 
> This is what graphics settings looks like on dev
> 
> 
> 
> Screenshot 2025-08-04 at 4 07 07 PM
> 
> 
> 
> 
> 
> [I realized that its not to do with th  graphics settings panel but how the sliders are formatted. The value of the slider is not displayed which, in my opinion, would be useful be to add]
> 
> 

These are style changes that are not within the scope of this PR

Dhruv-0-Arora (approved):
This is a work of ART

AlexD717 (approved):
Lets get this in!

As a different ticket, we should consider removing a lot of the console.log statements, as it currently clutters up the console.

BrandonPacewic (approved):
@PepperLola I'll let you be the final review and merge for this PR. From what we discussed yesterday in standup everyone is ready to get this in and begin working on it. Plus we can finally merge everything else from the backlog.

PR #1244, opened 2025-07-29
Electron Packaging `[AARD-1786]`

AlexD717:
## Task
Testing out packaging the project with electron to see how performant it is and whether we could use it as another distribution method.

AARD-1786

## Symptom
Currently Synthesis can only be played in the browser, and can't be played offline (mostly). One possible solution would be to package the app with electron and distribute the app.

## Solution
Packaged the app with electron and tested the performance.

## Testing
To test the app in development run the command
`bun run start` (fallbacks to npm)
To build the app run
`bun run publish` (fallbacks to npm)

The app size on MacOS is 1.35 GB, but on Windows I think it was 1 GB. This is around was is expected for electron apps.

Performance wise I was getting around 5 frames less on electron compared to the web (when moving Dozer around on the 2018 field). Technically speaking I think its possible to use a different more optimized version of Jolt, and not Jolt WASM which should give electron higher performance, but this appears to be a very difficult task.

## Verification
- Web app still works, but can build an electron app that also works.
  - Note: For some reason the published version of the electron app displays the main menu background as green. However the same behavior happens in the production version of [synthesis](https://synthesis.autodesk.com/fission/) so this isn't an electron issue.
  - Tested on
    - Windows
    - MacOS
    - Fedora
 
Should also work on other Linux distros, but they haven't been tested yet.

---

- [x] User-facing polish:


fission/src/ui/panels/mirabuf/ImportMirabufPanel.tsx
```diff
@@ -177,14 +177,18 @@ const ImportMirabufPanel: React.FC<PanelPropsImpl> = ({ panelId }) => {
     useEffect(() => {
         // To remove the prettier warning
         const x = async () => {
-            fetch(`/api/mira/manifest.json`)
+            // Detect if we're running in electron and use direct remote URL
+            const isElectron = window.electronAPI != null
+            const baseUrl = isElectron ? "https://synthesis.autodesk.com" : ""
```
rutmanz, line 204: Does electron not support proxying like vite does?

AlexD717: It doesn't support it like vite. There are different ways to get around that, like intercepting requests, but this is the easiest solution that I found. In the production build before this the app wasn't getting any field/robot files.

fission/package.json
```diff
@@ -1,25 +1,32 @@
...
     "fmt:fix": "bunx --bun biome format src --write",
     "style": "bun run fmt && bun run lint",
     "style:fix": "bun run fmt:fix && bun run lint:fix",
     "assetpack": "git lfs pull && tar -xf public/assetpack.zip -C public/",
-    "playwright:install": "bun x playwright install"
+    "playwright:install": "bun x playwright install",
+    "start": "electron-forge start",
+    "package": "electron-forge package",
+    "make": "electron-forge make",
+    "publish": "electron-forge publish"
```
rutmanz:
```suggestion
    "electron:start": "electron-forge start",
    "electron:package": "electron-forge package",
    "electron:make": "electron-forge make",
    "electron:publish": "electron-forge publish"
```

I think it makes sense to prefix these all with electron just for clarity

rutmanz (changes requested):
other than this, lgtm

rutmanz (approved):

ryanzhangofficial (commented):
Could you document this addition in one of the markdown tutorial files, or maybe in the main one? I think it’s a noteworthy capability and would be useful for users as a reference.

ryanzhangofficial (approved):

PepperLola (approved):
Worked well, it had performance issues when I was running the packaged app but that's probably just to do with electron

Dhruv-0-Arora (approved):

BrandonPacewic (approved):
Performance is not the best but thats exactly what I expected. Nice showcase of electron.

PR #1247, opened 2025-07-31
Robot Falling of the Edge of the World `[AARD-2034]`

AlexD717:
## Task
Currently robots can drive off the edge of the map and they will keep on following forever, which isn't something that we really want, as the only way to fix this is to reload the entire page.

AARD-2034

## Symptom
You can drive a robot of the edge of the map, and then you are forced to reload the page, clearing everything that you have previously spawned.

## Solution
Add a tracker that detects if a robots position has gone below the map boundary (constant), and in that case teleports it to its spawn position.

## Verification
Driving of the edge of the map causes you to teleport back to the center

---

- [x] User-facing polish:


ryanzhangofficial (approved):
Love it!

fission/src/systems/match_mode/RobotPositionTracker.ts
```diff
@@ -0,0 +1,64 @@
...
+            const rootBody = World.physicsSystem.getBody(rootNodeId)
+            const rootTransform = convertJoltMat44ToThreeMatrix4(rootBody.GetWorldTransform())
+
+            const rootPosition = new THREE.Vector3()
+            const rootRotation = new THREE.Quaternion()
+            const rootScale = new THREE.Vector3()
+            rootTransform.decompose(rootPosition, rootRotation, rootScale)
+
+            if (rootPosition.y < this._mapBoundaryY) {
+                console.log(`Robot fell off the map: ${robot.assemblyName}`)
```
rutmanz:
```suggestion
```

fission/src/systems/match_mode/RobotPositionTracker.ts
```diff
@@ -0,0 +1,64 @@
...
+                SimulationSystem.robotPenalty(robot, this._offMapPenalty, "Robot fell off the map")
+
+                // TODO: Once driver station is implemented, we should reset the robot to the driver station position
+                const resetPosition = new JOLT.RVec3(0, 0.2, 0)
+                const resetRotation = JOLT.Quat.prototype.sIdentity()
+                const zeroVelocity = new JOLT.Vec3(0, 0, 0)
```
rutmanz: Yeah my concern about resetting to 0,0 is that for fields like 2018 you just get stuck in something. Resetting to driver station positions would solve that once they're implemented, created `AARD-2040` so we don't forget about it

rutmanz (approved):

fission/src/systems/match_mode/RobotPositionTracker.ts
AlexD717: I like how AARD-2041 blocks AARD-2040

azaleacolburn (approved):
lgtm

BrandonPacewic (commented):
Is there a reason that this was added in the match mode directory? Does it not make sense outside of it?

BrandonPacewic (approved):

PR #1248, opened 2025-07-31
Extension Penalty `[AARD-2011]`

AlexD717:
## Task
Create penalties for overextending beyond the allowed amount.

AARD-2011

## Symptom
FRC has rules for extending beyond frame perimeter, but Synthesis has no penalties to support this. 

## Solution
The way it currently works is the starting size is recorded on match start, and then if the robot extends beyond the starting size + allowed extension a penalty is applied. This does mean that if the robot doesn't start in frame perimeter this will not work properly. However,
  - FRC rules require you to start within frame perimeter
  - There is no way to detect where the frame perimeter actually is, without modifying the exporter and having an extra field to select the bumpers.
  - Many robots are below maximum allowed size, so assuming maximum size for all robots, wouldn't really work.

## Verification
Tested with 2471 robot and moving the arm, with a custom match mode ruleset with an allowed extension of 0, and made sure that a penalty was still applied.

---

- [x] User-facing polish:


rutmanz (approved):
lgtm

ryanzhangofficial (approved):

BrandonPacewic (approved):

PR #1249, opened 2025-08-01
Scale Drag Mode Torque `[AARD-2035]`

rutmanz:
## Task

AARD-2035

## Symptom

Currently using the arrows in drag mode is a reasonable speed for some gamepieces and very slow for robots, very fast for some other gamepieces

## Solution

The torque is now proportional to the length of the diagonal inertia vector. This means it's not accounting for different moments in different axes, but it's close enough that everything goes at a reasonably similar speed and it's neither too fast nor too slow

## Verification

- Spin robots
- Spin 2018 Power Cube
- Spin 2023 Cone
- Spin 2023 Cube
---

- [x] User-facing polish:


AlexD717 (approved):
lgtm

As a different ticket I think it would be beneficial to add a slider into settings to be able to configure the rotation speed, and other properties of the drag mode system. Personally, I find this current rotation speed to fast and imprecise, but other people might like this speed.

ryanzhangofficial (approved):
looks great!

BrandonPacewic (approved):
:+1:

PR #1251, opened 2025-08-01
Refactor Match Mode Config `[AARD-2043]`

rutmanz:
To be merged after #1248 

## Task

AARD-2043

## Symptom

Currently the match mode config system is a little bulky and adding new properties requires more repetition than I think is necessary (I noticed this reviewing #1248)

## Solution

Refactored to avoid requiring a bunch of extra constants and explicit type checking

## Verification

Confirm that uploading an invalid config still shows errors, uploading a valid config doesn't, and the config values are what you expect

---

- [x] User-facing polish:


AlexD717 (changes requested):
This is no longer converting user input from feet to meters, and expect the input to be in meters.

rutmanz:
> This is no longer converting user input from feet to meters, and expect the input to be in meters.

@AlexD717 This is intentional, as everything else (the exporter) standardizes on metric units and it doesn't make sense imo to have this one place arbitrarily be in feet.

AlexD717:
> > This is no longer converting user input from feet to meters, and expect the input to be in meters.
> 
> @AlexD717 This is intentional, as everything else (the exporter) standardizes on metric units and it doesn't make sense imo to have this one place arbitrarily be in feet.

Everything else looks good, and I am fine approving this, but it will be easier for users to input imperial units rather than metric. This is because the FRC rulebook uses imperial units, so if someone wants to create their own custom ruleset it will be easier for them to just input the value in feet, and they won't have to bother with converting units. 

(Also merge conflicts)

azaleacolburn (approved):

AlexD717 (approved):

BrandonPacewic (approved):

PR #1252, opened 2025-08-04
Multiplayer Support `[AARD-2024, AARD-2021]`

azaleacolburn:
## Task

AARD-2021  AARD-2022 AARD-2023 AARD-2060 AARD-2067

## Symptom

We want to support multiplayer

## Solution

There were three sprints done to determine which technology / architecture would be used for multiplayer: #1198, #1220, and #1240, The latter of which we ended up choosing, read that spike PR for details on the architecture.

This PR introduces a new `MultiplayerSystem`, which manages all the peer connections, handles all the sent and received messages, and keeps data about which objects are owned by which peers, including its own client. Every tick, `PhysicsSystem.update` sends either a collision or update message, as outlined in #1240.

To avoid having to send field assemblies over network every time one is spawned, at first, only the assembly name is sent over, the receiving client searches for the assembly in its cache. If it doesn't find the assembly, the receiving client sends an assembly request message back to the first client, which responds with the full assembly. 

To avoid any central stateful server (which peerjs is designed to avoid), joining rooms is done by encoding several hashes in the user id. This allows clients joining a room to easily identify peers with the same room id, verify that they also know the room id, and avoid exposing the room id to anyone who doesn't know it (and prevent people from enumerating rooms).

## Verification

0. `cd <synthesis>/multiplayer`
1. `bun i && bun run start`
2. Visit `http://localhost:3000/fission?randomId=1` in two tabs. Click the multiplayer button in both. Press create game in one and copy the code, then paste it and click join game in the other
4. Try spawning a field (there may be a small delay here), then a robot, then driving around and picking game pieces up
5. Try spawning another robot controlled by the other client and collide them

---

- [x] User-facing polish:


rutmanz:
- If you use separate browsers, intake and ejector configurations aren't shared and you get this problem 
image

- The 2018 field doesn't articulate anymore because nobody is technically touching it when it needs to move
- You can't shove more than one gamepiece at a time now (because nobody is touching the further away one

AlexD717:
Some issues that were found
1) Joints don't update
2) Drag mode on robots the other user spawned in doesn't work
3) Deleting robots you spawned in, sometimes doesn't work properly or leads to performance losses
4) Deleting robots other people spawned in results in a bunch of errors

_Aries helped test these and had the same issues_

azaleacolburn:
> 1. Joints don't update

Fixed! Thank you so much!
> 2. Drag mode on robots the other user spawned in doesn't work

This feature isn't exactly supported by the Hybrid architecture we're using. We could augment it by having each client keep a list of every body they touched, but that's not how we're handling it for now. I'm not sure if this is even a desired feature (what if they're trying to drive and you just pick them up, that leads to all sorts of issues. Another solution might be only sending over `collision` (or `drag`) messages while in drag mode. Attempting to drag another robot doesn't result in a bunch of errors or anything breaking (it just doesn't work), so it's not something I think we need to worry about for now.
> 3. Deleting robots you spawned in, sometimes doesn't work properly or leads to performance losses

I can't replicate this, could you provide more details about the conditions these performance issues occurred in (field, robot, etc)
> 4. Deleting robots other people spawned in results in a bunch of errors

Fixed!

Thank you so much for all the feedback!

AlexD717:
> > 2. Drag mode on robots the other user spawned in doesn't work
> 
> This feature isn't exactly supported by the Hybrid architecture we're using. We could augment it by having each client keep a list of every body they touched, but that's not how we're handling it for now. I'm not sure if this is even a desired feature (what if they're trying to drive and you just pick them up, that leads to all sorts of issues. Another solution might be only sending over `collision` (or `drag`) messages while in drag mode. Attempting to drag another robot doesn't result in a bunch of errors or anything breaking (it just doesn't work), so it's not something I think we need to worry about for now.

I agree that this isn't something we should worry about, and isn't a needed feature.

> > 3. Deleting robots you spawned in, sometimes doesn't work properly or leads to performance losses
> 
> I can't replicate this, could you provide more details about the conditions these performance issues occurred in (field, robot, etc)
> 

I don't remember the exact setup I had, but if I encounter it again, I will make sure to record the conditions.

PepperLola (changes requested):
Looks pretty good overall. I don't think one user should be able to remove another user's robot though, which is currently possible through the context menu.

Also, unexpected behavior happens if you configure your robot after spawning it. For example, I had two clients running and spawned a dozer in each, both of which were missing an ejector configuration. If I added an ejector to one and picked up a gamepiece, the other client would be spammed with "you need to configure an ejector first" toasts. I was able to configure the ejector on that second client to point in a different direction and have a different velocity, and when I ejected the gamepiece they went in different directions at different velocities depending on the client. I was only able to get this to happen once, but at one point after configuring the ejector on the second client, the first client lost possession of the gamepiece but was still considered to be holding it, so the robot could move around without the gamepiece but ejecting it shot the gamepiece in the direction of the ejector from where it was on the ground (even though the robot was across the field). When the ejectors don't match, the robot will often (after one or two ejections) get stuck and not be able to intake anymore.

PepperLola (changes requested):
I like that you can't use the context menu on another user's robot and that it's unavailable in the configure panel. I still had a couple issues though:
- When I spawn a robot in one window, then spawn one in another, both clients have their focus set to the second robot rather than the first client keeping focus on their own robot
- When both robots have no ejector configured, if I configure ejector on one and then pick up a gamepiece, the other client gets spammed with "configure ejector" notifications. The gamepiece does go to the correct place and is ejected properly for both users

EDIT:
- Should everyone be allowed to spawn a field? I feel like it would make more sense to only allow the room creator to spawn fields

azaleacolburn:
> When both robots have no ejector configured, if I configure ejector on one and then pick up a gamepiece, the other client gets spammed with "configure ejector" notifications.

Ok so I might just remove the toast for non-client objects

> Should everyone be allowed to spawn a field? I feel like it would make more sense to only allow the room creator to spawn fields

Yes, my original plan for multiplayer was that the field and game would be configured in the lobby, but we don't have one of those so we kind of just made it a collaborative sandbox environment. For now, I think this is out of the scope of this PR and should be deliberated on and addressed separately.

AlexD717:
Had an issue with scoring zones not syncing between users

PepperLola (approved):
Looks good overall, I think the issues that are left can be addressed in another PR tomorrow. Would be good to get this in today

rutmanz:
Remaining for follow up PR:
- zone / preference syncing on fields 
- focus fix
- assets recieved from peer sometimes not caching properly
- Metadata not syncing always
- (add more with an edit por favor)

AlexD717 (changes requested):
Requested Changes
Starting a new match, does not reset scores back to 0, even in singleplayer

Not needed changes
Change multiplayer button to say "Multiplayer (Beta)" (Or some other way to inform the users that this still has some issues.

rutmanz:
@AlexD717 do you have gamepieces in zones with persistent points? The old behavior was to reset to zero anyway and removing those gamepieces would result in negative scores, but I didn't think that made sense. I can change it back if you want though

AlexD717:
> @AlexD717 do you have gamepieces in zones with persistent points? The old behavior was to reset to zero anyway and removing those gamepieces would result in negative scores, but I didn't think that made sense. I can change it back if you want though

I think what you have right now is better and makes more sense

AlexD717 (approved):

PR #1253, opened 2025-08-04
Add Robot Spawn Positions `[AARD-2041]`

rutmanz:
## Task

AARD-2041

## Symptom

Currently robots spawn at the center of the field (2018 is particularly annoying here) and have no defined starting places for match mode. This means that they will start wherever they were last sitting before match mode started, which isn't necessarily what you want

This will be particularly important for multiplayer games where you have several robots to deal with
## Solution

Adds default spawn positions for each driver station to the devtools interface and allows them to be configured through there. I have configured default positions for the 2018 field and updated the assetpack
image

## Verification

1. Run `bun assetpack`
2. Spawn 2018 field v13
3. Spawn Dozer (and notice that it isn't inside a wall)
4. Start match mode and check the "Move Robots to Starting Positions" checkbox
5. Observe how your robot moved to that spot
---

- [x] User-facing polish:


fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -295,32 +309,78 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+    }
+
+    public moveToSpawnLocation() {
+        let pos: SpawnLocation = defaultRobotSpawnLocation()
+        if (this.miraType == MiraType.FIELD) {
+            pos = defaultFieldSpawnLocation()
+        } else {
+            const field = [...World.sceneRenderer.sceneObjects.values()].find(
+                (obj): obj is MirabufSceneObject => obj instanceof MirabufSceneObject && obj.miraType === MiraType.FIELD
+            )
```
azaleacolburn: I'm fairly certain that we have code just like this in the codebase already, at some point we should probably extract it into utility functions like `getField`, `getAllRobots`, etc

rutmanz: Created AARD-2048. I think this is out of the scope of this PR though

azaleacolburn (approved):

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -290,32 +304,78 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
         const bounds = this.computeBoundingBox()
         if (!Number.isFinite(bounds.min.y)) return
 
-        const offset = new JOLT.Vec3(
-            -(bounds.min.x + bounds.max.x) / 2.0,
-            0.1 + (bounds.max.y - bounds.min.y) / 2.0 - (bounds.min.y + bounds.max.y) / 2.0,
-            -(bounds.min.z + bounds.max.z) / 2.0
+        // If anyone has ideas on how to make this more concise I would appreciate.
+        // It took much longer than expected to deal with this
+        // (set position seems to use some arbitrary part of the robot, Dozer's is like half a meter in front to the left and 2471's is in the center)
```
BrandonPacewic, line 350: I believe that this is simply due to how the robot was exported. This would point to the design origin in fusion (which for Dozer is in front and slightly to the left).

ryanzhangofficial:
When I spawn two 2024 kitbots and 2024 field and start matchmode with the toggle on and different driver stations, they both snap to the center of the field and are on top of each other.

rutmanz:
@ryanzhangofficial This is because that field doesn't have actual points configured (I've only done 2018), but I changed the defaults so that they won't overlap anymore

I can create default spawn points for all of the fields but it might make more sense as a separate ticket (thoughts?)

AlexD717 (approved):
lgtm

Can this new FRC field be added to where our current fields are hosted?

ryanzhangofficial:
> @ryanzhangofficial This is because that field doesn't have actual points configured (I've only done 2018), but I changed the defaults so that they won't overlap anymore
> 
> I can create default spawn points for all of the fields but it might make more sense as a separate ticket (thoughts?)

I think that would be helpful. What happens if a user imports a custom mira field file though? It would be confusing to still have the toggle if the functionality isn't there.

BrandonPacewic (commented):
Some small conflicts from #1251 but otherwise looks good.

BrandonPacewic (changes requested):
I think to @ryanzhangofficial's point we should at least handle the case where there is no default positions but we still check the option. Currently it still teleports the robot. Maybe add a toast or gray out the option all together if no positions are set.

BrandonPacewic (approved):
Awesome user facing quality of life change.

ryanzhangofficial (approved):

PR #1254, opened 2025-08-04
Input Scheme Bug Fixes `[AARD-2039, AARD-2042, AARD-2038, AARD-2037]`

AlexD717:
## Task
There are a lot of small bugs with creating and assigning input schemes, and this PR solves all of them.

AARD-2039
AARD-2042
AARD-2038
AARD-2037

This branch was made from the UI refactor branch, which is why its showing so many line changes. This will have to be merged after the UI refactor is merged (#1241).

## Symptom
Current bugs
- No way to cancel creating an input scheme in the naming modal
Screenshot 2025-08-04 145353

- You can name a new input scheme with any name, including default names and other names already used. This can cause other bugs in the code that expect each robot to have a unique name (You could also have a blank scheme)
- In the initial robot config panel pressing the delete scheme button showed the deleted scheme as still there
- In the initial robot config panel, creating a new input scheme didn't show that scheme in the list
- In the Configure Assets Input tab, creating a new input scheme didn't show that scheme in the list
- In the initial robot config panel, newly created input schemes were not showing as in use
- A delete button was showing up for input schemes that were in use, in the initial robot config panel, even though the button wouldn't do anything
- There is no way to go back to the list of all input schemes from the input scheme configuration without clicking on robots/fields and then back to inputs.

- Going into configure assets then inputs, if you had a robot with a custom input scheme, it would show the configuration for that input scheme, and not all input schemes

## Solution
Fixed all the bugs outlined above by
- Adding name verification
- Adding a cancel button
- Creating an event that is called when InputSchemes is changed and tells the component to update the available schemes
- Hides delete button if scheme is in use

## Verification
Bugs outlined above no longer happen

---

- [ ] User-facing polish:


fission/src/ui/modals/configuring/inputs/NewInputSchemeModal.tsx
```diff
@@ -37,8 +70,13 @@ const NewInputSchemeModal: React.FC<ModalImplProps<void, void>> = ({ modal }) =>
...
             )
         }
-        configureScreen(modal!, { title: "New Input Scheme", hideCancel: true }, { onBeforeAccept })
-    }, [name, setSelectedScheme, openPanel, modal])
+
+        const onCancel = () => {
+            // Will automatically close the modal
+        }
+
+        configureScreen(modal!, { title: "New Input Scheme" }, { onBeforeAccept, onCancel })
```
PepperLola: There's not really a reason to include this (`onCancel`) if it doesn't do anything

fission/src/ui/panels/configuring/initial-config/InputSchemeSelection.tsx
```diff
@@ -26,15 +25,30 @@ export default function InputSchemeSelection({ brainIndex, onSelect, onEdit, onC
         SynthesisBrain.brainIndexMap.get(brainIndex)?.driveType ?? DriveType.ARCADE
     )
     const [availableSchemes, setAvailableSchemes] = useState<InputSchemeAvailability[]>()
-    useEffect(() => {
+
+    const refreshAvailableSchemes = () => {
```
PepperLola: Wrap this in a `useCallback`? (with `robotDriveType` dependency)

PepperLola (changes requested):
Looks pretty good, I just found a couple of improvements.

I think it makes more sense for the new input modal to stay open when there is an invalid name, just not let you create an input scheme with that name. Maybe add ui screen props for [disabling](https://mui.com/material-ui/react-button/#text-button) the accept (and maybe also cancel) button. Also, instead of adding a toast, maybe use the [text input error prop](https://muhimasri.com/blogs/mui-validation/#textfield-required-validation) to show whether the name is valid or not.

If I create a custom input scheme and keep the "configure inputs" panel open, then delete the input scheme, clicking accept or cancel in the configure inputs panel creates the input scheme again. I think maybe the configure panel should listen for the input scheme changed event and close itself if the one it's configuring no longer exists (but only when in the inputs mode).

If I try to edit an input scheme through the robot scheme part of the configure panel (configure panel -> click on robot -> controls -> set scheme -> edit button of scheme) it just opens the robot selection menu in the configure panel rather than the inputs tab.

When I create an input scheme and it opens the configure panel, if I open another configure panel instance and click on the inputs tab, it only lets me configure the same input scheme and doesn't show me the list of all of them. It's not a very big issue but it might be worth looking into if you can do something similar to what we do with [`configMode` and `configurationType`](https://github.com/Autodesk/synthesis/blob/55a36029816324ed1260f820ce5f67196ef51840/fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx#L117) to the selected scheme.

fission/src/ui/modals/configuring/inputs/NewInputSchemeModal.tsx
AlexD717: I have to have a onCancel, otherwise I get an error. I can make it a one line thing `const onCancel = () => {}`, if thats what you would prefer.

AlexD717:
> When I create an input scheme and it opens the configure panel, if I open another configure panel instance and click on the inputs tab, it only lets me configure the same input scheme and doesn't show me the list of all of them. It's not a very big issue but it might be worth looking into if you can do something similar to what we do with [`configMode` and `configurationType`](https://github.com/Autodesk/synthesis/blob/55a36029816324ed1260f820ce5f67196ef51840/fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx#L117) to the selected scheme.

I think we decided that we only wanted to support one panel of each type, so this is no longer a necessary fix.

Dhruv-0-Arora (changes requested):
Could you also add another bug fix / UI modification:

When clicking on an Input in the Inputs tab:

Screenshot 2025-08-05 at 6 11 31 PM

there is no way to navigate back to the general inputs area (image above)

Screenshot 2025-08-05 at 6 12 31 PM

Kinda like this:
Screenshot 2025-08-05 at 6 15 29 PM


Additionallly, another feature that is out of the scope of this PR is to make the configure panel show the joints configuration horizontally so it is more readable (not sure if this is possible but I believe it should be with UI refactor):

Screenshot 2025-08-05 at 6 14 40 PM

AlexD717:
> Additionallly, another feature that is out of the scope of this PR is to make the configure panel show the joints configuration horizontally so it is more readable (not sure if this is possible but I believe it should be with UI refactor):
> 

AARD-2002 should fix this. However, I would prefer this to remain vertical but add a scroll bar, rather than making it horizontal.

Dhruv-0-Arora (dismissed):
Conflicts; otherwise looks good

PepperLola (approved):
Looks good. Do we want canceling the configure panel that opens right after creating a custom input scheme to delete the input scheme? That would be pretty straightforward since you could just add it to the onCancel for that specific panel when opening. Not super important and I'm not sure if that would even be expected from a user's perspective.

BrandonPacewic (approved):
> Looks good. Do we want canceling the configure panel that opens right after creating a custom input scheme to delete the input scheme? That would be pretty straightforward since you could just add it to the onCancel for that specific panel when opening. Not super important and I'm not sure if that would even be expected from a user's perspective.

Made AARD-2077 just to reference this if we want to change it in the future.

PR #1255, opened 2025-08-04
Versioned Preference System

Dhruv-0-Arora:
AARD-2025

## Symptom

After removing/adding something to the preference system during development, users' preferences will be 'outdated' which can cause misalignments between what is stored in their local storage cache and 

## Task

Add a versioning to the Preference System that will enable Synthesis to migrate between versions when the production has been pushed to prevent users from running into issues when a new version is released and their local preferences are outdated.

## Solution

Added a `version` to global preferences to determine whether a client's versioning system is outdated & a migration system to translate the old preferences to be compatible with the new version of fission.

## Verification

Manually changing the version of the preference system and modifying any of the keys of the preferences themselves.

---

- [ ] User-facing polish:

### DIFF
--- a/fission/package.json
+++ b/fission/package.json
@@ -1,7 +1,7 @@
 {
   "name": "synthesis-fission",
   "private": false,
-  "version": "0.0.1",
+  "version": "7.2.0",
   "type": "module",
   "scripts": {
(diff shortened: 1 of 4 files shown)


Dhruv-0-Arora:
Close - not necessary

PR #1257, opened 2025-08-05
Upload File Accepts Only Mira Files `[AARD-2046]`

AlexD717:
## Task
Change the Mira upload file for robot/field uploads to only accept .mira files.

AARD-2046

## Symptom
Currently when navigating to spawn asset, and then import form file, you can select and upload any file (such as .mp4, .pdf, etc), even though uploading any of these formats will just result in an error message.

Trying to find where your .mira file in downloads is located, can also be annoying, but configuring the upload to only accept .mira files, makes it easier to find. This is because on Windows it automatically hides all unsuported file formats, and on MacOS it greys out unsupported files.

## Solution
Make it so that the input field only accepts .mira files

## Verification
- Can't upload a non .mira file
- Can upload a .mira file

---

- [ ] User-facing polish:


PepperLola (approved):
Looks good, I agree with this change

ryanzhangofficial (approved):

PR #1259, opened 2025-08-06
Implement adjustable center of gravity `[AARD-1921]`

itspvty:
## Task

AARD-1921

This PR adds a config for robots that allows you to move their center of gravity, and allows you to adjust the torque output consequence (the functional equivalent of scaling the weight of the robot).

## Symptom
A user may wish to observe the physical output of the robot when its weight is at a pre-calculated or desired position. This is most often used for setting safety limits on move/rotation speeds.

## Solution
To simulate a modified center-of-gravity, we compute the lever arm between where the gravitational force acts and where it "should" act, and gives us a rotational moment (torque). This torque is calculated, dampened (to avoid Jolt-related inaccuracies), and then distributed to each rigid body.

## Verification
Place the center-of-gravity at various points and observe the effect on the robot's balance. Verifying using real values is not feasible due to limitations in Jolt.

---

- [ ] User-facing polish:


fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -458,8 +506,28 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+                        World.physicsSystem.getBody(rootNodeId).GetWorldTransform()
+                    )
+                    const robotWorldPos = new THREE.Vector3()
+                    const robotWorldQuat = new THREE.Quaternion()
+                    const robotWorldScale = new THREE.Vector3()
+                    robotTransform.decompose(robotWorldPos, robotWorldQuat, robotWorldScale)
+
+                    const worldCoG = this._modifiedCenterOfGravity.clone()
+                    worldCoG.applyQuaternion(robotWorldQuat)
+                    worldCoG.add(robotWorldPos)
```
rutmanz: this code looks duplicated from earlier

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -903,6 +975,107 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+
+            const body = World.physicsSystem.getBody(bodyId)
+            if (body.IsStatic()) return
+
+            const inverseMass = body.GetMotionProperties().GetInverseMass()
+            if (inverseMass <= 0) return
+
+            const mass = 1 / inverseMass
+
+            const correctionFactor = (mass / totalMass) * this._cogEffectStrength
```
rutmanz: Why there is a coefficient here on the gravity force?

fission/src/ui/panels/configuring/assembly-config/interfaces/CenterOfGravityInterface.tsx
```diff
@@ -0,0 +1,222 @@
...
+    const relativePos = gizmoWorldPos.clone().sub(robotWorldPos)
+    relativePos.applyQuaternion(robotWorldQuat.clone().invert())
+
+    selectedRobot.modifiedCenterOfGravity = relativePos
+    selectedRobot.cogEffectStrength = effectStrength
+    selectedRobot.updateMeshTransforms()
+}
+
+const CenterOfGravityInterface: React.FC<CenterOfGravityInterfaceProps> = ({ selectedRobot }) => {
+    const gizmoRef = useRef<GizmoSceneObject | undefined>(undefined)
```
rutmanz, line 54: I think the physics of the robot should be paused while doing this so it doesn't move around while you're setting the CoM

fission/src/ui/panels/configuring/assembly-config/interfaces/CenterOfGravityInterface.tsx
```diff
@@ -0,0 +1,222 @@
...
+            <StatefulSlider
+                min={0}
+                max={2}
+                defaultValue={effectStrength}
+                label={"Effect Strength"}
+                onChange={value => {
+                    setEffectStrength(value)
+                    selectedRobot.cogEffectStrength = value
+                }}
+                tooltip="Adjusts how strongly the modified center of gravity affects the robot's physics (0 = no effect, 1 = normal, 2 = exaggerated)"
```
rutmanz: like what does exaggerated mean? Shouldn't there be a "correct" amount? And why is it a discrete integer slider?

rutmanz (commented):
Do we want to standardize our language around Center of Mass vs Center of Gravity?

rutmanz (changes requested):
oops meant to request changes

fission/src/ui/panels/configuring/assembly-config/interfaces/CenterOfGravityInterface.tsx
itspvty: > like what does exaggerated mean?

See above:
> This PR adds a config for robots that allows you to ... adjust the torque output consequence (the functional equivalent of scaling the weight of the robot).

> Shouldn't there be a "correct" amount? And why is it a discrete integer slider?

Yes

rutmanz: In that case I think it would make more sense to give it units and call it mass

itspvty: > In that case I think it would make more sense to give it units 

It's a scale
> and call it mass

sure

rutmanz (changes requested):
Doesn't look like the previous review was fully addressed

itspvty:
> Doesn't look like the previous review was fully addressed

Yes, I disagree with your suggestion that the physics be paused because if the user creates a try-and-repeat test scenario that involves any form of movement when pinpointing their CoG, this feature would prevent them from doing so. There are few or no cases where perpetual robot movement significantly impacts the selection.

It also wasn't made clear where the code was duplicated from (hence the emoji reaction), I searched for 5 minutes and then stopped with the presumption that it didn't matter enough to spend more time on; if it's truly a concern, then I will address it I suppose.

ryanzhangofficial:
@Dhruv-0-Arora I think this should probably have its own tutorial. Has a pretty specific use case. Easy to use wrong.

rutmanz (approved):

PR #1262, opened 2025-08-07
Hotfix: Chrome Rendering Issue `[AARD-2058]`

AlexD717:
## Task
In the latest release of Chrome (139), something broke in ThreeJS that resulted in robots/fields looking like this
Screenshot 2025-08-07 at 9 09 45 AM
when they are supposed to look like this
Screenshot 2025-08-07 at 11 10 56 AM

Each file (fields and robots) had this issue, with some being more severe than others.

AARD-2058

## Symptom
Robots and fields don't spawn in properly.

## Solution
Switch to using InstancedMesh, rather than BatchedMesh, as that fixes the issue. However, InstancedMesh is a lot worse performance wise that BatchedMesh, so we use InstancedMesh, only if the browser is on Chrome 139 on higher.

This is a short term solution, and we should look at how to better solve this. A similar issue was opened up on ThreeJS, and we sent a message in their Discord so hopefully ThreeJS fixes this.

The same issue also happens on the current production code of Synthesis, so we might want to consider merging this into prod. 

## Verification
- Robots are spawned in properly on Chrome 139
- Using a browser on a later version of Chrome uses BatchedMesh and spawned robots in correctly.

---

- [ ] User-facing polish:

### DIFF
--- a/fission/src/mirabuf/MirabufInstance.ts
+++ b/fission/src/mirabuf/MirabufInstance.ts
@@ -93,8 +93,8 @@ const transformGeometry = (geometry: THREE.BufferGeometry, mesh: mirabuf.IMesh) 
 class MirabufInstance {
     private _mirabufParser: MirabufParser
     private _materials: Map<string, THREE.Material>
-    private _meshes: Map<MirabufPartInstanceGUID, Array<[THREE.BatchedMesh, number]>>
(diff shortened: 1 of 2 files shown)


azaleacolburn (approved):
nice fix

rutmanz (approved):

BrandonPacewic (approved):

PR #1264, opened 2025-08-08
Clear All Data Button `[AARD-2059]`

AlexD717:
## Task
Create a button that deletes all stored data.

AARD-2059

## Symptom
When adding/testing new features, there were scenarios where I have wanted to clear all stored data, but there was no easy way to do that. (I have to go into chrome settings, clear data, etc.) 

## Solution
Add a button in debug tools that clears all saved data (not just preferences). A Conformation Modal was also added to make sure that the button is not accidently pressed.

## Verification
Pressing clear all data spawns a modal asking for confirmation. Clicking accept removes all preferences and cached robots. Clicking cancel closes the modal and all the data is still there.

---

- [ ] User-facing polish:


fission/src/ui/panels/DebugPanel.tsx
```diff
@@ -70,6 +71,30 @@ const DebugPanel: React.FC<PanelImplProps<void, void>> = ({ panel }) => {
...
+                                message:
+                                    "Are you sure you want to clear all preferences and cached data? This cannot be undone.",
+                                acceptText: "Clear & Reload",
+                                cancelText: "Cancel",
+                                onConfirm: () => {
+                                    window.localStorage.clear()
+                                    sessionStorage.clear()
+                                    window.location.reload()
+                                    console.log("All data cleared")
+                                },
```
rutmanz: Should also clear the OPFS caches for mirabuf stuff

```suggestion
                                onConfirm: async () => {
                                    window.localStorage.clear()
                                    sessionStorage.clear()
                                    await navigator.storage.getDirectory().then(async (root) => {
                                        for await (const key of root.keys()) {
                                            await root.removeEntry(key, {recursive:true})
                                        }
                                    }).catch(() => {console.warn("couldn't empty opfs")})
                                    window.location.reload()
                                    console.log("All data cleared")
                                },
```

rutmanz (changes requested):

rutmanz (approved):

fission/src/ui/modals/common/ConfirmModal.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export type ConfirmModalCustomProps = {
+    title?: string
+    message: string
+    acceptText?: string
+    cancelText?: string
+    allowClickAway?: boolean
+    onConfirm?: () => void
+    onCancel?: () => void
+}
```
PepperLola: A lot of this is redundant and already included in the regular (not custom) props. You could delete `title`, `acceptText`, `cancelText`, `allowClickAway` and just import them from `modal.prop` instead of `modal.props.custom`. For some reason there are some type issues with the modal/panel opening functions that I must have overlooked that make the linter complain about passing in the custom callbacks, but once those are resolved you will be able to use the regular `onAccept` and `onCancel` callback props instead of `onConfirm` and `onCancel`.

PepperLola (changes requested):
Left a comment, otherwise everything looks good. I will open a PR to fix the typing issues with the UI provider, so please wait for that to be merged and then you will be able to remove `onConfirm` and `onCancel` from the custom props.

EDIT: I also think this confirm modal could use a bit more thought - what happens if someone is doing something in a modal that we want to confirm? Opening the confirm modal would close their modal and they would lose their progress.  I suppose there could be custom logic per panel passed through the `onConfirm` or `onCancel` callbacks, so if that's the solution I would like to see a demo showing that it will work the way we want it to. Also, maybe `onCancel` should open the parent by default?

PepperLola (approved):
Looks good

PR #1265, opened 2025-08-08
Add Confirmation Before Dev Tool Removal and Modification [`AARD-2033`]

ryanzhangofficial:
## Task
Add a confirmation dialog when removing a Dev Tools–placed scoring zone, allowing the user to either remove it temporarily (current behavior) or permanently by also deleting it from the Dev Tools cache.

[Jira Ticket](https://jira.autodesk.com/browse/AARD-2033)

## Symptom
Currently if you remove a protected zone or scoring zone from the configure field panel that is placed by the dev tool it will remove it until the next refresh. Only the dev tool is able to cache scoring zone placement.

## Solution
Popup a confirmation window when we remove a scoring zone that was placed by dev tools, and give us the option of removing it from dev tools as well (saving to cache).

## Verification
- Tested temporary removal + refresh
- Tested permanent removal + refresh

Screenshot 2025-08-08 102909

---

- [ ] User-facing polish:


rutmanz (changes requested):
I think it would be better to say that the zones were defined in the field file/asset file/mirabuf file and avoid all reference to the developer tools. I think that, in theory, the developer tools are used by us to add this information to the cached files that people download, and so we shouldn't expect that they've used them / know they're there.

Also, this doesn't handle modifying the zones

fission/src/ui/modals/DevtoolZoneRemovalModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+            )
+            console.error(`Failed to ${actionText} zone in field file:`, error)
+        } finally {
+            setIsRemoving(false)
+            onClose()
+        }
+    }
+
+    return (
+        <Dialog open={isOpen} onClose={onClose} maxWidth="sm" fullWidth>
```
rutmanz: I think it's fine to make language consistent rather than doing this ternary stuff. Removing a zone is a modification or change to the field, and I think it makes the code a lot cleaner to use the same words

fission/src/ui/modals/DevtoolZoneRemovalModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+                    <Typography variant="body2" color="text.secondary">
+                        Choose how you'd like to{" "}
+                        {actionType === "modification" ? "save your modifications" : "remove it"}:
+                    </Typography>
+                    <Stack spacing={1}>
+                        <Typography variant="body2">
+                            <strong>Temporary {actionType === "modification" ? "modification" : "removal"}:</strong>{" "}
+                            {actionType === "modification"
+                                ? "Save changes until next field reload. Original zone will reappear when you refresh or reload the field."
+                                : "Remove zone until next field reload. The zone will reappear when you refresh or reload the field."}
```
rutmanz: "refresh or reload the field" is kind of a weird expression, "refresh the page" is much clearer IMO

fission/src/ui/modals/DevtoolZoneRemovalModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+                        </Typography>
+                        <Typography variant="body2">
+                            <strong>Permanent {actionType === "modification" ? "modification" : "removal"}:</strong>{" "}
+                            {actionType === "modification"
+                                ? "Save changes to the field file cache. This will persist your modifications on future loads."
```
rutmanz: "Save changes to the local asset file. This will persist your modifications until you remove it from the cache."

fission/src/ui/modals/DevtoolZoneRemovalModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+    isOpen: boolean
+    onClose: () => void
+    zoneType: "scoring" | "protected"
+    zoneName: string
+    onTemporaryRemoval: () => void
+    onPermanentRemoval: () => void
+    actionType?: "removal" | "modification"
+}
+
+const DevtoolZoneRemovalModal: React.FC<DevtoolZoneRemovalModalProps> = ({
```
rutmanz: should probably be renamed if this is being generalized

fission/src/util/DevtoolZoneUtils.ts
```diff
@@ -0,0 +1,171 @@
...
+): Promise<void> {
+    const field = World.sceneRenderer.mirabufSceneObjects.getField()
+    if (!field) throw new Error("No field loaded")
+
+    const parts = field.mirabufInstance.parser.assembly.data?.parts
+    if (!parts) throw new Error("No field parts found")
+
+    const editor = new FieldMiraEditor(parts)
+
+    if (zoneType === "scoring") {
```
rutmanz: I don't think it makes very much sense to have all of this in this conditional, it looks like you can generalize this for either zone type and avoid making unnecessarily specific code

fission/src/util/DevtoolZoneUtils.ts
```diff
@@ -0,0 +1,171 @@
...
+        const devtoolZones = editor.getUserData("devtool:scoring_zones") as ScoringZonePreferences[] | undefined
+        if (!devtoolZones) return
+
+        // Find and replace the zone in field file data
+        const updatedZones = devtoolZones.map(devZone => {
+            if (
+                devZone.name === originalZone.name &&
+                devZone.alliance === originalZone.alliance &&
+                devZone.parentNode === originalZone.parentNode &&
+                JSON.stringify(devZone.deltaTransformation) === JSON.stringify(originalZone.deltaTransformation)
```
rutmanz: it appears you're using these properties because they overlap on the two different zone preferences types. I think it would make sense to enforce this more rigorously by having a `BaseZonePreferences` type with the common properties that they both extend. This would be basically the same as `ScoringZonePreferences|ProtectedZonePreferences`, except that it would be robust aginst changes to those types

rutmanz: Alternatively, generating IDs on load for the zone preferences objects (they don't need to persist in the cache or across reloads) might be a more robust way to compare two zones that doesn't require the structuredclone

rutmanz (changes requested):

fission/src/util/DevtoolZoneUtils.ts
```diff
@@ -21,22 +33,14 @@ export function isZoneFromDevtools(
...
-                JSON.stringify(devZone.deltaTransformation) === JSON.stringify(zone.deltaTransformation)
-        )
-    } else {
-        // For protected zones, we'd need to add field file support first
-        // For now, return false as protected zones don't have field file support yet
+    if (zoneType === "protected") {
         return false
     }
+    
+    const devtoolZones = editor.getUserData("devtool:scoring_zones") as ScoringZonePreferences[] | undefined
```
rutmanz: You shouldn't need to cast this, if you do then you should update the interface to match instead

rutmanz (approved):
otherwise lgtm

AlexD717:
Why are we only adding this confirmation for scoring zones?


_Also Biome test fail_

ryanzhangofficial:
> Why are we only adding this confirmation for scoring zones?
> 
> _Also Biome test fail_

As noted in comments in code, protected zones are not supported yet.

AlexD717:
> > Why are we only adding this confirmation for scoring zones?
> 
> As noted in comments in code, protected zones are not supported yet.

Yeah, I can see that, I am just wondering why we are only adding this for scoring zones. Are we planning to add this for protected zones later on? From a user perspective it feels weird if they are only getting the popup for scoring zones, and nothing else.

ryanzhangofficial:
> > > Why are we only adding this confirmation for scoring zones?
> > 
> > 
> > As noted in comments in code, protected zones are not supported yet.
> 
> Yeah, I can see that, I am just wondering why we are only adding this for scoring zones. Are we planning to add this for protected zones later on? From a user perspective it feels weird if they are only getting the popup for scoring zones, and nothing else.

Yes that's the plan.

I think the order of merging should be this PR -> Zach's protected zone pr -> protected zone removal and modification

AlexD717:
Is it also intentional that you have to go into developer tools and press save to actually have scoring zones persist in cache? Currently spawning in a scoring zone and then refreshing doesn't have it persist. If this is the case, I am personally not a fan of this, and think that they should auto cache.


https://github.com/user-attachments/assets/62a81bcf-1ba9-4a75-86f7-dd0bb0804a42

rutmanz:
I agree with @AlexD717, but this appears to be the current behavior on dev as well

ryanzhangofficial:
> I agree with @AlexD717, but this appears to be the current behavior on dev as well

Yes I agree too, currently fixing this issue. It isn't a problem on prod so must've been an issue that came up throughout the summer.

ryanzhangofficial:
Would appreciate thoughts on this approach. I thought it would be appropriate because of the rewording of the modals, but at the same time it makes devtools seem less useful.

PR #1267, opened 2025-08-11
Match Mode GUI Config `[AARD-2026]`

AlexD717:
## Task
Match Mode GUI Config

AARD-2026

## Symptom
Currently, if users want to customize the match length, penalties, etc., they have to make their own json file following proper syntax, which isn't exactly user friendly.

## Solution
In the Match Mode Setup panel, a new option now exists to create a match mode config. 
Screenshot 2025-08-11 101815
Pressing that opens up a new panel that guides the user to create their own config.
Screenshot 2025-08-11 101904
Pressing accept starts match mode with the new config, and also saves the config to local storage allowing it to be used again.
There is also a Download as json button at the bottom, that downloads the entire config as a json file, allowing it to be shared across multiple devices.

## Verification
- After creating a configuration and pressing accept, your defined values is applied to the match
- Match mode config can be downloaded as a json, and then uploaded
- Can't download an invalid json config
- Can't enter letters into number only configs

---

- [ ] User-facing polish:


rutmanz (changes requested):
1. I think the button (and behavior) after going into that screen should be to save, rather than to start the match.
2. I think that it makes more sense to have a toggle to enable the height limit and extension limit, rather than saying to put in the string "Infinity"
3. I think having an edit button for an existing config (potentially instead of the delete button, and the delete button goes inside of the edit screen) would make sense
4. If there's an easy way to do it, I think the checkmarks should be replaced with a play button to make it clearer what they do

AlexD717:
> 1. I think the button (and behavior) after going into that screen should be to save, rather than to start the match.

Done
> 2. I think that it makes more sense to have a toggle to enable the height limit and extension limit, rather than saying to put in the string "Infinity"

Done
> 3. I think having an edit button for an existing config (potentially instead of the delete button, and the delete button goes inside of the edit screen) would make sense

I think this is out of the scope of this PR. Will make a separate Jira ticket for this
> 4. If there's an easy way to do it, I think the checkmarks should be replaced with a play button to make it clearer what they do

Done

fission/src/ui/components/StyledComponents.tsx
```diff
@@ -50,6 +51,7 @@ export class SynthesisIcons {
     public static readonly CONNECT = <GrConnect />
     public static readonly INFO = <AiOutlineInfoCircle />
     public static readonly BUG = <FaBug />
+    public static readonly PLAY = <FaPlay />
```
rutmanz: I think that `IoPlayOutline` from `react-icons/io5` feels more on-brand and less big

fission/src/ui/panels/configuring/CreateNewMatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,430 @@
...
+        document.body.removeChild(link)
+        URL.revokeObjectURL(url)
+    }, [createConfigFromForm])
+
+    useEffect(() => {
+        configureScreen(
+            panel!,
+            {
+                title: "Create New Match Mode Config",
+                cancelText: "Cancel",
```
rutmanz, line 303:
```suggestion
                cancelText: "Cancel",
                acceptText: "Save",
```

fission/src/ui/panels/configuring/CreateNewMatchModeConfigPanel.tsx
```diff
@@ -0,0 +1,430 @@
...
+        rules: [],
+        type: "checkbox",
+    },
+    enableHeightPenalty: {
+        defaultValue: false,
+        rules: [],
+        type: "checkbox",
+    },
+    maxHeight: {
+        defaultValue: fallbackConfig.maxHeight === Infinity ? 1.2 : fallbackConfig.maxHeight,
```
rutmanz: it might make sense to just change these defaults to not be Infinity (either `Number.MAX_SAFE_INTEGER` or just use like `-1` as a special value to be more semantically clear

rutmanz (changes requested):

PepperLola (approved):
Looks good. I do think that we should omit the id when downloading and uploading the JSON and have it generate one when uploading, since it doesn't really make much sense to keep and could potentially cause issues later (and just clutters the config file, if someone wanted to edit it manually). I can create a ticket for that since it's not within the scope of this PR.

rutmanz (approved):

BrandonPacewic (approved):
> Looks good. I do think that we should omit the id when downloading and uploading the JSON and have it generate one when uploading, since it doesn't really make much sense to keep and could potentially cause issues later (and just clutters the config file, if someone wanted to edit it manually). I can create a ticket for that since it's not within the scope of this PR.

I made AARD-2088 since I wasn't 100% sure a task had made made for this / I couldn't find it.

Looks good to me.

PR #1268, opened 2025-08-11
UI Post-Refactor Revisions `[AARD-2002]`

itspvty:
## Task

AARD-2002

This task involved collecting bugs relevant to the UI refactor from you guys and writing fixes for each one.

Task list:

- [x] Toasts have a "findDOMNode is deprecated" warning in devtools
- [x] The logo on the sidebar is draggable
- [x] Panels are too tall (even with no content)
- [x] The developer tool "No mira field loaded" red color is ugly, as well as some other red texts
- [x] You can open more than one unique panel
- [x] Panels on top of each other don't have proper separation (maybe add box shadow)
- [x] The "spawn" window and the "configure assets" windows shouldn't be able to be open at the same time
- [x] The dropdown shadows are too subtle/weird
- [x] Can't drag panels if pressing their title
- [x] The robot right-click context window shows the text "A Robot" in black at the top
- [x] Sliders do not display numerical values
- [x] There isn't enough spacing between many buttons (ex, match mode config menu)
- [x] The drag mode indicator at the bottom left has dark text on a dark bg
- [x] The timer on the scoreboard is dark text on dark bg
- [x] Where present, "delete" and "edit" buttons are placed weirdly (a line after their respective element often)
- [x] Add padding to the match mode config panel items
- [x] Block opening config/spawn when one of the two is already open
- [x] Fix slider overflow
- [x] Add info content to empty dropdowns
- [x] Pressing plus on the configure asset panel doesn't open the spawn assets panel
- [x] No feedback for why the config panel doesn't open when the spawn asset panel is open
- [x] Panels can be too tall
- [x] The configure panel for inputs skips to a specific scheme
- [x] The APS import folder doesn't have any text when empty
- [x] ~~Draggable region distinction~~
- [x] Fix analytics consent panel color
- [x] "Requires Browser Refresh" poorly styled; to tooltip
- [x] There are two scrollbars on tall panels
- [x] Blocked from closing the configure assets panel even with no configuration active
- [x] "Grab" cursor on draggable regions
- [x] Fix font sizes/logic throughout (titles/headers)
- [x] Fix spacing issues in more places
- [x] APS spinner alignment fix
- [x] Fix "submit" and "cancel" button overflow
- [x] Slider label/position uniformity
- [x] Fix input configure placements
- [x] Increase the maxHeight of all panels to 85vh
- [x] Make the panel title and button background the same color as the content
- [x] Keybind setters say "..." instead of "PRESS ANYTHING" when activated

## Symptom
The UI refactor was a large code edit and brought various visual bugs with it. The UI was, in most cases, fully usable, but the styling wasn't production-ready.

## Solution
Fixing individual issues over time by collecting purposeful data.

## Verification
Look good? Yay! Look bad? Aw man :(

---

- [ ] User-facing polish:


rutmanz:
- [ ] checkbox

PepperLola (changes requested):
Looks good overall but I noticed a couple things.

- I think the match mode config panel items could use some padding
- I'm not sure whether it makes more sense to close configure/spawn when the other is opened or to block the opening, because the user could be in the middle of configuring something and then opening spawn panel would cancel all their changes
- When dragging a slider in the settings modal all the way to the left or right the handle gets cut off, and on the right side it overflows and causes there to be a horizontal scroll bar
- I'm not sure how to address this, but I think the empty accordions in the spawn panel look a bit weird when expanded.

rutmanz:
> I'm not sure how to address this, but I think the empty accordions in the spawn panel look a bit weird when expanded.

Maybe just a message that is like "No Saved Assets" / "No Assets Found" rather than leaving it empty

rutmanz:
> I'm not sure whether it makes more sense to close configure/spawn when the other is opened or to block the opening, because the user could be in the middle of configuring something and then opening spawn panel would cancel all their changes

I think configure closing spawn is fine, but if configure is editing something (not just open) then spawn should show a warning

rutmanz:
1. I think that configure assets should also not be openable when the initial config panel (right after spawning) is open because it lets you configure the same robot in two different places which I don't think makes sense
2. The input screen doesn't look quite right still, and I also don't seem to be able to drag it anymore (use one of the controller input schemes)
image


3. Reset orientation is kind of weird here, maybe just have one reset button? 
image


4. I don't think we always want to see values (particularly for scalars that don't actually have real-world meaning) on sliders. Also "Preferences" should be more of a header IMO
image

rutmanz (changes requested):
see above ^

fission/src/ui/ThemeProvider.tsx
```diff
@@ -45,6 +45,33 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({ children }) => {
...
+                MuiMenu: {
+                    defaultProps: {
+                        elevation: 8,
+                    },
+                    styleOverrides: {
+                        paper: {
+                            boxShadow: "0 4px 12px rgba(0,0,0,0.28), 0 2px 6px rgba(0,0,0,0.18)",
+                        },
+                    },
+                },
```
AlexD717, line 74: What do you think about adding something like this? This blurs the background if you have a modal open which I think is better.
```suggestion
                },
                MuiBackdrop: {
                    styleOverrides: {
                        root: {
                            backgroundColor: "rgba(0, 0, 0, 0.5)",
                            backdropFilter: "blur(2px)",
                        },
                    },
                },
```

rutmanz (changes requested):
- [x] Pressing the plus button on the configure asset panel doesn't open the spawn assets panel anymore
- [x] There is no feedback for why the config panel doesn't open when the spawn asset panel is open. I think it makes more sense to make the newly-opened one replace the other, unless you're editing a configuration, at which point it should show a warning toast
- [x] Panels can still be too tall

Image

- [x] The configure panel for inputs seems to skip to a specific scheme instead of showing the list (and there are no breadcrums)
- [x] The APS import folder doesn't have any text when empty
- [x] If only parts of a panel are going to be draggable, there should be a visible distinction (maybe a handle like [this](https://react-icons.github.io/react-icons/search/#q=TbGridDots))

PepperLola (changes requested):
The analytics consent panel text is a weird color and the buttons (at least the close button) need fixing.
Image

fission/src/ui/components/Panel.tsx
```diff
@@ -57,21 +56,54 @@ export const Panel = <T, P>({ children, panel, parent }: PanelElementProps<T, P>
...
+                    overflowY: "auto",
                 }}
+                ref={nodeRef}
             >
-                {props.title && <CardHeader title={props.title} className="select-none" />}
-                <CardContent>
+                {props.title && (
+                    <CardHeader
+                        title={props.title}
+                        className="panel-drag-handle select-none"
```
rutmanz:
```suggestion
                        className="panel-drag-handle select-none hover:cursor-grab active:cursor-grabbing"
```

fission/src/ui/components/Panel.tsx
```diff
@@ -57,21 +56,54 @@ export const Panel = <T, P>({ children, panel, parent }: PanelElementProps<T, P>
...
+                        titleTypographyProps={{ variant: "subtitle1" }}
+                        action={
+                            <div
+                                className="h-2 w-10 rounded-sm"
+                                style={{
+                                    background:
+                                        "linear-gradient(90deg, rgba(255,255,255,0.2) 25%, rgba(0,0,0,0.2) 25%, rgba(0,0,0,0.2) 50%, rgba(255,255,255,0.2) 50%, rgba(255,255,255,0.2) 75%, rgba(0,0,0,0.2) 75%)",
+                                    backgroundSize: "8px 100%",
+                                }}
+                            />
```
rutmanz: ok so having seen how this looks, I think it's clearer to just use different cursor formats to indicate where it's draggable (suggestion above)

rutmanz (changes requested):
- [x] The configure panel for inputs seems to skip to a specific scheme instead of showing the list
  - Spawn Dozer
  - Go to configure assets > inputs
  - See that it shows the control scheme for the robot you just spawned rather than the list
- [x] The graphics settings "Requires Browser Refresh" thing is not styled as a header (which makes it look like another setting) and makes more sense just as a tooltip on anti-aliasing IMO
- [x] There are [two scroll bars](https://github.com/user-attachments/assets/3ecb58bf-d8b2-40cb-a155-cfbbbe84679a) on tall panels (configure inputs is where I found it)
- [ ] Blocks me from closing the configure assets panel even when there is no configuration active [img](https://github.com/user-attachments/assets/bce85bcf-9cd0-42ad-b7f6-6af24be7d94d)

ryanzhangofficial (dismissed):
Two issues: 

1) Font sizes/logic throughout. For example, in the attached image, the title "Assembly Setup" is so small compared to the two fields below it "Alliance" and "Station", which looks odd visually.

2) I don't think the button spacing has been fully addressed. It has been for the play buttons in match mode config. But not Create Match Mode Config and Upload File. I'm not sure if debug tools will still be in prod, but those buttons also seem cramped together.

I think Issue 2 can be addressed in a later ticket, but Issue 1 are changes in this PR that have made the UI worse.

Image

PepperLola (changes requested):
- I would like the APS spinner to be either the same height as the text (ideal) or centered vertically.

Image

- The default height of the initial config panel makes the accept/cancel buttons overflow. Maybe a worthwhile change would be making the modal/panel content scrollable but keeping the buttons visible (and the title as well)

Image

- If it's not too difficult it would be nice to have the slider labels match the other ones (in size + alignment/indentation); right now they look a bit out of place

Image

- This toggle group should be centered horizontally like in the spawn assets panel, and the buttons shouldn't touch the bottom of the panel (latter would be fixed by always showing accept/cancel)

Image

rutmanz (approved):
Approving because you've addressed my previous review and everybody else has caught my remaining feedback

AlexD717:
How hard would it be for you to go through and increase the maxHeight of all panels to something closer to 85vh? I think that would be nicer.

PepperLola (changes requested):
- Can the panel title and button background be the same color as the content? I prefer the lighter background color between the two of them
- Can we make the input buttons say "..." when you click them rather than "PRESS ANYTHING"?

PepperLola (changes requested):
Can the sliders here be stacked vertically?

Image

The title and buttons have the old background in modals

Image

Can we switch the various configure panels to use tabs instead of the toggle group?

Image

There are two intake duration sliders

Image

PepperLola (changes requested):
The requested changes were not addressed.

PepperLola (approved):
Looks good

PR #1269, opened 2025-08-12
Holds & Hides Cursor When Dragging View Cube `[AARD-1967]`

AlexD717:
## Task
Prevent the cursor from moving when the view cube is being held down to avoid issues where the cursor can fall offscreen. 

AARD-1967

## Symptom
If you have multiple monitors you can drag the cursor across the multiple monitors, and if the cursor hits a screen edge, you can no longer drag it.

## Solution
Use the pointer lock api to hide and lock the cursor when pressing on the view cube.

## Verification
- Rotating the view cube, results in the cursor being hidden
- You can now rotate the cube forever, where normally it would have been stopped by the screen edge.

---

- [x] User-facing polish:


fission/src/ui/components/ViewCube.tsx
```diff
@@ -61,27 +63,42 @@ const ViewCube: React.FC<ViewCubeProps> = ({
...
-                const deltaY = event.clientY - lastMousePos.y
+            if (!isDragging) return
+
+            let deltaX = 0
+            let deltaY = 0
+
+            if (isPointerLocked) {
+                // Use movementX and movementY when pointer is locked
+                deltaX = event.movementX || 0
+                deltaY = event.movementY || 0
```
rutmanz: why `|| 0`?

fission/src/ui/components/ViewCube.tsx
```diff
@@ -91,21 +108,51 @@ const ViewCube: React.FC<ViewCubeProps> = ({
...
                 setDragStartElement(null)
+
+                if (document.pointerLockElement) {
+                    document.exitPointerLock()
+                }
+            }
+        }
+
+        const handlePointerLockChange = () => {
+            setIsPointerLocked(!!document.pointerLockElement)
```
rutmanz:
```suggestion
            setIsPointerLocked(document.pointerLockElement != null)
```

rutmanz (changes requested):
otherwise lgtm

itspvty (approved):
Very cool

fission/src/ui/components/ViewCube.tsx
AlexD717: In certain edge cases event.movementX and event.movementY could be undefined.

rutmanz (approved):

ryanzhangofficial (approved):
Looks good!

fission/src/ui/components/ViewCube.tsx
```diff
@@ -848,6 +878,12 @@ const ViewCube: React.FC<ViewCubeProps> = ({
             setLastMousePos(mousePos)
             setDragStartPos(mousePos)
             setDragStartElement(getClickedElement(event))
+
+            // Request pointer lock for better cursor control
+            if (pointerLock.current && containerRef.current) {
```
PepperLola: What is the purpose of the `pointerLock` ref? This is the only place I see it being referenced and the value is never updated. Can it be removed?

fission/src/ui/components/ViewCube.tsx
azaleacolburn: Wouldn't we use `??` in that case then?

fission/src/ui/components/ViewCube.tsx
AlexD717: It can be removed. It was just there to allow me to easily test what would happen if the browser didn't support pointerLock.

BrandonPacewic (approved):

itspvty:
Introduces issue addressed by #1289

PR #1274, opened 2025-08-14
Command Tool `[AARD-2079]`

AlexD717:
## Task
Create a command tool, that allows synthesis users to type in "/" followed by a command that then executes a specific action. 

AARD-2079

## Symptom
It can get annoying for developers and users to have to go through multiple panels to achive what they want. For advanced users of Synthesis a command tool would let them quickly type commands to implement a specific action. For example, instead of having to go to debug tools, then clicking toggle drag mode, then clicking hide debug tools panel, you could just type "/drag mode" and it will toggle the drag mode.

## Solution
Implement a simple Command palette that appears when pressing "/" and lists options of different commands. It uses an algorithm to determine what you are most likely typing and then suggesting that as an option. This means that both "/drag" or "/drqg" (typo example) will triggle the toggle drag mode command.

## Verification

- Pressing "/" opens up a command pallet
  - Doesn't open in the welcome screen (singleplayer/multiplayer selection screen)
  - Doesn't open if a modal is open
- Pressing "esc" closes the command palette
- Starting to type in a command has different options appear, with options that you most likely want appearing at the bottom
- pressing enter runs the command you have selected (or clicking on the command with the mouse)
- Can use arrow keys to navigate between selected commands (or mouse)
- Clicking outside of the command palette closes it

---

- [ ] User-facing polish:


azaleacolburn:
> Doesn't open in menu selection

1. I'm not sure what you mean by this, but it does still open when you have panels open (including the main menu panel which is what I assume you mean). If you mean within the command palette, then yes indeed typing `/` doesn't open another one.
2. I like how the drag mode command works, I think the commands that open panels should toggle them instead.
3. I think that trying to navigate past the first or last option should wrap around to the other end of the list.

AlexD717:
> > Doesn't open in menu selection
> 
> 1. I'm not sure what you mean by this, but it does still open when you have panels open (including the main menu panel which is what I assume you mean). If you mean within the command palette, then yes indeed typing `/` doesn't open another one.

I meant the welcome screen. Will clarify that
Screenshot 2025-08-14 at 10 38 03 AM

I agree with your other suggestions and will implement them

AlexD717:
> 2. I like how the drag mode command works, I think the commands that open panels should toggle them in

Waiting on pr #1268 to be merged in first, as that implements support for detecting if panels/modals are already opened

fission/src/ui/components/CommandPalette.tsx
```diff
@@ -0,0 +1,374 @@
...
+    }, [])
+
+    const openPalette = useCallback(() => {
+        setIsOpen(true)
+        setTimeout(() => inputRef.current?.focus(), 0)
+    }, [])
+
+    const openImportPanel = useCallback(
+        (configurationType: ConfigurationType) => {
+            openPanel<void, { configurationType: ConfigurationType }>(ImportMirabufPanel, { configurationType })
```
rutmanz:
```suggestion
            openPanel(ImportMirabufPanel, { configurationType })
```

rutmanz (changes requested):
1. Looks like typing in the search bar also drives robots around (WASD and arrow keys)
2. I think whenever you reference a robot name, you should also reference the control scheme name (currently two dozers can't be distinguished)

rutmanz (changes requested):
The Luna scheme conflicts with this binding. I would either make the triggering key rebindable or prevent inputs from using it

rutmanz (approved):

BrandonPacewic (changes requested):
I’d prefer if we set up a system where other parts of the simulator can register their own commands rather than hard-coding a single inline list here. That way each feature owns its own actions, and the palette just queries a shared registry.

I’m fine with keeping the inline list temporarily so we can get this merged without blocking other work, but we should make an effort to lay out how this will be adopted within this current PR.

AlexD717:
> I’d prefer if we set up a system where other parts of the simulator can register their own commands rather than hard-coding a single inline list here. That way each feature owns its own actions, and the palette just queries a shared registry.
> 

Added support for this, and moved the commands that made sense to the appropriate files

itspvty (approved):
I really like this concept overall and the use of fuse.js is very cool.

ryanzhangofficial (approved):
love it!

BrandonPacewic (approved):

PepperLola (changes requested):
Could you get rid of this highlight ring?

Image

Also, I think it would be worth adding options for the different tabs of the configure panel just like there are for spawning.

AlexD717:
> Could you get rid of this highlight ring?
> 
> Image

In my opinion, the command pallet without any border looks weird, when you have something other than the bottom part selected.
Screenshot 2025-08-22 082846


If I make the boarder 1px dark grey, it adds a slight line, which separates the text box, making it more clear. 
Screenshot 2025-08-22 083021

 Are you fine with the second option or do you still think the first is better? (@PepperLola)

azaleacolburn:
I like the last option best personally

PepperLola (changes requested):
I like the separator but I think it would make sense either to have the border around the whole thing or just have a separator between the input and the options (only the top of the input), and maybe then get rid of the border radius on the top left and right.

PepperLola (changes requested):
When I added a bunch of new commands, they weren't all accessible since it was only including enough to fit. I think the list should have a max height but should be scrollable and include all the registered commands
Image

AlexD717:
> I like the separator but I think it would make sense either to have the border around the whole thing

The border is around the whole thing right now

> When I added a bunch of new commands, they weren't all accessible since it was only including enough to fit. I think the list should have a max height but should be scrollable and include all the registered commands Image

If you start typing it very quickly filters them out

PepperLola:
> If you start typing it very quickly filters them out

That's true, but I think you should be able to scroll through all of them to see what options are available. It's true that you can filter some out so I guess it's not super important to include in this PR but I definitely think it would be worthwhile to add.

EDIT: on second thought, I think it should be included in this PR as it's pretty core functionality and I think necessary for a good user experience

PepperLola (approved):
I think it looks good. It would be nice if the scrolling were smooth, and there were a couple bugs I couldn't figure out how to consistently reproduce (sometimes it would always scroll the selected option to the second slot from the top, regardless of whether I was scrolling up or down). I think it's in a good state to be merged, keeping these fixes in mind for the future.

PR #1275, opened 2025-08-14
Custom Event Refactor `[AARD-2080]`

rutmanz:
## Task

AARD-2080

## Symptom

Currently there are a bunch of custom event classes with inconsistent implementations that both make it harder to find events and make the whole system more verbose. Unsubscription is also often done incorrectly or not done at all

## Solution

I introduced an `EventSystem` class that manages all custom events and has generic `create`, `dispatch`, and `listen` methods. 

## Verification

Go through and try and trigger the list of events found in `EventDataMap` and confirm that they still work as before. This PR should not affect functionality

---

- [x] User-facing polish:


rutmanz:
(because other open PRs add new events in ways that won't necessarily show up as conflicts, I'll want to do one last check before this is merged)

AlexD717 (approved):
lgtm

fission/src/mirabuf/ProtectedZoneSceneObject.ts
```diff
@@ -136,21 +136,22 @@ class ProtectedZoneSceneObject extends SceneObject {
...
-                        this.zoneCollisionRemoved(body1)
-                    }
-                }
-                OnContactRemovedEvent.addListener(this._collisionRemoved)
+                this._unsubscribers.push(
+                    EventSystem.listen("OnContactRemovedEvent", ({ message }) => {
+                        const body1 = message.GetBody1ID()
+                        const body2 = message.GetBody2ID()
+
+                        if (body1.GetIndexAndSequenceNumber() == this._joltBodyId?.GetIndexAndSequenceNumber()) {
```
azaleacolburn, line 148: This condition appears a lot, could we perhaps extract it into a function like `bodyEq(body1: Jolt.BodyID, body2: Jolt.BodyID): bool`.

azaleacolburn (changes requested):

fission/src/mirabuf/ProtectedZoneSceneObject.ts
rutmanz: not sure why that would be better or in scope?

azaleacolburn (approved):

BrandonPacewic (changes requested):
There are some non-trival conflicts here. Since this is not user facing this will not be put in for `v7.2`.

rutmanz:
Just merged `dev` and dealt with the conflicts

BrandonPacewic (approved):
I went through and double checked changes to events made pre and post this PR. I believe we should be good to go. If anything else comes up I'll open another PR.

Thanks for doing this. Love me a good refactor!

PR #1276, opened 2025-08-15
Store Protected Zones in Devtools `[AARD-2064]`

rutmanz:
## Task

AARD-2064

## Symptom

Currently protected zones, which are a field specific property, aren't stored in the mirabuf file via developer tools and have to be configured by each user

## Solution

Added protected zones to the developer tools menu to allow them to be persisted into the mirabuf data

## Verification

1. `bun run assetpack`
2. Open synthesis, clear your preferences, and import `FRC Field 2023_v7.mira`
3. Observe that scoring zones and protected zones are already configured
4. Drive around in match mode to confirm that they actually work properly

---

- [x] User-facing polish:


AlexD717 (approved):
lgtm

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -316,7 +316,7 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
     }
 
     // Centered in xz plane, bottom surface of object
-    private getPositionTransform(vec: THREE.Vector3) {
+    public getPositionTransform(vec: THREE.Vector3 = new THREE.Vector3()) {
```
azaleacolburn, line 321: What's going on here? Why are we doing this?

fission/src/Window.d.ts
```diff
@@ -2,4 +2,5 @@ declare interface Window {
     convertAuthToken(code: string): void
     gtag?: (command: "config" | "set" | "get" | "event" | "consent", ...args: unknown[]) => void
     dataLayer?: unknown[][]
+    World?: unknown
```
azaleacolburn, line 5: Why are we storing World in Window? Don't we usually just import it?

azaleacolburn (commented):
I'm a bit confused by some of the things you're doing, both with how they relate to this PR and why we would want to do them in general.

rutmanz:
The answer to both of those comments is that they are to allow you to access helpful information from the browser devtools, which famously cannot import `World` or `THREE.Vector3`. The reason it's part of this PR is that it can be quite useful when trying to get positioning information when configuring predefined zones (I found myself wanting it when defining the 2023 field zones).

https://github.com/Autodesk/synthesis/blob/d9474b23e2257b374c3b5af860554e71d866555e/fission/src/systems/World.ts#L90-L92

fission/src/Window.d.ts
rutmanz: It's typed as unknown to prevent using it in code, see other comment

fission/src/mirabuf/MirabufSceneObject.ts
rutmanz: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

azaleacolburn (approved):

BrandonPacewic (approved):
This pull request is 3rd in que.

PR #1281, opened 2025-08-18
Scoreboard Shows in Match Mode `[AARD-2082]`

AlexD717:
## Task
Make it so that the scoreboard auto shows up when starting a match

AARD-2082

## Symptom
Currently, if users have the scoreboard disabled in preferences, and then start a match the scoreboard is hidden. This can lead them to think that match mode is broken, as there is no timer anywhere.

## Solution
Scoreboard will now always show up in match mode, ignoring user preferences, and then go back to user preferences on match end.

## Verification
Disable the scoreboard in preferences, and verify that the scoreboard disappears. Start match mode, the scoreboard should appear. Pressing "Back to Sandbox" on match end makes the scoreboard disappear.

---

- [ ] User-facing polish:


azaleacolburn (approved):
nice work!

rutmanz (approved):

ryanzhangofficial (approved):

PepperLola (approved):
Currently if the scoreboard is disabled, when you enter match mode it will display but the timer will be zero until it updates. If the scoreboard is already visible, the time displays correctly immediately. We can fix that some other time; for now I just want to get this merged.

PR #1287, opened 2025-08-20
Turn Off Biome Import Sorting `[AARD-2096]`

azaleacolburn:
## Task

AARD-2096

## Symptom

Previously we were using `bun run style` for checking which ran `biome format` and `biome lint`, neither of which sorted imports. `biome check` on the other hand, runs both commands and checks if the imports are sorted.

Sorting imports is not desired, as it makes diffs unnecessarily large for minimal hygenic and aesthetic benefit, but calling two commands when there exists one that already calls both isn't either.

## Solution

Switch `bun run style` to use `biome check` and biome no longer sorts imports.

## Verification

- Move import statements out of order, run `bun run style:fix` (or `biome check --fix`) and observe that they have not been reordered.

---

- [x] User-facing polish:


rutmanz (approved):

itspvty (approved):

AlexD717 (approved):

ryanzhangofficial (approved):

BrandonPacewic (approved):
This pull request is first in que.

PR #1291, opened 2025-08-21
Use JSON file for match mode configs `[AARD-2086]`

rutmanz:
## Task

AARD-2086

## Symptom

Currently default match mode configs are defined in-code, which means that updating the list (as would be necessary on kickoff) would require a redeploy of synthesis

## Solution

This changes them to be defined by a JSON file in the assetpack, which is loaded on startup

## Verification

Check that the configurations still show up and work correctly
This does change the assetpack, so you'll have to run `bun run assetpack` before testing

---

- [x] User-facing polish:

### DIFF
--- a/fission/src/systems/match_mode/DefaultMatchModeConfigs.ts
+++ b/fission/src/systems/match_mode/DefaultMatchModeConfigs.ts
@@ -1,70 +1,44 @@
 import type { MatchModeConfig } from "@/ui/panels/configuring/MatchModeConfigPanel"
-import { convertFeetToMeters } from "@/util/UnitConversions"
+import { API_URL } from "@/util/Consts.ts"
+
+type ManifestMatchModeConfig = Omit<MatchModeConfig, "id"> & { id: string }
+interface MatchConfigManifest {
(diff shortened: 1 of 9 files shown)


ryanzhangofficial (approved):
makes sense!

fission/src/systems/match_mode/DefaultMatchModeConfigs.ts
```diff
@@ -1,70 +1,44 @@
...
-            sideExtensionPenalty: 5,
+    static async reload() {
+        const manifest = await fetch(this.MANIFEST_LOCATION)
+        const json: MatchConfigManifest | undefined = await manifest.json().catch(e => {
+            console.error(e)
+            return undefined
+        })
+        if (json == null) {
+            console.error("Could not load match mode manifest")
+            return
```
AlexD717: Shouldn't this or the above return statement be changed for consistency so that they both either return nothing or undefined?

AlexD717 (approved):
Looks good, just a very small nitpick

BrandonPacewic (approved):

PR #1301, opened 2026-01-10
feat: import unstable mira files with warnings

BrandonPacewic:
## Symptom

With CAD models imported from other pieces of software there is often geometry that is unresolvable causing the importer to crash.

## Solution

Detect and log common fail points of "unstable" mira files and pass over objects that are not resolvable, rather than crashing. 

## Verification

Importing in fission works normally.

---

- [ ] User-facing polish:


PepperLola (approved):
Worked for me, I was able to import the new field. The one with all the gamepieces has big performance issues but it imports correctly.

PR #1306, opened 2026-04-30
chore(nix): various dev shell improvements

0xda157:
The point of a nix dev shell is to include tools needed to interact with the project, so one doesn't have to install them globally. Therefor we should include any programs required, not just playwright.

I also changed shebangs from `/bin/bash` to `/usr/bin/env bash` which allows users of OSs that don't install programs to `/bin` (such as nixos) to run the scripts

Direnv allows for automatically entering and exiting nix shells when cd-ing in/out of the directory in question, which makes thing much more convenient for nix users. It also removes the possibility of using a shell which no longer reflects the code, which can be a common source of issues.


.gitignore
```diff
@@ -8,4 +8,5 @@ renderer/
 .DS_Store
 *.pkg
 *.exe
-.idea
\ No newline at end of file
+.idea
+**/.direnv
```
HunterBarclay, line 12: @BrandonPacewic What's the stance on newlines at EOF?

fission/flake.nix
```diff
@@ -17,8 +17,11 @@
     in
     {
       devShells = forEachSupportedSystem (pkgs: {
-        default = pkgs.mkShell {
```
HunterBarclay, line 20: Provide fission as the default shell.

```nix
devShells = forEachSupportedSystem (pkgs: rec {
  fission = pkgs.mkShell {
    ...
  };
  default = fission;
  ...
});
```

flake.nix
```diff
@@ -27,8 +30,19 @@
...
           };
         };
+        exporter = pkgs.mkShell {
+          packages = with pkgs; [
+            python3
+            black
+            isort
+            bun
+          ];
+        };
```
HunterBarclay: @azaleacolburn Shell for using emscripten to compile Jolt?

HunterBarclay (dismissed):
@0xda157 Could you provide a Nix README file and add a section to the main README referring to it? I think this has some really good potential, but as someone who has never really looked into Nix before reading this PR, it took me like an hour to figure out how to even start a shell correctly.

HunterBarclay:
@0xda157 I'd also suggest shell hooks for the following

### `fission`
```nix
shellHook = "cd ./fission"
```

### `exporter`
```nix
shellHook = "cd ./exporter/SynthesisFusionAddin"
```

0xda157:
> Could you provide a Nix README file and add a section to the main README referring to it? I think this has some really good potential, but as someone who has never really looked into Nix before reading this PR, it took me like an hour to figure out how to even start a shell correctly.

I added a section to contributing.md about it, but I'm not so sure a guide on how to install/use nix belongs in the synthesis contributing guide.



> I'd also suggest shell hooks for the following
>
> ```nix
> shellHook = "cd ./fission"
> ```
> 
> ```nix
> shellHook = "cd ./exporter/SynthesisFusionAddin"
> ```

Shell hooks generally aren't used like this, because shells can be entered from anywhere (e.g. running `nix develop` inside `./fission`). It'd be better to do the reverse using direnv. Once you install direnv and `direnv allow` the `.envrc` file direnv will automatically run `nix develop` for you when you change directories into the relevant directory.

.gitignore
BrandonPacewic: Blank newline at EOF is preferred.

BrandonPacewic (approved):
Thanks for the contribution! Made a few changes on top:

- Removed the `NixFlakeCheck.yml` CI workflow. Very few contributors currently use Nix, so adding CI checks for it would add overhead for everyone. The flake can be validated locally with `nix flake check`.
- Removed the Nix section from `CONTRIBUTING.md` for the same reason. Moved the relevant install/setup links into comments at the top of `flake.nix` itself, where Nix users will naturally look.
- Added a `multiplayer` dev shell and updated `multiplayer/.envrc` to point to it instead of `.#fission`.
- Added a root-level `.envrc` so direnv activates when working from the repo root.
- Standardized `fission` shell to use `packages` instead of `nativeBuildInputs` for consistency with the `exporter` shell.
- Fixed a typo in a comment: "deviations" to "derivations".

.gitignore
Dhruv-0-Arora: ^ bumping this up

Dhruv-0-Arora (approved):

PR #1308, opened 2026-06-15
Remove Redundant Setters/Getters `[SYNTH-110]`

azaleacolburn:
## Task

AARD-2150

## Symptom

Currently, there are several getters and setters that are not necessary for one of the following reasons:

- They do nothing besides return / set a property, so the property should be `public`
- The property is not `public` and is never set

## Solution

This PR removes those getters / setters

## Verification

All tests pass

---

- [x] User-facing polish:

### DIFF
--- a/fission/src/mirabuf/MirabufSceneObject.ts
+++ b/fission/src/mirabuf/MirabufSceneObject.ts
@@ -83,9 +83,9 @@ export function getSpotlightAssembly(): MirabufSceneObject | undefined {
 }
 
 class MirabufSceneObject extends SceneObject implements ContextSupplier {
-    private readonly _assemblyName: string
+    public readonly assemblyName: string
     public readonly mirabufInstance: MirabufInstance
-    private readonly _mechanism: Mechanism
+    public readonly mechanism: Mechanism
 
     private _brain: Brain | undefined
     public alliance: Alliance | undefined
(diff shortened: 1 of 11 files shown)


fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -105,14 +105,14 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
 
-    private _nameOverride?: string
+    public nameOverride?: string
     private _nameTag: SceneOverlayTag | undefined
     private _centerOfMassIndicator: THREE.Mesh | undefined
     private _basePositionTransform: THREE.Vector3 | undefined
-    private _intakeActive = false
-    private _ejectorActive = false
+    public intakeActive = false
+    public ejectorActive = false
```
Dhruv-0-Arora, line 113: Are these variables ever referenced anywhere else in the project? If not we should remove them (might be out of the scope of this PR but I don't think it deserves its own).
I don't exactly see why these variables are necessary.

fission/src/systems/World.ts
```diff
@@ -45,9 +45,7 @@ class World {
     public static get sceneRenderer() {
         return World._sceneRenderer
     }
-    public static get physicsSystem() {
-        return World._physicsSystem
-    }
+
     public static get simulationSystem() {
         return World._simulationSystem
     }
```
rutmanz, line 55: why did you change physics system but none of the others?

rutmanz (commented):
Should probably wait for #1309 because unit tests aren't running

fission/src/systems/World.ts
azaleacolburn: This shouldn't have even been changed in the first place

fission/src/mirabuf/MirabufSceneObject.ts
azaleacolburn: They are referenced elsewhere, the question as to whether they are necessary is separate. It is set in `SynthesisBrain.ts`, `IntakeDriver.ts` and its value is referenced in `IntakeSensorSceneObject.ts`.

Dhruv-0-Arora (approved):

0xda157 (approved):
LGTM

PR #1312, opened 2026-06-16
Add 2026 kitbot asset `[SYNTH-89]`

rutmanz:
## Task

SYNTH-89

## Symptom

Currently the most recent "real" robot is the 2024 kitbot, which is 2 years out of date and won't give new students a robot they've actually seen

## Solution

I exported the 2026 kitbot cad and added it to the assetpack

## Verification

After adding to assetpack, I spawned the robot and drove it around

---

- [x] User-facing polish:


Dhruv-0-Arora (approved):
Maybe we can later not have all assets in one zip and instead spread them out so users could maybe pick and choose their assets (not sure how this would be beneficial it just seems better practice).

AlexD717 (approved):
lgtm

Something that should be considered for the future is that currently this Kitbot doesn't really behave in a realistic way. This could result in some users leaving, since the robots just don't feel really realistic. Ex, for this Kitbot it feels too fast and drifts when turning, making it feel more "cartoony" than realistic.

PR #1315, opened 2026-06-17
fix(UX): opening panel while assembly setup is open forces refresh `[SYNTH-120]`

Dhruv-0-Arora:
## Symptom

SYNTH-120

If a user tries to spawn an assembly before completing with another one, they are hit with this error. Even with the panel closed and the assembly completed, there is no way to spawn another assembly.

image

## Solution

There was a missing react hook dependency, so I added it.

---

- [ ] User-facing polish:


Dhruv-0-Arora:
> [!NOTE]
> This moves the logic of blocking the 'import' panel from being opened when 'Assembly Setup' panel is open to `UIProvider` (from `ImportMirabufPanel`).
> This also prevents any panel from being opened while the 'Assembly Setup' panel is open

fission/src/ui/UIProvider.tsx
```diff
@@ -145,6 +148,9 @@ export const UIProvider: React.FC<UIProviderProps> = ({ children }) => {
                     setPanels(p => [...p.filter(x => x !== existingDuplicate), existingDuplicate])
                     return existingDuplicate.id
                 }
+            } else if (unconfirmedImport) {
+                addToast("warning", "You are currenting importing a model!", "Confirm it before opening another panel.")
```
PepperLola:
```suggestion
                addToast("warning", "You are currently importing a model!", "Confirm it before opening another panel.")
```

PepperLola (dismissed):
This now blocks any panel from opening while an import is in progress which, although potentially inconvenient, fixes the issue where you can start an import, click configure assets, and then the panel is dismissed but other panels are blocked from opening. I also don't really like including logic specific to a single panel in the UIProvider for the whole UI - maybe it could make sense to put a `blocking` flag on a panel that prevents other panels from opening while it's open?

rutmanz (changes requested):
I'm inclined to agree with @PepperLola, it feels really weird to put logic specific to one panel into the UIProvider file. If you can't put it in the panel's file, then I like his idea of making a generic "blocking" property that prevents other panels from opening at the same time

fission/src/ui/components/MainHUD.tsx
```diff
@@ -189,7 +189,15 @@ const MainHUD: React.FC = () => {
...
-                    <MainHUDButton startIcon={SynthesisIcons.WRENCH} onClick={() => openPanel(ConfigurePanel, {})}>
+                    <MainHUDButton
+                        startIcon={SynthesisIcons.WRENCH}
+                        onClick={() =>
+                            openPanel(ConfigurePanel, {}, undefined, {
+                                blocking: true,
+                                blockingMessage: "Close Assembly Setup before opening another panel.",
+                            })
+                        }
+                    >
```
Dhruv-0-Arora: this is going to conflict with topbar

fission/src/ui/helpers/StateProviderHelpers.ts
```diff
@@ -6,9 +6,6 @@ export interface StateProviderProps {
 }
 
 export interface AppState {
-    // ImportMirabufPanel
-    unconfirmedImport: boolean
-    setUnconfirmedImport: (_state: boolean) => void
```
Dhruv-0-Arora, line 11: cleaning this up entirely

fission/src/ui/UIProvider.tsx
```diff
@@ -180,27 +190,6 @@ export const UIProvider: React.FC<UIProviderProps> = ({ children }) => {
...
-                        if (isActivelyConfiguring) {
-                            // Show a warning toast about unsaved configuration
-                            enqueueSnackbar("You have unsaved configuration open. Close it before spawning.", {
-                                variant: "warning",
-                                action: snackbarAction,
-                            })
-                            setPanels(p => [...p.filter(x => x !== existing), existing])
-                            return existing.id
-                        }
-                    }
```
Dhruv-0-Arora, line 203: I believe this is just dead code?

fission/src/ui/helpers/UIProviderHelpers.ts
rutmanz: I think it makes much more sense for this to be in `UIScreenProps` so that it's configured by the `configureScreen` function within the panel definition, rather than every time it's opened. It's more an inherent property of the panel and not really something that should depend on who opens it (plus it's quite redundant).

https://github.com/Autodesk/synthesis/blob/53c31112433b039c650cfdd0320a6eac34482185/fission/src/ui/helpers/UIProviderHelpers.ts#L23-L32

rutmanz (changes requested):

AlexD717 (changes requested):
The edit control buttons are now fully blocked from opening (similar story with new input). These should either be fully removed (not advised) or some other solution implemented. 

Image

Dhruv-0-Arora:
@AlexD717 I personally think we should remove it. I can't think of a use case of using the edit feature and also when you click it in dev, it finishes spawning the assembly which I think is extremely confusing especially if you are trying to actively configure the inputs for the assembly (and then it just completes setup while you are still trying to set it up).

fission/src/ui/helpers/UIProviderHelpers.ts
```diff
@@ -45,7 +47,7 @@ export interface ModalProps<P> extends UIScreenProps<P> {
  */
 export interface PanelProps<P> extends UIScreenProps<P> {
     type: "panel"
-    position: PanelPosition
+    position?: PanelPosition
```
Dhruv-0-Arora, line 50: 🤷‍♂️ this is out of scope of this PR but I figured this should be optional

fission/src/ui/UIProvider.tsx
```diff
@@ -146,6 +146,14 @@ export const UIProvider: React.FC<UIProviderProps> = ({ children }) => {
                     return existingDuplicate.id
                 }
             }
+
+            // if any (generic) open panel declares itself as blocking, prevent opening a new one
+            const blockingPanel = panels.find(p => p.props.blocking)
+            if (blockingPanel) {
+                const msg = blockingPanel.props.blockingMessage ?? "Close the current panel before opening another."
+                addToast("warning", msg)
+                return blockingPanel.id
```
PepperLola: I'm not sure this behavior makes sense; `openPanel` is supposed to return the id of the panel that was opened - if no panel was opened (since it was blocked), then shouldn't the function return null? I know this was the behavior before with `existing` but on retrospect I'm not sure it makes much sense. Would have to look at where the result is used and if we depend on this behavior or we expect it to be non-null.

fission/src/ui/panels/configuring/initial-config/InitialConfigPanel.tsx
```diff
@@ -72,22 +71,23 @@ const InitialConfigPanel: React.FC<PanelImplProps<void, void>> = ({ panel }) =>
...
-
         configureScreen(
             panel!,
-            { title: "Assembly Setup", acceptText: "Finish", cancelText: "Remove" },
+            {
+                title: "Assembly Setup",
+                acceptText: "Finish",
+                cancelText: "Remove",
+                blocking: true,
+                blockingMessage: "Finish Assembly Setup before spawning another asset.",
```
rutmanz:
```suggestion
                blockingMessage: "Finish Assembly Setup first!",
```

Other panels aren't necessarily for spawning other assets

fission/src/ui/panels/configuring/initial-config/InputSchemeSelection.tsx
```diff
@@ -90,13 +79,6 @@ export default function InputSchemeSelection({
...
-                        {/** Edit button - same as select but opens the inputs modal */}
-                        {EditButton(() => {
-                            InputSystem.setBrainIndexSchemeMapping(brainIndex, scheme)
-
-                            setSelectedScheme(scheme)
-                            onEdit?.()
-                        })}
 
                         {/** Delete button (only if the scheme is customized and not in use) */}
                         {scheme.customized && status !== InputSchemeUseType.IN_USE ? (
```
rutmanz: if you're deleting the edit button you may as well delete the delete button. I don't think it makes sense to have one "edit input" feature here but not the other. Either leave them both or commit to making them edit inputs elsewhere

fission/src/ui/UIProvider.tsx
rutmanz: Changing the type signature of openPanel to return void doesn't seem to cause any type errors, and I can't find an instance of the panel id being used

rutmanz (changes requested):

fission/src/ui/panels/configuring/initial-config/InputSchemeSelection.tsx
Dhruv-0-Arora: hm ... I think in general, the way that we are handling input scheme edits and changes isn't very intuitive. We should probably rework that.

fission/src/ui/UIProvider.tsx
Dhruv-0-Arora: I think it makes sense to return blockingPanel.id just because if this is ever used at any point, it would make sense to return the currently open panel (the one that is blocking other panels from opening).

rutmanz: I mean I don't really agree with that bc places opening the panel don't have any reason to suspect that the panel is blocked. If I try to open a configuration panel and use the ID that comes back to close it later then I'm actually closing the import panel which isn't really logical. I would change the method signature to just make the returned id optional and return null here. It's not currently used anywhere so there shouldn't be anything else you have to handle

rutmanz (commented):
Other than the ID thing, Lgtm

rutmanz (approved):

AlexD717 (changes requested):
Can the new input button also be removed? I don't think it makes sense to let them make a new input, but then not customize it.

Image

Other than that, looks good to me.

AlexD717 (approved):
lgtm

PepperLola (approved):

PR #1316, opened 2026-06-17
Revert "Hotfix: Chrome Rendering Issue `[AARD-2058]`"

AlexD717:
Reverts Autodesk/synthesis#1262

### DIFF
--- a/fission/src/mirabuf/MirabufInstance.ts
+++ b/fission/src/mirabuf/MirabufInstance.ts
@@ -160,6 +160,13 @@ class MirabufInstance {
      * Creates ThreeJS meshes from the parsed mirabuf file.
      */
     private createMeshes() {
+        this.createBatchedMeshes()
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


fission/src/mirabuf/MirabufInstance.ts
```diff
@@ -160,6 +160,13 @@ class MirabufInstance {
      * Creates ThreeJS meshes from the parsed mirabuf file.
      */
     private createMeshes() {
+        this.createBatchedMeshes()
+    }
+
+    /**
+     * Creates BatchedMesh, more efficient, but broken in newer versions of Chrome
```
rutmanz, line 167: I wouldn't say newer, if you can find specific versions that would be great but otherwise I would just say "some versions of chrome" because it's not current versions (presumably)

rutmanz (commented):
I might leave the old instanced mesh method in just in case but not use it, or leave a comment with this PR or something so if it breaks again it can be found

PR #1320, opened 2026-06-17
Fix Jolt Memory Management `[SYNTH-109]`

azaleacolburn:
## Task

SYNTH-109

## Symptom

Currently, we create a lot of Jolt objects (that are part of the unmanaged WASM heap) and never free them.

This can be seen by creating a lot of objects in fission, watching memory usage increase, then destroying them and watch it remain high.

This causes severe memory usage problems in Fission, leading to performance problems and in severe cases, crashing.

## Solution

- [x] Ensure that every Jolt object is freed with `JOLT.destroy`
- [x] Perform refactors to make memory management easier in the future
  - [x] Extract shared behavior of `ProtectedZoneSceneObject` and `ScoringZoneSceneObject` into `ZoneSceneObject`
  - [x] Functions in `TypeConversion.ts` that consume a Jolt body destroy it conditionally based on a boolean flag argument.
  - [x] Functions in `PhysicsSystems.ts` that set the physical state of a body (e.g. `setBodyPosition`) destroy the Jolt vectors and quaternions they consume based on a boolean flag argument.

The process I went through to fix memory management in this PR was basically to manually borrow check each instance of `/new Jolt/`. However, I am not a machine and almost certainly made a mistake. If a reviewer finds one, I would appreciate it if they let me know.

## Verification

Based on my anecdotal observations, the following are true, however, I haven't run comprehensive standardized benchmarks, so I can't say for sure. If any reviewer does not observe the following to be true in a specific case, I would appreciate it if they let me know.

- [x] Memory usage has decreased overall
- [x] Memory usage decreases when objects are destroyed

---

- [x] User-facing polish:


fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1251,7 +1176,15 @@ class PhysicsSystem extends WorldSystem {
 
         if (!collector.HadHit()) return undefined
```
rutmanz: This early return prevents any of this from being destroyed, also collector is never cleaned up

fission/src/systems/scene/DragModeSystem.ts
```diff
@@ -355,6 +355,8 @@ class DragModeSystem extends WorldSystem {
                     -currentVel.GetY() * stopBrakingStrength,
                     -currentVel.GetZ() * stopBrakingStrength
                 )
+                // TODO
```
rutmanz: are you going to do this?

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
```diff
@@ -177,7 +179,8 @@ class SynthesisBrain extends Brain {
         // Determines which wheels and stimuli belong to which side of the robot
         for (let i = 0; i < wheelDrivers.length; i++) {
             const wheelPos = convertJoltVec3ToJoltRVec3(
-                fixedConstraints[i].GetConstraintToBody1Matrix().GetTranslation()
+                fixedConstraints[i].GetConstraintToBody1Matrix().GetTranslation(),
+                false
             )
 
             const robotCOM = World.physicsSystem
```
rutmanz, line 186: can't comment bc it's too far away but line 189 declares a vector that is never destroyed

rutmanz (commented):
the question about what Jolt does with those objects that you pass to SetLinearVelocity and stuff is probably important to figure out before merging

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Good catch!

fission/src/systems/scene/DragModeSystem.ts
azaleacolburn: After reading the Jolt source code for the `AddForce` function, I'm fairly certain it does not destroy the vector

rutmanz (dismissed):

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
azaleacolburn: According to the source code `GetTranslation()` (and other similar functions) don't return any references to internal objects. However, destroying the returned vector makes tests fail, so I initially assumed an internal reference was returned. I need to investigate this further.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -456,45 +456,54 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+            ;(x.comMesh.material as THREE.Material).dispose()
+        })
+        this._debugBodies?.clear()
+
+        this._physicsLayerReserve?.release()
+
+        this._intakePreferences
+        this._ejectorPreferences
+        this._simConfigData
+        this._fieldPreferences
```
0xda157:
```suggestion
```

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1435,10 +1388,17 @@ class PhysicsSystem extends WorldSystem {
...
+     * @param id The id of the body. Will not be destroyed by this function
+     * @param position The new position of the body. Will be destroyed by this function
      */
-    public setBodyPosition(id: Jolt.BodyID, position: Jolt.RVec3, activate: boolean = true): void {
+    public setBodyPosition(
+        id: Jolt.BodyID,
+        position: Jolt.RVec3,
+        activate: boolean = true,
+        destroy: boolean = true
+    ): void {
```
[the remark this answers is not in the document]
azaleacolburn: The existing call sites were in a test

fission/src/mirabuf/MirabufSceneObject.ts
BrandonPacewic: Duplicate, already called on line 421.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -419,17 +463,17 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
         if (this._centerOfMassIndicator) {
             World.sceneRenderer.scene.remove(this._centerOfMassIndicator)
             this._centerOfMassIndicator = undefined
         }
 
-        if (this._brain && this._brain instanceof SynthesisBrain) {
-            this._brain.clearControls()
-        }
+        World.simulationSystem.unregisterMechanism(this.mechanism)
+        World.physicsSystem.destroyMechanism(this.mechanism)
```
BrandonPacewic: Appears to be duplicate, called on 456 and 457 respectively.

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1232,11 +1172,18 @@ class PhysicsSystem extends WorldSystem {
...
-    public rayCast(from: Jolt.Vec3, dir: Jolt.Vec3, ...ignoreBodies: Jolt.BodyID[]): RayCastHit | undefined {
-        const ray = new JOLT.RRayCast(convertJoltVec3ToJoltRVec3(from), dir)
+    public rayCast(
+        from: Jolt.Vec3,
+        dir: Jolt.Vec3,
+        destroy: boolean = true,
+        ...ignoreBodies: Jolt.BodyID[]
+    ): RayCastHit | undefined {
+        const rayVec = convertJoltVec3ToJoltRVec3(from, destroy)
+        const ray = new JOLT.RRayCast(rayVec, dir)
```
BrandonPacewic, line 1180: `JOLT.RRayCast` copies `rayVec`. Need to destroy it ourselves.

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1232,11 +1172,18 @@ class PhysicsSystem extends WorldSystem {
...
 
         const raySettings = new JOLT.RayCastSettings()
         raySettings.mTreatConvexAsSolid = false
+
         const collector = new JOLT.CastRayClosestHitCollisionCollector()
```
BrandonPacewic, line 1185: `collector` is never destroyed on either return path.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -467,54 +511,82 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
-        })
+
+                if (this._centerOfMassIndicator) {
+                    const inverseMass = body.GetMotionProperties().GetInverseMass()
+
+                    if (inverseMass > 0) {
+                        const mass = 1 / inverseMass
+
+                        const com = body.GetCenterOfMassPosition().Mul(mass)
+                        weightedCOM = weightedCOM.AddRVec3(com)
```
BrandonPacewic, line 541: `.AddRVec3(...)` appears to return a whole new object. Meaning the original `weightedCOM` is leaked.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -467,54 +511,82 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
-            this._centerOfMassIndicator.position.set(netCoM.GetX(), netCoM.GetY(), netCoM.GetZ())
-            this._centerOfMassIndicator.visible = PreferencesSystem.getGlobalPreference("ShowCenterOfMassIndicators")
+            const setPositionAndVisibility = (netCoM: Jolt.RVec3) => {
+                this._centerOfMassIndicator!.position.set(netCoM.GetX(), netCoM.GetY(), netCoM.GetZ())
+                this._centerOfMassIndicator!.visible =
+                    PreferencesSystem.getGlobalPreference("ShowCenterOfMassIndicators")
+            }
+
+            if (totalMass > 0) {
+                const netCoM = weightedCOM.Div(totalMass)
```
BrandonPacewic: `Div` mutates the object in-place. Destroying both `netCoM` and `weightedCOM` results in a double free.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -356,28 +369,42 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
         this.mirabufInstance.parser.rigidNodes.forEach(rn => {
             const jBodyId = this.mechanism.getBodyByNodeId(rn.id)
             if (!jBodyId) return
-            const offset = convertJoltRVec3ToJoltVec3(
-                World.physicsSystem.getBody(jBodyId).GetPosition().Sub(bodyCenter)
-            )
-            const newPos = convertJoltVec3ToJoltRVec3(initialTranslation)
+
+            const position = World.physicsSystem.getBody(jBodyId).GetPosition()
+            const rOffset = position.Sub(bodyCenter)
```
BrandonPacewic: `.Sub` mutates in-place and returns `this`. Results in double free.

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -115,13 +115,16 @@ class PhysicsSystem extends WorldSystem {
         setupCollisionFiltering(joltSettings)
 
         this._joltInterface = new JOLT.JoltInterface(joltSettings)
-        JOLT.destroy(joltSettings)
```
BrandonPacewic, line 118: It does not appear that the new `JoltInterface` object owns the `JoltSettings` struct. Even though it takes a reference it only ends up copying members of the struct. Meaning that the struct itself would still need to be destroyed.

BrandonPacewic (approved):

AlexD717:
Very goofy bug... I can't seem reproduce it on dev though

1) Spawn in 2025 field
2) Enable drag mode
3) Drag one of the balls inside the corral. Dragging some balls will result in this error. The error appears most common when you try to drag the ball into one of the branches

Screenshot 2026-06-23 at 4 13 00 PM

4) Spawn a previously locally cached field (only tested with 2023 and 2019 field). I had it happen once where this was fine, but in the majority of cases this seems to result in an error.
5) Try to orbit around the map (should fail)
6) Observe a memory out of bounds error

Screenshot 2026-06-23 at 4 18 07 PM


7) App sometimes becomes fully unresponsive requiring a rerun of `bun run dev` to fix (Sometimes orbit continues to not work but I can open panels)


Video example of the bug

https://github.com/user-attachments/assets/cca3d014-0ed1-4c56-a593-782c6c5af515

Video example of what happens on dev


https://github.com/user-attachments/assets/d1a705de-fc9e-46bf-a946-4076ab192167

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1571,6 +1603,7 @@ class PhysicsSystem extends WorldSystem {
         const contactListener = new JOLT.ContactListenerJS()
 
         contactListener.OnContactAdded = (bodyPtr1, bodyPtr2, manifoldPtr, settingsPtr) => {
+            console.log("contact")
```
AlexD717:
```suggestion
```

Creates a lot of unnecessary logs

fission/src/mirabuf/ZoneSceneObject.ts
```diff
@@ -0,0 +1,159 @@
...
+        World.physicsSystem.setShape(bodyId, shape.Get(), false, Jolt.EActivation_Activate)
+
+        JOLT.destroy(boundingVec)
+        JOLT.destroy(shapeSettings)
+        JOLT.destroy(shape)
+    }
+
+    // Creates a mesh for the user to visualize the sensor
+    private createVisualMesh(props: VisualProperties) {
+        const unitVector = new JOLT.Vec3()
```
AlexD717:
```suggestion
        const unitVector = new JOLT.Vec3(1, 1, 1)
```

Currently the protected/scoring zone meshes don't properly render

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -741,20 +755,24 @@ class PhysicsSystem extends WorldSystem {
             })
         }
 
+        // TODO for azalea
+        // There is 100% some memory bug with the creation of spare ghost bodies
+        // However, this looks scary, so I'll touch it in another commit
```
AlexD717: Was a ticket made for this?

AlexD717 (changes requested):
A couple very minor changes

fission/src/mirabuf/ScoringZoneSceneObject.ts
```diff
@@ -57,143 +19,66 @@ class ScoringZoneSceneObject extends SceneObject {
...
     public constructor(parentAssembly: MirabufSceneObject, index: number, render?: boolean) {
-        super()
+        super(parentAssembly, parentAssembly.fieldPreferences?.scoringZones[index]!, "RenderScoringZones", render)
 
-        this._parentAssembly = parentAssembly
-        this._prefs = this._parentAssembly.fieldPreferences?.scoringZones[index]
-        if (this._prefs && "persistentPoints" in this._prefs) {
-            this._prefs.shouldPointsAccumulate = !this._prefs.persistentPoints
-            delete this._prefs.persistentPoints
-        }
```
AlexD717, line 67: Removing this results in errors when spawning older fields (like the default 2023 field), and then trying to toggle accumulated points. This can either be fixed by
- Updating the default .mira files for this field
- Keeping this code
- Adding a default value to accumulating points on `ScoringZoneConfigInterface.tsx`. Ex replace line 21 with (`const [accumulating, setAccumulating] = useState<boolean(selectedZone.shouldPointsAccumulate ?? false)`)

I will let you decide what you think it best.

azaleacolburn: Oh big thinking. The third one is definitely wrong because we want to keep the old `persistentPoints` values (even though it's the cleanest option). Option one is probably correct, but way more annoying. I honestly don't remember why I removed this code, so I will return it.

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Done!

fission/src/mirabuf/ZoneSceneObject.ts
azaleacolburn: Oh this is a crazy oversight on my part tysm

AlexD717 (approved):
lgtm

PR #1321, opened 2026-06-18
Close ui views on Escape keypress `[SYNTH-135]`

rutmanz:
## Task

SYNTH-135

## Symptom

Pressing escape does not close anything except for the command palette

## Solution

Pressing escape closes the command palette, open modal, or most recently opened panel (in that priority order) if they are open and have the back/cancel button enabled.

## Verification

- Pressing escape with multiple panels/modals/palettes only closes the frontmost one
- If there is no back/cancel button (like the scoreboard), escape does nothing 
- Focusing outside of the browser and coming back, escape behaves as you would expect

---

- [X] User-facing polish:


fission/src/systems/input/InputSystem.ts
```diff
@@ -119,10 +122,26 @@ class InputSystem extends WorldSystem {
     /** Called when any key is first pressed */
     private handleKeyDown(event: KeyboardEvent) {
         InputSystem._keysPressed[event.code as KeyCode] = true
+
+        if (event.key == "Escape") {
+            console.log("Escape pressed!")
```
Dhruv-0-Arora: wut 😶‍🌫️

fission/src/systems/input/InputSystem.ts
```diff
@@ -41,6 +41,9 @@ class InputSystem extends WorldSystem {
         })
     }
 
+    // Janky solution to centralize escape key closing logic, first in the list is higher priority, returning true consumes the keypress
+    public static escapeKeyListeners: (null | (() => boolean))[] = [null, null, null]
```
Dhruv-0-Arora, line 45: While technically this is a better solution imo than just making a useEffect with
```ts
if (InputSystem.isKeyPressed("Escape")) {

}
```
i don't know if it is a good idea to make this whole implementation of a priority queue (or not queue) system because there aren't that many things that really need escape functionality in such a way that there will be a queue of things waiting on it.

fission/src/systems/input/InputSystem.ts
```diff
@@ -119,10 +122,26 @@ class InputSystem extends WorldSystem {
...
     /* Called when any key is released */
     private handleKeyUp(event: KeyboardEvent) {
+        if (event.key == "Escape" && InputSystem._keysPressed["Escape"] == false) {
+            // Sometimes focus issues prevent the keydown from being fired, but the keyup is still sent.
+            const anyMatched = InputSystem.escapeKeyListeners.some(cb => cb != null && cb())
+            if (anyMatched) {
+                event.preventDefault()
+            }
+        }
+
```
Dhruv-0-Arora: im not a huge fan of adding code to handleKeyUp()
if you REALLY REALLY need to I personally would be able to sleep a lot better at night if it was in another function that was called from here.

Dhruv-0-Arora: same thing for handleKeyDown

also, why the mismatch between the location of 
```ts
InputSystem._keysPressed[.... 
``` 
in the two different functions? its above the block of code added in one and not in the other

fission/src/ui/components/CommandPalette.tsx
```diff
@@ -125,16 +133,11 @@ const CommandPalette: React.FC = () => {
...
+                e.preventDefault()
                 if (!World.isAlive) return
                 if (isMainMenuOpen) return
                 if (modal) return
-                e.preventDefault()
                 openPalette()
-            } else if (e.key === "Escape") {
-                if (isOpen) {
-                    e.preventDefault()
-                    closePalette()
```
Dhruv-0-Arora, line 136: the command Palette should not have its escape close functionality in InputSystem ...

especially in onKeyDown there should not be a 
```ts
if (e.key === "/") {
```

Dhruv-0-Arora: This has no relevance to this PR really though

fission/src/ui/components/CommandPalette.tsx
```diff
@@ -195,12 +198,9 @@ const CommandPalette: React.FC = () => {
             } else if (e.key === "Enter") {
                 e.preventDefault()
                 execute(activeIndex)
-            } else if (e.key === "Escape") {
-                e.preventDefault()
-                closePalette()
             }
         },
-        [activeIndex, execute, visible.length, closePalette]
+        [activeIndex, execute, visible.length]
```
Dhruv-0-Arora, line 203: why isn't openPalette in here
Edit: I didn't read properly

Dhruv-0-Arora (changes requested):
Please disagree with me 😁

fission/src/systems/input/InputSystem.ts
rutmanz: No it needs to be centralized to prevent more than one thing from being closed with each keypress

AlexD717:
What are your thoughts of closing the most recently dragged/interacted with panel rather than the latest opened?

rutmanz:
The window layering (which window is on top) is already determined by opening order, I think that if we were to change the escape order to most-recently-interacted it would also make sense to change the z-indexes based on that but maybe a separate ticket

AlexD717:
In my opinion, it makes sense to have the most recently interacted with panel be on top, but I agree that it should be a separate ticket.

AlexD717 (approved):

fission/src/systems/input/InputSystem.ts
```diff
@@ -119,13 +122,28 @@ class InputSystem extends WorldSystem {
...
         InputSystem._keysPressed[event.code as KeyCode] = true
+        this.checkEscapeKey(event)
     }
 
     /* Called when any key is released */
     private handleKeyUp(event: KeyboardEvent) {
+        if (InputSystem._keysPressed["Escape"] == false) {
+            // Sometimes focus issues prevent the keydown from being fired, but the keyup is still sent.
+            this.checkEscapeKey(event)
+        }
```
Dhruv-0-Arora, line 133: yippee functions are fantastic

Dhruv-0-Arora (approved):
👌🏽

PepperLola (approved):

PR #1322, opened 2026-06-18
Fix Alliance Cancellation in Configure Panel `[SYNTH-88]`

RoushilS:
## TASK
SYNTH-88

## Symptom

After opening the configure panel and changing the Alliance or Station and pressing cancel, neither reverts to its original state. 

## Solution

Fixed by creating `originalAlliance` and `originalStation` variables and adding refs to them. Then add `onCancel` behavior to revert.

## Verification

- [ ] Alliance/Station can be switched correctly
- [ ] Alliance/Station can be saved correctly
- [ ] Alliance/Station can be canceled correctly

- [x] User-facing polish:


fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
```diff
@@ -302,6 +320,9 @@ const ConfigurePanel: React.FC<PanelImplProps<void, ConfigurePanelCustomProps>>
             originalFieldPrefs.current = null
             originalMotorPrefs.current = null
             originalInputSchemes.current = null
+
+            originalAlliance.current = undefined
+            originalStation.current = undefined
```
rutmanz, line 325: For consistency with the existing fields, I think it would make more sense to use `null` instead of `undefined`. `null` also generally has the semantic meaning of "intentionally left blank"

[Image](https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript)

rutmanz (changes requested):
Image

Looks like it doesn't revert the nametag color when it reverts the alliance. Otherwise looks great

fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
RoushilS: I agree. However, the reason I didn't do that is because in MirabufSceneObject, it resolves to undefined, not null, and I didn't want to mess with it. I can change it if it's important though

RoushilS: It reffers to alliance/station

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -224,10 +224,18 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+        }
     }
 
     public set station(station: Station | undefined) {
         this._station = station
+        if (this._nameTag) {
+            this.updateNameTag()
+            EventSystem.dispatch("SceneOverlayUpdateEvent")
+        }
     }
```
rutmanz:
```suggestion
    public set alliance(alliance: Alliance | undefined) {
        this._alliance = alliance
        this.updateNameTag()
    }

    public set station(station: Station | undefined) {
        this._station = station
    }
```

rutmanz: updateNameTag already handles the null check and updates the alliance, and changing the station doesn't affect the nametag color/position. SceneOverlayUpdateEvent is fired every frame anyway if nametags are visible, so no need to fire it again

fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
rutmanz: Yeah okay I see what you mean

rutmanz (changes requested):
seems to work great, just this little cleanup and should be good

fission/src/mirabuf/MirabufSceneObject.ts
RoushilS: Oh ok. Thanks!

rutmanz (approved):

AlexD717 (approved):
lgtm. Minor merge conflict with dev

PepperLola (approved):
Looks good to me

PR #1323, opened 2026-06-18
Disable Low Performance Graphics Alert on lowest setting `[SYNTH-116]`

RoushilS:
## TASK
SYNTH-116

## Symptom

If you are experiencing low performance, an alert appears and the graphics panel prompts you to switch to default graphics, even if you already are on the default.

## Solution

Created check method to see if user is on default graphics, then implemented a boolean check to not trigger low performance protocol if this is the case.

## Verification

A low performance environment can be induced by going to the performance tab on Devtools, and throttling CPU.

- [ ] Low Performance alert does not appear when user is already on default graphics
- [ ] Low Performance alert works as intended if not on default graphics

- [x] User-facing polish:


PepperLola (commented):
Are low graphics settings the default? The idea behind this PR is to not pointlessly prompt users to lower their graphics settings if there are no settings lower than the current one. The current code makes sense if the default is the lowest.

fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -222,6 +222,19 @@ class PreferencesSystem {
...
+    public static graphicsPreferencesAreDefault(): boolean {
+        const current = this._preferences[GRAPHICS_PREFERENCE_KEY]
+        const defaults = defaultGraphicsPreferences()
+        return (
+            current!.lightIntensity === defaults.lightIntensity &&
+            current!.fancyShadows === defaults.fancyShadows &&
+            current!.maxFar === defaults.maxFar &&
+            current!.cascades === defaults.cascades &&
+            current!.shadowMapSize === defaults.shadowMapSize &&
+            current!.antiAliasing === defaults.antiAliasing
```
rutmanz: I would actually just check that fancy shadows and antialiasing are off, because max far and cascades and map size don't matter if that's the case. And I don't think light intensity affects performance

rutmanz (changes requested):

Dhruv-0-Arora:
Wait, why are we trying to not alert when the user is on default graphics settings? Is it because that is already the lowest settings or something?

RoushilS:
> Wait, why are we trying to not alert when the user is on default graphics settings? Is it because that is already the lowest settings or something?

> Are low graphics settings the default? The idea behind this PR is to not pointlessly prompt users to lower their graphics settings if there are no settings lower than the current one. The current code makes sense if the default is the lowest.

@PepperLola @Dhruv-0-Arora

You are both correct. As far as SYNTH-91 (#1325) goes, I had a couple misunderstandings about performance at different levels, and will remove some unnecessary features in that PR. 

So yes, default is the lowest setting.

fission/src/systems/preferences/PreferencesSystem.ts
RoushilS: fixed in latest commit

RoushilS:
I have fixed the other PR so this one is ready now.

fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -222,6 +222,11 @@ class PreferencesSystem {
         this.savePreferences()
     }
 
+    public static graphicsPreferencesAreLow(): boolean {
+        const current = this._preferences[GRAPHICS_PREFERENCE_KEY]
+        return current!.fancyShadows === false && current!.antiAliasing === false
```
AlexD717:
```suggestion
        const current = this.getGraphicsPreferences()
        return current.fancyShadows === false && current.antiAliasing === false
```

While this is okay, you should use the dedicated getGraphicsPreferences function. This also grantees a GraphicsPreferences return so you don't have to use the null assertion operators.

AlexD717 (changes requested):

fission/src/systems/preferences/PreferencesSystem.ts
rutmanz: Agree with that, and we should almost always avoid using the null assertion (`current!.fancyShadows`) and instead use optional chaining (`current?.fancyShadows`) to avoid potentially causing runtime errors

rutmanz (approved):

AlexD717 (approved):

PR #1325, opened 2026-06-18
Create Graphics Settings Presets `[SYNTH-91]`

RoushilS:
## TASK
SYNTH-91

## Symptom

If the user wants to configure the simulation quality and performance to be optimized on their device, they have no quick way to do so. 

## Solution

Test various performance settings at 5 levels. Create a dropdown UI that allows clean preset selection. Also implement save and cancel behavior. Add useEffect to StatefulSlider to automatically update when base variable changes, as there was previously a disconnect when a user selected the preset but it didn't update. Add reusable setGraphicsPreferences method to make code more concise.

Additionally, reviewers ensuring that these actually are the best graphics presets for the simulator would be very much appreciated.

## Verification

- [ ] Graphics presets are useful on various performance levels. 
- [ ] UI correctly changes graphics levels
- [ ] UI is intuitive and functional.
- [ ] Various user edge cases (cancel, switching but not saving, opening when not on default setting, etc.) all work.

- [x] User-facing polish:


fission/src/systems/preferences/PreferenceTypes.ts
```diff
@@ -92,6 +92,61 @@ export function defaultGraphicsPreferences(): GraphicsPreferences {
...
+
+export function lowGraphicsPreferences(): GraphicsPreferences {
+    return {
+        lightIntensity: 3,
+        fancyShadows: false,
+        maxFar: 20,
+        cascades: 2,
+        shadowMapSize: 2048,
+        antiAliasing: false,
+    }
```
rutmanz: I could be wrong but I'm pretty sure light intensity doesn't change performance, it's just like a brightness setting. And (assuming the UI is done correctly) the shadow-related properties are ignored when fancy shadows are off, making low graphics just like medium graphics performance-wise but darker. Maybe remove low and shift the other ones so medium is low, high is medium (default), etc?

rutmanz (changes requested):
Image
I think it's a little unintuitive to have the label stay when you manually change settings off of the preset, maybe change it to say "custom" or just be blank when other settings are modified. This also has the odd consequence of any modified graphics settings showing up as "medium" when you reload.

fission/src/systems/preferences/PreferenceTypes.ts
Dhruv-0-Arora: Also another note, there are differences in maxFar, cascades, shadowMapSize - all of which actually do not make a difference when fancyShadows is set to false. So, there is pretty much no quality difference between the two

fission/src/ui/components/StatefulSlider.tsx
```diff
@@ -12,6 +12,11 @@ const StatefulSlider: React.FC<
     }
 > = props => {
     const [value, setValue] = useState(props.defaultValue)
+
+    useEffect(() => {
+        setValue(props.defaultValue)
+    }, [props.defaultValue])
```
Dhruv-0-Arora, line 19: I would add a comment here explaining why it was added

```suggestion
    // when setting the values with a dropdown menu the sliders didn't update
    useEffect(() => {
        setValue(props.defaultValue)
    }, [props.defaultValue])
```

Dhruv-0-Arora: or you can put the comment inside the useEffect as well however you would like to do it

fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -218,6 +228,43 @@ const GraphicsTab: React.FC<GraphicsTabProps> = ({ onActionsChange }) => {
...
+            setSelectedPreset("low")
+        } else if (presetMatches(mediumGraphicsPreferences())) {
+            setSelectedPreset("medium")
+        } else if (presetMatches(highGraphicsPreferences())) {
+            setSelectedPreset("high")
+        } else if (presetMatches(veryHighGraphicsPreferences())) {
+            setSelectedPreset("veryHigh")
+        } else if (presetMatches(ultraGraphicsPreferences())) {
+            setSelectedPreset("ultra")
+        }
```
Dhruv-0-Arora: this seems like a very inefficient way to implement this. There is a lot of code being added. Less code is better  :)

Maybe using an iterable dictionary could be an idea:
```ts
const GRAPHICS_PRESETS: Record<string, GraphicsPreferences> = {
    low: lowGraphicsPreferences(),
    medium: mediumGraphicsPreferences(),
    high: highGraphicsPreferences(),
    veryHigh: veryHighGraphicsPreferences(),
    ultra: ultraGraphicsPreferences(),
};
```

and use preset match like

```ts
const isPresetMatch = (current: GraphicsPreferences, preset: GraphicsPreferences) => {
    const keys = Object.keys(preset) as Array<keyof GraphicsPreferences>;
    return keys.every((key) => current[key] === preset[key]);
};
```


Also the repetition of code from above can signal maybe you could turn that functionality into a function

Dhruv-0-Arora (changes requested):

RoushilS:
Because of my newfound knowledge, and the reviews, I pushed some changes and tested everything again. I think I found and patched all of the bugs but there could be more.

rutmanz (approved):
lgtm

rutmanz:
might be good to wait for #1323 so that the [`graphicsPreferencesAreLow()`](https://github.com/Autodesk/synthesis/pull/1323/changes#diff-d813fba5df7b2e924d33d03842415122005de6982b4b2f7db460605fa5a3aeddR225) method can be reused. Alternatively implement it again here and then just deduplicate in the merge conflict resolution

AlexD717 (approved):
lgtm. I like how the light intensity doesn't affect graphics settings.

fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -218,6 +228,63 @@ const GraphicsTab: React.FC<GraphicsTabProps> = ({ onActionsChange }) => {
...
+        return (
+            a.fancyShadows === b.fancyShadows &&
+            a.maxFar === b.maxFar &&
+            a.cascades === b.cascades &&
+            a.shadowMapSize === b.shadowMapSize &&
+            a.antiAliasing === b.antiAliasing
+        )
+    }
+
+    const getPresetForPreferences = (prefs: GraphicsPreferences): GraphicsPreset => {
```
Dhruv-0-Arora: Maybe a different name?
```suggestion
    const getGraphicsPreset = (prefs: GraphicsPreferences): GraphicsPreset => {
```

Dhruv-0-Arora: I'm not entirely sure why we need to write a function that classifies what preset is currently selected when there already is 
```ts
    const [selectedPreset, setSelectedPreset] = useState<GraphicsPreset>("custom")
```

Dhruv-0-Arora: In other words, can't we just rely more on the selectedPreset? - also maybe change the name of that variable to selectedGraphicsPreset? Just an opinion though.

fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -218,6 +228,63 @@ const GraphicsTab: React.FC<GraphicsTabProps> = ({ onActionsChange }) => {
...
+        setShadowMapSize(prefs.shadowMapSize)
+        setAntiAliasing(prefs.antiAliasing)
+        World.sceneRenderer.changeLighting(prefs.fancyShadows)
+        setSelectedPreset(getPresetForPreferences(prefs))
+    }
+
+    useEffect(() => {
+        const current = PreferencesSystem.getGraphicsPreferences()
+        setSelectedPreset(getPresetForPreferences(current))
+    }, [])
```
Dhruv-0-Arora: hmm so we are getting the preferences from local storage and then deciding whether they match up with a particular preset? I think a better way of doing this would be to just store the type itself into local storage (using the PreferenceSystem) that way we can get rid of the getPresetForPreferences function entirely.

Dhruv-0-Arora (changes requested):

fission/src/ui/modals/configuring/SettingsModal.tsx
RoushilS: Selected preset is the dropdown control, meaning that it relies on the local data and is not a source of truth. I made the name change though.

fission/src/ui/modals/configuring/SettingsModal.tsx
RoushilS: In my opinion `getGraphicsPreset` (renamed from `getPresetforPreferences`) function is a bit cleaner than the alternate solution.
It avoids creating a `GraphicsPreset` field in each graphics preset.
A constant check has better prevention against edge cases, as it is less likely to go out of sync compared to an operation that is only performed on a change.
It's possible that I'm misunderstanding what you're saying though.

Dhruv-0-Arora (approved):

PepperLola (approved):

PR #1326, opened 2026-06-18
Camera Fixes & Face Mode `[SYNTH-67] & [SYNTH-90] & [SYNTH-119]`

AlexD717:
## Task

SYNTH-67 & SYNTH-90 & SYNTH-119

### Camera Jitters When Rotating Robot
SYNTH-67

When rotating some robots (like Dozer), with the camera focused on it, the camera jolts and jitters around unexpectedly.

### Camera Lock Mode is Broken
SYNTH-90

Lock mode is currently broken, and switching it on changes nothing.

### Bystander Camera
SYNTH-119

Create a camera mode that allows the user to position themself somewhere in the field, and then turn to look at a selected robot.

## Solution

- Have the camera follow the center of mass of robots instead of the physical center, fixing the camera jitter as robots rotate around the center of mass (Fixes SYNTH-67)
- Apply the robot rotation to the focus matrix, fixing lock mode (SYNTH-90)
- Implement a new face mode for the camera, locking its position and rotating to always face the robot (SYNTH-119)
- Add a camera config button to the MainHUD to open the previously unused Camera Config Panel
   - Add a command to allow direct opening of the config panel

## Verification

- [x] Spawning Dozer and rotating it no longer jitters the camera
- [x] Switching to follow mode (the default) results in the camera following the robot position (not rotation)
- [x] Switching to lock mode results in the camera following the robots position and rotation
- [x] Switching to face mode results in the camera position becoming fixed with the camera rotating to face the focused robot
- [x] Right click on the robot allows for camera mode switching
- [x] Camera config panel allows for camera mode switching
- [x] Switching from between camera modes doesn't result in random camera jumps
- [x] Updating camera mode in the MainHUD also updates the Camera Config Panel
- [x] Drag mode camera interaction is smooth

---

- [x] User-facing polish:


RoushilS (commented):
I'm wondering if it might it be better to keep the camera config panel within the settings panel. I think it would make the mainHUD cleaner.

AlexD717:
> I'm wondering if it might it be better to keep the camera config panel within the settings panel. I think it would make the mainHUD cleaner.

IMO it makes more sense to keep it separate that way you get to immediately see your changes, while the settings panel fully blocks the screen. However, if most people think it makes more sense to move it to settings I can do that.

RoushilS (commented):
For some reason when I focus the camera on the field, it brings up robot options that don't do anything when I click on them. Recording attatched.
https://github.com/user-attachments/assets/71cd4796-cf1b-45cc-864b-ef9146e16a33

AlexD717:
> For some reason when I focus the camera on the field, it brings up robot options that don't do anything when I click on them. Recording attatched. https://github.com/user-attachments/assets/71cd4796-cf1b-45cc-864b-ef9146e16a33

Disabled the ability to switch camera modes if you are focused on a field.

RoushilS:
I think you can still change them in the camera config panel.

RoushilS (changes requested):

rutmanz:
There seems to be no way to move at all when in face mode (maybe just zooming towards/away from focus object with scrolling would be good), and I'm not sure what the point of being able to focus on a field really is, compared to just unfocusing the robot

AlexD717:
> There seems to be no way to move at all when in face mode (maybe just zooming towards/away from focus object with scrolling would be good),

I can add scrolling, but panning isn't something that exist on any camera mode at all (that should be a different ticket)

> and I'm not sure what the point of being able to focus on a field really is, compared to just unfocusing the robot

One advantage could be centering the camera on the center of the field. If you unfocus on the robot you are just left in place.

rutmanz:
image
I'm not sure what "orbit" is for, but maybe replace it with a way of changing the focused entity directly from the camera control menu? I think the flow for context menu focus -> main menu -> camera control panel is kind of... unintuitive and if I were to try to focus on something specific I would want to be able to do it all from the camera control panel

AlexD717:
> I'm not sure what "orbit" is for

Orbit is the only current CameraControlsType. I didn't really want to touch that section, I think Hunters idea was to have multiple camera control types. Ex, one for locking your position into on of the driver station views, no longer allowing you to orbit but allow slight panning.

> maybe replace it with a way of changing the focused entity directly from the camera control menu? I think the flow for context menu focus -> main menu -> camera control panel is kind of... unintuitive and if I were to try to focus on something specific I would want to be able to do it all from the camera control panel

Changing focus from the camera menu is a great idea. I think it should just be an extra dropdown, rather than something replacing the orbit selection. In addition, I think that should be a separate ticket.

rutmanz:
I mean I get the desire to leave the camera control type stuff there, but I think that if this PR is reintroducing that panel it should probably make sure that panel makes sense / doesn't have random unused stuff in it. A lot of the UX problems in Synthesis come from leaving things in when they aren't fully implemented/used IMO.

RoushilS:
if DS view will be implemented soon after this goes in I think it's okay to keep.

RoushilS:
When I switch between the follow and locked settings in the configure camera panel, and am focused on Dozer, the view cube changes orientation. This does not happen when focused on a field, or none. This doesn't seem to cause any bugs, and doesn't seem to be a big deal but I thought I should let you know.

RoushilS:
Also when I'm focused on None I can still switch between all three of the views.

RoushilS (approved):
Looks good!

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -515,14 +515,16 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
-                new THREE.Vector3(
-                    (boundingBox.max.x + boundingBox.min.x) / 2,
-                    boundingBox.max.y + 0.1,
-                    (boundingBox.max.z + boundingBox.min.z) / 2
+
+            const rootNodeId = this.getRootNodeId()
+            if (rootNodeId) {
+                const centerOfMass = World.physicsSystem.getBody(rootNodeId).GetCenterOfMassPosition()
+                this._nameTag.position = World.sceneRenderer.worldToPixelSpace(
+                    new THREE.Vector3(centerOfMass.GetX(), boundingBox.max.y + 0.1, centerOfMass.GetZ())
```
PepperLola: Is this behavior we want? I feel like for a robot with a center of gravity that doesn't lie in the center of its bounding box it would make more sense for the name tag to be above the center of the bounding box rather than off to the side above its center of gravity.

PepperLola: Same with the camera focus - is there a reason you chose that instead of the center of the bounding box?

fission/src/ui/panels/configuring/CameraSelectionPanel.tsx
```diff
@@ -1,69 +1,147 @@
...
+                    <MenuItem key={t.id} value={t.id}>
+                        {t.assemblyName}
+                    </MenuItem>
+                ))}
+            </Select>
+        </div>
+    )
+}
+
 const OrbitSettings: React.FC<OrbitSettingsProps> = ({ controls }) => {
```
PepperLola: Now that we have multiple settings here I think a different name would make more sense? Since not all the options under OrbitSettings necessarily orbit the robot.

fission/src/mirabuf/MirabufSceneObject.ts
AlexD717: Robots rotate around the center of mass. If we focus on their bounding box center which is slightly different from the center of mass, it causes the camera to jump when rotating rapidly. Same with the nametag, instead of being centered it jumps around.

fission/src/ui/panels/configuring/CameraSelectionPanel.tsx
AlexD717: What name would you suggest? I'm assuming that apart from orbit we would have driver station view controls.

RoushilS: Trail maybe

AlexD717: Renamed to "Target" where possible

RoushilS (approved):

rutmanz (changes requested):
Ok I'm a fan of this panel, I think the one remaining thing would just be to add hover text to each of these modes explaining what they are (and how follow and locked are different for fields, or disable one of them).
<details>
image
</details>

Also what @RoushilS pointed out about focusing on nothing
<details>
image
</details>

AlexD717:
@rutmanz, I don't see the issue with being able to change the view while focused on none (if that's the issue you meant). That just means that when you do switch to a focus target you are already in the mode you want to be. I think it would be more frustrating if you are focused on a robot in say face mode, you then switch to none for a while as you just want a fixed position, and then when you switch back to a robot you are in follow mode rather than the previous face mode.

RoushilS:
I think he meant that it gets locked to the mode you were on when switching to it, so that you can't edit it but it still remembers the mode. Both that suggestion and how it is right now would be ok in my opinion.

rutmanz:
What's the difference between Follow and Locked for a field?

AlexD717:
@rutmanz there is no difference. I allow them so that if you switch from robot -> field, then from field -> robot your selected mode is kept.

rutmanz:
But then why is face disabled for fields? I think it would be fine to just prevent changing the mode while not focused on a robot (like just hide or completely disable the mode selector) and only make the view modes apply to robots. As a user I would be thrown off by having a selector active that doesn't change anything

AlexD717:
Face mode is disabled as it locks the camera position, which prevents rotating around the camera center point. I will disable the mode selector if focused on a field.

rutmanz:
Sorry I know this back and forth feels kind of nit-picky but I think the mode changer should also be disabled when "None" is selected for focus. 

I don't think this matters as much but I would also say I don't think you need any kind of "not supported" text when the focus selector is hidden. To me it kind of implies that it might be supported at some point but the decision to not have that is because focus modes aren't a thing for fields, not just because we haven't gotten to it yet. I don't think people would be wondering "why can't I change the focus mode" when they're not focused on something that moves. Up to you though

AlexD717:
No worries, I appreciate all your feedback. Making an intuitive user experience is very important for Synthesis.

rutmanz (approved):

PR #1327, opened 2026-06-18
Rename persistent points `[SYNTH-100]`

rutmanz:
## Task

SYNTH-100

## Symptom

"Persistent points" is kind of confusing and it's not clear that it means "gamepieces must persist for points to count"

## Solution

Renamed the field to "shouldPointsAccumulate" and flipped it, so now it represents if points should still count after the gamepieces leave

## Verification

- Gameplay still works as expected
- no type errors
- loading 2023 field has accumulating points set to false (migration works)

---

- [X] User-facing polish:


AlexD717 (approved):
lgtm. I think the biggest improved is the addition of the tooltip, the name itself isn't that much cleaner (but I can't think of anything else).

RoushilS:
I agree with Alex. Namewise, I think that `shouldPointsPersist` is more clear. If we're being really descriptive, `retainPointsOnExit` could work. Looks good though.

RoushilS (approved):

azaleacolburn (approved):

PR #1331, opened 2026-06-19
Update dependencies `[SYNTH-145]`

rutmanz:
## Task

SYNTH-145

## Symptom

Security audit is angry and we're using a bunch of outdated libraries because they haven't been updated since last summer.

## Solution

Updated deps and used overrides for packages that depend on insecure versions of other packages. Then had to spend a bunch of time identifying and fixing code that fell victim to breaking changes.

> [!IMPORTANT]
Everything is currently failing due to something with the mui packaging and I don't know why. It looks like there is [an open issue about this](https://github.com/vitejs/vite/issues/22499) on vite but I downgraded the vite version and it didn't get fixed. Will continue to investigate.

## Verification

---

- [ ] User-facing polish:


0xda157:
Maybe we adopt dependabot or similar so we aren't upgrading everything at once (which is very likely to cause regressions)?

PR #1338, opened 2026-06-23
Fix Inconsistent "Upload Valid MatchModeConfig" Test `[SYNTH-155]`

AlexD717:
## Task

SYNTH-155

## Symptom

Sometimes (particularly in the Github CI), the testing MatchModeConfigPanel > Upload Valid MatchModeConfig will fail (see below). This happens seemingly arbitrarily on PRs that are completely unrelated to MatchMode. In most case rerunning the test fixes the issue.

Here is one example of this happening: https://github.com/Autodesk/synthesis/actions/runs/27793880614/job/82249165846 
Here is another: https://github.com/Autodesk/synthesis/actions/runs/28036284575/job/82990405745 

## Solution

The test previously waited for a set amount of time and then checked if the number of configs increased. On the Github CI updating the number of configs sometimes took to long, resulting in a test fail. Now for ValidConfig tests it waits for the number of configs to update.

## Verification

- [x] Unit tests pass
- [x] Match Mode Stress Test (run the test 50 times) passed https://github.com/Autodesk/synthesis/actions/runs/28047649691/job/83029764417?pr=1338

---

- [x] User-facing polish:


fission/src/test/ui/MatchModeConfigPanel.test.tsx
```diff
@@ -87,16 +87,19 @@ describe("MatchModeConfigPanel", () => {
...
-            )
+            await waitFor(() => {
+                const finalCount = getMatchModeCount(container)
+                assert(
+                    finalCount === initialCount + 1,
+                    `Expected count to increase from ${initialCount} to ${initialCount + 1}, but got ${finalCount}`
+                )
+            })
         } else {
+            await new Promise(resolve => setTimeout(resolve, 100))
```
PepperLola: Is this just to make sure that all the event's effects occur before we run the test to make sure it really didn't store another config?

AlexD717: The test behavior works as follows. If we are testing a valid config, the match mode count should increase. We wait, repeatedly checking if has increased (until the default timeout, or it increases). If its a invalid config, we need to verify that the number hasn't increase. Because of this we just wait for the set amount of time to make sure it hasn't increased.

rutmanz: It sounds like this is leaving open a potential false pass then, if we've had issues where the config isn't uploaded within 100ms, then an invalid config could be imported successfully (incorrectly) and still have the test pass. Not sure if it's worth worrying about though

rutmanz (approved):

AlexD717:
> It sounds like this is leaving open a potential false pass then, if we've had issues where the config isn't uploaded within 100ms, then an invalid config could be imported successfully (incorrectly) and still have the test pass. Not sure if it's worth worrying about though

You're right. Since both valid and invalid configs result in a toast popup at the end, one solution could be to wait until gloablAddToast is called, and then check if the count has increased (what are your thoughts on this @rutmanz).

rutmanz:
> > It sounds like this is leaving open a potential false pass then, if we've had issues where the config isn't uploaded within 100ms, then an invalid config could be imported successfully (incorrectly) and still have the test pass. Not sure if it's worth worrying about though
> 
> You're right. Since both valid and invalid configs result in a toast popup at the end, one solution could be to wait until gloablAddToast is called, and then check if the count has increased (what are your thoughts on this @rutmanz).

It feels kind of sketchy to have the test case depend on a behavior that is kind of disconnected (like it would work but it feels like the same as using console logs as a signal for tests). If there something else you can spy on (like whatever method handles the uploaded file) that would be better I think but if not this works

rutmanz (approved):
nice!

PepperLola (approved):

PR #1339, opened 2026-06-23
Document Jolt Ownership Model `[SYNTH-156]`

azaleacolburn:
## Task

SYNTH-156

## Symptom

Currently, in order to correctly use Jolt Physics functions, users have to reference the Jolt source code to determine whether the function in question consumes or copies any references it is given and whether it returns any internal references to its state or returns a copy.

## Solution

I used Claude Code for this task, as it was basically a direct translation from the JoltJS `.idl` files. The prompt I used:

'''
Currently, in order to correctly use Jolt Physics functions, users have to reference the Jolt source code to determine whether the function in question consumes or copies any references it is given and whether it returns any internal references to its state or returns a copy.

The fact that this is undocumented makes the correct use of Jolt memory very difficult.

To be considered done, a document called `JOLT_FUNCTIONS_OWNERSHIP_INVARIANTS.md` should be created in synthesis/docs/. It should contain a list of all Jolt Physics functions that we use and any relevant ownership invariants they have.

All calls to Jolt functions should be searched for in `fission/src/`. The definitions of all Jolt function should be referened and found in `../../JoltPhysics/jolt/`.

An argument is considered `CONSUMED` if it is either destroyed by the Jolt function or held by the Jolt function and destroyed later (e.g. when the Jolt object is destroyed). An argument is considered `CLONED` if it is not `CONSUMED`, but still holds some important memory. Some arguments do not need to be cleaned up, for example, any of the type `Jolt.EActivation`. If an argument is neither `CONSUMED` nor `CLONED`, then it is `COPIED`. It is still the responsibility of the caller to free `CLONED` arguments. These are analogous concepts to ownership and borrowing in Rust

The format of each function in the document should be as follows:

- `[FUNCTION NAME]`
  - Arguments
    [A LIST OF ARGUMENTS], either `CONSUMED`, `CLONED`, or `COPIED`
  - Returns [Either `INTERNAL_REF` or `COPY`, or `NONE`]

For example:

- `BodyInterface.AddBody(bodyID: Jolt.BodyId, activationMode: Jolt.EActivation)`
  - Arguments
    - `bodyID`: COPIED
    - `activationMode`: COPIED
  - Returns: `NONE`

For functions without ownership concerns, simply list then with a note saying "No Ownership Concerns"
'''

## Verification

I manually reviewed each documented function to verify the documentations correctness. I could of course, have made mistakes. The use of claude in this case basically saved me having to type everything out.

---

- [x] User-facing polish:


Dhruv-0-Arora (commented):
This is a lot of documentation 🤯 

Personally, im not a huge fan of this being added in docs/
I think it should be within the jolt fork submodule itself. In the 
docs/ directory I find that it might just get lost and people won't 
think to look there when working with jolt.

Dhruv-0-Arora (approved):
Conversation was resolved. Looks good 👍

BrandonPacewic (approved):

PR #1340, opened 2026-06-23
Firefox Mobile Rapid Orbiting `[SYNTH-68]`

AlexD717:
## Task

SYNTH-68

## Symptom

When orbiting the camera on Firefox mobile, the camera rapidly spins uncontrollably.

## Solution

This is a known issue with `PointerEvent.movementX`/`movementY`, as different browser use different units for these values causing inconsistent rotation ([https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX)). The fix is to calculate these values manually by comparing the current and previous event.

## Verification

- [x] Orbiting works on Chrome
- [x] Orbiting works on Chrome Mobile (Android)
- [ ] Orbiting works on Chrome Mobile (iPhone)
- [x] Orbiting works on Firefox
- [x] Orbiting works on Firefox Mobile (Android)
- [ ] Orbiting works on Firefox Mobile (iPhone)
- [x] Orbiting works on Safari
- [ ] Orbiting works on Safari Mobile

This code can be tested on your phone by running `bun run dev --host` and then navigating to the ip of your laptop.

---

- [x] User-facing polish:


AlexD717:
Someone should make sure this fixes orbiting on iOS, as I am unable to test that.

0xda157:
was the original issue on firefox mobile on android or ios? because those use different browser engines.

AlexD717:
I was able to reproduce the issue on Android

fission/src/systems/scene/ScreenInteractionHandler.ts
```diff
@@ -194,12 +194,13 @@ class ScreenInteractionHandler {
...
                 }
 
+                const prevPrimary = this._primaryTouchPosition!
                 this._primaryTouchPosition = [e.clientX, e.clientY]
 
                 if (this._secondaryTouch == undefined) {
                     this.interactionMove({
                         interactionType: PRIMARY_MOUSE_INTERACTION,
-                        movement: [e.movementX, e.movementY],
+                        movement: [e.clientX - prevPrimary[0], e.clientY - prevPrimary[1]],
```
Dhruv-0-Arora, line 204: If this is a firefox issue and is patched in the future, could we add a comment here so if it is patched and this fix causes an issue, we can easily identify it?

Dhruv-0-Arora (changes requested):

fission/src/systems/scene/ScreenInteractionHandler.ts
AlexD717: I added a comment explaining why we are calculating it manually. However, I don't think Firefox is planning on fixing this anytime soon. This ticket was created two years ago, and the issue seems to be that Firefox just handles units in a different way, which is intentional.

Dhruv-0-Arora:
Need to test mobile functionality off of firewall

AlexD717:
> Need to test mobile functionality off of firewall

I hotspot of my phone to test. Or you can just test it at home.

PepperLola (commented):
Interestingly I don't have this issue on dev (iOS), so it would be best for someone else who does to review this PR instead

Dhruv-0-Arora:
@AlexD717 I also did not have this issue on my phone when I was using firefox (in dev). How did you get this issue?

Dhruv-0-Arora (commented):
Above ^

AlexD717:
@Dhruv-0-Arora I believe this could be a Android only issue, which would explain why you and Julian were not able to reproduce it.

https://github.com/user-attachments/assets/546a7d5d-b42c-49e6-8376-22e1a103190d

This is what happens to me on dev (I know that this isn't dev, but the same thing happens)

Dhruv-0-Arora:
@AlexD717 I tested prod on firefox on my ios phone again and I didn't seem to run into this issue. 

If this truly is an issue, then I think this PR should be merged. We should definitely validate first whether this is actually an issue or not. Maybe we can ask someone else who has an android device or if anyone has android studio with a simulator if they can open production and try it there?

0xda157 (approved):
diff LGTM. I can reproduce the issue on prod on firefox on my android phone but am unable to test this pr

Dhruv-0-Arora (approved):

PR #1343, opened 2026-06-24
Swerve Drive Support `[SYNTH-93]`

BrandonPacewic:
## Task

SYNTH-93

## Symptom

Swerve drivetrains were completely unsupported. This was something we previously had in Unity but remained un-adapted for our Fission web based simulator due to some latent physics system issues that stalled development.

## Solution

Full automatic detection and handling of Azimuth Swerve modules. Our current detection algorithm mirrors the Unity version (see attached documentation in the diff).

## Verification

1. Load a swerve robot, one can be found [here](https://github.com/user-attachments/files/29312923/SDS_Tutorial_Bot_v4.zip).
2. Import said robot and select the drivetrain to be of type `Swerve` from `Arcade`.
3. And that it! Enjoy your fully drivable swerve robot!

Big thanks to @rutmanz for doing most of the work for me.

---

- [x] User-facing polish:


azaleacolburn (commented):
~The robot does drive, but it's horribly slow and jittery, to the point of not being usable. Is this something we want to fix in this PR, or should we have another ticket for that?~

EDIT: This was an azalea skill issue (I should have configured the drivetrain differently).

One thing that would be helpful would be a "Reset Orientation" button. Also, the strafe left and right controls are reversed on Felix.

fission/src/systems/simulation/behavior/synthesis/drive/SwerveDriveBehavior.ts
```diff
@@ -0,0 +1,188 @@
...
+            ?.getRootNodeId()
+    }
+
+    /** @returns true if the difference between a and b is within acceptableDelta */
+    private static withinTolerance(a: number, b: number, acceptableDelta: number) {
+        return Math.abs(a - b) < acceptableDelta
+    }
+
+    // Sets the drivetrain's target linear and rotational velocity
+    private driveSpeeds(forward: number, strafe: number, turn: number) {
```
azaleacolburn: This function is ~100 lines and should probably be refactored later.

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
```diff
@@ -75,10 +96,28 @@ class SynthesisBrain extends Brain {
...
+            const useSwerve = this.driveType === DriveType.SWERVE && swerveInfo.inSwerve
+            if (this.driveType === DriveType.SWERVE && !swerveInfo.inSwerve) {
+                console.warn("[Swerve] swerve detection failed for this robot; falling back to arcade drive.")
+            }
+
+            if (useSwerve) {
+                this._behaviors.push(this.createSwerveDriveBehavior(swerveInfo.hinges))
+            } else {
+                this._behaviors.push(this.createSkidSteerDriveBehavior(this.driveType === DriveType.ARCADE))
+            }
```
azaleacolburn: pedantic: we should probably just have a ternary for a variable which we unconditionally pass as an argument to `this._behaviors.push`

fission/src/ui/panels/configuring/assembly-config/jointConfigGroups.ts
```diff
@@ -0,0 +1,149 @@
...
+        prefs.driveVelocity = velocity
+        prefs.driveAcceleration = force
+        return
+    }
+
+    driver.maxForce = force
+    const name = driver.info?.name
+    if (!name) return
+    const motors = (prefs.motors ?? []).filter(m => m.name !== name)
+    motors.push({ name, maxVelocity: velocity, maxForce: force })
```
azaleacolburn: This object literal should probably have a type constraint on it (`satisfies MotorPreferences`).

fission/src/ui/panels/configuring/assembly-config/jointConfigGroups.ts
```diff
@@ -0,0 +1,149 @@
...
+const drivetrainGroupProvider: GroupProvider = (drivers, isSwerve) => {
+    const wheels = drivers.filter((d): d is WheelDriver => d instanceof WheelDriver)
+    const azimuth = isSwerve ? drivers.filter((d): d is HingeDriver => d instanceof HingeDriver && d.continuous) : []
+
+    const controls: JointConfigControl[] = []
+    if (wheels.length) controls.push(makeControl("Drive", wheels))
+    if (azimuth.length) controls.push(makeControl("Module Rotation", azimuth))
+    if (controls.length === 0) return { claimed: [] }
+
+    return { group: { id: "drivetrain", name: "Drivetrain", controls }, claimed: [...wheels, ...azimuth] }
```
azaleacolburn, line 108: This should have a type constraint too, or at least this function body should be typed.

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
```diff
@@ -196,21 +236,116 @@ class SynthesisBrain extends Brain {
...
+            )
+            return this.createSkidSteerDriveBehavior(true)
+        }
+
+        // Pair each wheel with its nearest azimuth hinge so paired drivers share an index.
+        // Both positions are taken as world-space anchors, matching the original which paired
+        // on WheelDriver.Anchor / RotationalDriver.Anchor.
+        const wheelPositions = wheelDrivers.map(w => {
+            const t = w.constraint
+                .GetWheelWorldTransform(0, new JOLT.Vec3(1, 0, 0), new JOLT.Vec3(0, 1, 0))
```
azaleacolburn: These jolt vectors should be destroyed as well, unless `GetWheelWorldTransform` does that for us

fission/src/systems/simulation/synthesis_brain/SwervePairing.ts
```diff
@@ -0,0 +1,38 @@
...
+            const distSq = dx * dx + dy * dy + dz * dz
+            if (distSq < minDistSq) {
+                minDistSq = distSq
+                closestSlot = slot
+            }
+        })
+        if (closestSlot === -1) return -1
+        const [chosen] = available.splice(closestSlot, 1)
+        return chosen
+    })
```
azaleacolburn, line 37: There's a lot going on here, could we pull some things (e.g. the lambdas) out into a few different temporary variables.

fission/src/systems/simulation/behavior/synthesis/drive/SwerveDriveBehavior.ts
```diff
@@ -0,0 +1,188 @@
...
+
+        let maxVelocity = new THREE.Vector3()
+        const com = convertJoltVec3ToThreeVector3(World.physicsSystem.getBody(rootNodeId).GetCenterOfMassPosition())
+
+        const velocities: THREE.Vector3[] = []
+        for (let i = 0; i < this._hinges.length; i++) {
+            const driver = this._hinges[i]
+
+            const axis = convertJoltVec3ToThreeVector3(driver.worldAxis).normalize()
+            const radius = convertJoltVec3ToThreeVector3(driver.worldAnchor).sub(com)
```
azaleacolburn: We should be careful about this, once #1320 gets merged, these conversion calls will destroy their arguments, unless a flag is passed in.

fission/src/systems/simulation/behavior/synthesis/drive/SwerveDriveBehavior.ts
```diff
@@ -0,0 +1,188 @@
...
+                w.accelerationDirection = 0
+            })
+            World.physicsSystem.enablePhysicsForBody(rootNodeId)
+            return
+        }
+
+        // Adjusts how much turning versus translation is favored
+        turn *= 1.5
+
+        const chassisVelocity: THREE.Vector3 = robotForward
```
azaleacolburn: I'm suspicious about physics objects (velocities in this case) being represented as ThreeJS vectors.

fission/src/systems/simulation/behavior/synthesis/drive/SwerveDriveBehavior.ts
```diff
@@ -0,0 +1,188 @@
...
+        hinges.forEach(h => {
+            h.controlMode = DriverControlMode.POSITION
+            // Swerve steering modules rotate continuously.
+            h.setContinuousRotation()
+        })
+
+        // Zero field-oriented drive to the robot's spawn heading so "forward" starts aligned with
+        // its nose. The reset input re-zeroes it later; falls back to world +X if the body isn't ready.
+        const rootNodeId = this.resolveRootNodeId()
+        if (rootNodeId != undefined) {
```
azaleacolburn: If we're making a non-type safe comparison, then we might as well wright `if (rootNodeId)`

azaleacolburn (dismissed):
I didn't notice any obvious semantic issues, but I have a few notes about memory management and style. I would be happy to make all of the changes I recommend.

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
```diff
@@ -196,21 +236,116 @@ class SynthesisBrain extends Brain {
...
+                    `${hingeDrivers.length} azimuth hinges). Falling back to arcade.`
+            )
+            return this.createSkidSteerDriveBehavior(true)
+        }
+
+        // Pair each wheel with its nearest azimuth hinge so paired drivers share an index.
+        // Both positions are taken as world-space anchors, matching the original which paired
+        // on WheelDriver.Anchor / RotationalDriver.Anchor.
+        const wheelPositions = wheelDrivers.map(w => {
+            const t = w.constraint
```
azaleacolburn: `t` should be destroyed

RoushilS:
The robot likes to fall down when I use the move tool.

https://github.com/user-attachments/assets/401e5ccb-4fd9-4719-b167-ad743bb42b03

fission/src/systems/input/DefaultInputs.ts
```diff
@@ -200,6 +200,64 @@ class DefaultInputs {
...
+        return {
+            schemeName: "Felix",
+            descriptiveName: "WASD + Arrows (Swerve)",
+            customized: false,
+            usesGamepad: false,
+            usesTouchControls: false,
+            supportedDrivetrains: [DriveType.SWERVE],
+            inputs: [
+                AxisInput.onKeyboard("swerveForward", "KeyW", "KeyS"),
+                AxisInput.onKeyboard("swerveStrafe", "KeyD", "KeyA"),
```
RoushilS:
```suggestion
                AxisInput.onKeyboard("swerveStrafe", "KeyA", "KeyD"),
```

Switch Inverted Controls

fission/src/systems/simulation/synthesis_brain/SwervePairing.ts
BrandonPacewic: This function is already very readable imo

fission/src/systems/simulation/behavior/synthesis/drive/SwerveDriveBehavior.ts
BrandonPacewic: Preexisting this PR

BrandonPacewic:
1. I'm kicking reset orientation down the road.
2. The move bug seems to reflect on the cad file itself more than this PR.

I have made the following two tickets:
- SYNTH-170
- SYNTH-173

RoushilS:
@azaleacolburn reset orientation already exists, it should be "r" on felix. my PR introduces a physical button.

Dhruv-0-Arora (approved):
I tested this with a controller and it worked really well.

https://github.com/user-attachments/assets/bdca1398-5535-425d-8a25-f0f1c4f54f64

rutmanz (changes requested):
The biggest thing I noticed is that you can't rotate and translate at the same time, which is not how (good) real swerve implementations work. You can see it in @Dhruv-0-Arora's video too, but if you press the rotate buttons it just rotates regardless of what translation inputs you're giving it.

I would also update the assetpack to include a swerve robot by default (or make a ticket to do that) so that people can see it when they spawn in. Might also make sense to have swerve vs differential embedded in the mirabuf data, possibly through a devtool, also could be another ticket. Otherwise looks great


https://github.com/user-attachments/assets/3ce86d8c-8b88-4819-9296-ceee3a0aff97

fission/src/systems/simulation/behavior/synthesis/drive/SwerveDriveBehavior.ts
```diff
@@ -0,0 +1,194 @@
...
+        if (forward === 0 && strafe === 0 && turn === 0) {
+            this._wheels.forEach(w => {
+                w.accelerationDirection = 0
+            })
+            World.physicsSystem.enablePhysicsForBody(rootNodeId)
+            return
+        }
+
+        // Adjusts how much turning versus translation is favored.
+        turn *= 1.5
```
RoushilS, line 103: @rutmanz you can play around with this hardcoded factor which helps. though the robot does suffer from path curvature drift.

Dhruv-0-Arora:
@rutmanz when I noticed the turning and driving issue I thought it was to do with our physics implementation. If we are to fix it I think we should make it its own pr

BrandonPacewic (commented):
I added SYNTH-184. I don't think we should include this robot since it reaches a lot of fail states in the importer (see logging when importing the robot).

I addressed not being able to translate and rotate at the same time and added feedforward logic to fix the spline-like robot drift.

RoushilS (approved):
I have done a lot of testing and can't find any other issues. lgtm!

fission/src/systems/input/DefaultInputs.ts
```diff
@@ -200,6 +200,64 @@ class DefaultInputs {
...
+            descriptiveName: "Dual Stick (Swerve)",
+            customized: false,
+            usesGamepad: true,
+            usesTouchControls: false,
+            supportedDrivetrains: [DriveType.SWERVE],
+            inputs: [
+                AxisInput.onGamepadJoystick("swerveForward", 1, true),
+                AxisInput.onGamepadJoystick("swerveStrafe", 0, false),
+                AxisInput.onGamepadJoystick("swerveTurn", 2, false),
+                ButtonInput.onGamepad("swerveResetFieldForward", 8),
```
RoushilS, line 249: I do have one minor thing though - when I use gamepad with the current joystick invert defaults, they don't match up. Having swerveForward = false, swerveStrafe = false, and swerveTurn = true worked for me. @Dhruv-0-Arora or @AlexD717  did you have the same issue?

BrandonPacewic: If someone can confirm this for me that would be much appreciated as I don't have a way of testing this myself.

rutmanz (approved):

fission/src/systems/input/DefaultInputs.ts
AlexD717: I can confirm this issue. @RoushilS fix of swerveForward = false, swerveStrafe = false, and swerveTurn = true fixed it for me. In addition, inverting all of them (swerveForward = true, swerveStrafe = true, and swerveTurn = true) also fixed the issue.

AlexD717 (approved):
lgtm, the controller issue has been fixed. I agree with @rutmanz, that for a separate PR it would be very nice if we could embed if a robot is swerve directly in the mirabuf file. That way the spawn asset menu can automatically switch to a swerve drivetrain.

PR #1345, opened 2026-06-25
APS Icon Fixes (Safer SVG Rendering)

AlexD717:
## Symptom

Logging in with APS caused some top bar icons to become invisible.

## Solution

Reworked top bar icon rendering to keep inline SVG behavior while using a safer implementation:

- Removed `dangerouslySetInnerHTML` usage in `TopBarIcons.tsx`.
- Kept SVG inlining from local icon assets, but now parse SVG markup with `DOMParser` and append cloned SVG nodes.
- Preserved SVG ID scoping to avoid cross-icon ID collisions.

## Verification

- [x] Top bar icons render as inline SVG (not ``).
- [x] Multiple icon instances render correctly without collisions.
- [x] Added targeted test coverage in `src/test/ui/TopBarIcons.test.tsx`.
- [x] Logging in with APS causes no icons to disappear.

---

- [ ] User-facing polish:


azaleacolburn (commented):
I think we need to talk about these changes before merging, as the solution seems pretty scuffed and I don't fully believe everything Claude wrote in the `TopBarIcons.tsx` comment.

Also, the usage of `dangerouslySetHtml` is questionable, even though it doesn't use user input.

Dhruv-0-Arora:
@copilot what are your thoughts on the changes made in TopBarIcons.tsx
Do you think that there is a better and safer way to implement that fix?

Dhruv-0-Arora:
@AlexD717 could you take a look again and see if this still fixes it for you?

AlexD717:
@Dhruv-0-Arora Yes, this also fixed it for me.

PR #1346, opened 2026-06-25
Safari support for caching mira assemblies `SYNTH-131`

Dhruv-0-Arora:
## Task

Adding support for caching robots on Safari. Currently, we use OPFS for caching robots which is only for Chrome, Firefox, and Edge. I used IndexDB as a Fallback option when OPFS doesn't work so that safari can cache robots and they don't need to be downloaded and every time a user wants to spawn an assembly.

This PR abstracts a lot of the MirabufLoader functionality into the MirabufStorageBackend file.

SYNTH-131

## Verification

Ensure that assemblies spawn properly and can be configured properly when cached and spawned from the cache.

---

- [ ] User-facing polish:


AlexD717:
If we are now planning to support safari, should we now also have our tests run on webkit (safari), in addition to the current chromium and firefox tests? This could be a different PR, since if we do add it, a lot of tests are currently failing.

rutmanz:
When I run this in Safari (Version 26.5, macOS 26.5), it says "using opfs backend" in the logs and my (brief) research indicates that Safari has supported OPFS for a while. Caching also works fine on Safari on the dev branch for me.

[The MDN page](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) also says it's been supported on both mobile and desktop safari since 2021.

What's the behavior for you?

Dhruv-0-Arora:
@rutmanz Safari does support OPFS but it doesn't have the createWritable() method which the original OPFS caching implementation uses mainly to save files.
Also, when I was using safari on my MacOS, it was logging
```Mirabuf cache: using IndexedDB backend```
I'm not exactly sure why yours is using OPFS.

Also caching is disabled on safari in dev ?? When you spawn a robot it should not be added to "saved robots". That is at least what happens for me. I'm not sure how your thing worked.

Screenshot 2026-06-29 at 2 01 56 PM


@AlexD717 adding webkit to our CI tests is a good idea. I think we might want to consider that for a future PR since the current contract tests I added are validating the IndexedDB logic directly.

File: fission/src/test/mirabuf/MirabufStorageBackend.test.ts
Comment (Dhruv-0-Arora) [inline fission/src/test/mirabuf/MirabufStorageBackend.test.ts:1]: The generated tests all seem quite useful. I don't think we should remove any. I reviewed them quite thoroughly.

**Comment by Dhruv-0-Arora** [comment]
Update: @rutmanz found that the createWritable() function was added in Safari 26. This means that this PR is entirely redundant / useless for anyone using safari 26. However, since there are users (such as myself) that still use versions under safari 26, I think it is best to continue with looking into this PR.

File: fission/src/test/mirabuf/MirabufLoader.test.ts [hunk 1]
```diff
@@ -103,18 +103,14 @@ describe("MirabufLoader", () => {
             assert.exists(field1)
             assert.exists(robot1)
             expect(MirabufLoader.getAll()).toHaveLength(2)
+
             await MirabufLoader.removeAll()
+
+            // Metadata index must be empty.
             expect(MirabufLoader.getAll()).toHaveLength(0)
+            // Assemblies must no longer be retrievable through the public API.
```
rutmanz: this comment does not make sense (and smells AI generated)

fission/src/test/mirabuf/MirabufStorageBackend.test.ts
```diff
@@ -0,0 +1,388 @@
...
+                },
+                async removeAll() {
+                    return new Promise((resolve, reject) => {
+                        const req = db.transaction("assemblies", "readwrite").objectStore("assemblies").clear()
+                        req.onsuccess = () => resolve()
+                        req.onerror = () => reject(req.error)
+                    })
+                },
+            }
+        })
```
rutmanz: Uh this is cheating. This test doesn't test the actual behavior of anything because it's reimplementing. Mock an error from OPFS if you want to test like this

fission/src/test/mirabuf/MirabufStorageBackend.test.ts
rutmanz: There are quite a few that are redundant imo but it's not necessarily a problem

rutmanz (changes requested):

fission/src/test/mirabuf/MirabufLoader.test.ts
Dhruv-0-Arora: Ya the tests were - the commit message for the tests also makes sure we don't forget

rutmanz (approved):

fission/src/mirabuf/MirabufLoader.ts
```diff
@@ -22,35 +23,11 @@ export interface MirabufRemoteInfo {
...
-            console.log(`No access to OPFS`)
-            return false
-        }
-    } catch (_e) {
-        console.log(`No access to OPFS`)
-        return false
-    }
-})()
+/** @deprecated Use storageBackend != null instead. Kept for backward compatibility. */
+export const canOPFS = storageBackend != null
```
AlexD717: This doesn't appear to be used or accessed anywhere, in which case it doesn't make sense for us to still have it.

fission/src/mirabuf/MirabufStorageBackend.ts
```diff
@@ -0,0 +1,205 @@
...
+    try {
+        const root = await navigator.storage.getDirectory()
+        const dir = await root.getDirectoryHandle(OPFS_DIR_NAME, { create: true })
+        if (dir.name !== OPFS_DIR_NAME) return null
+
+        // Probe for createWritable() support; this is what Safari (< 26) lacks on the main thread
+        const testHandle = await dir.getFileHandle("__opfs_probe__", { create: true })
+        const writable = await testHandle.createWritable()
+        await writable.close()
+        await dir.removeEntry("__opfs_probe__")
```
AlexD717: As far as I understand, since Safari 26 lacks createWritable(), the code will throw on line 174, and lines 175-176 are never executed, resulting in a __opfs_probe __ being created and never cleaned up. Downgrading from Safari 26 isn't easy, so this isn't something I can test or verify.

AlexD717 (changes requested):
Could we also add a comment somewhere, so that in a few years once Safari 18 is fully obsolete we no longer have to support IndexDB as well.

fission/src/mirabuf/MirabufStorageBackend.ts
Dhruv-0-Arora: 👍

AlexD717 (approved):
Changes look good to me. I'm not able to test on Safari 18, but Safari 26 and Chrome still work.

PR #1350, opened 2026-06-25
Jolt static-temporary frees and PhysicsSystem teardown `[SYNTH-109]`

BrandonPacewic:
These changes sit on top of the merged WASM memory cleanup PR (#1320 SYNTH-109) and fix three corruption/leak bugs it left behind. All Jolt `[Value]`-returning methods hand back a pointer to a reused static temporary, not a heap allocation, passing those to `JOLT.destroy()` frees a non-heap address and corrupts the dlmalloc free list, which later traps as memory access out of bounds.

- Previously called `JOLT.destroy(constraintMatrix)` on the result of `GetConstraintToBody1Matrix()` (a static temp) -> heap corruption on robot import. Fix: removed that destroy. Only wheelPos (a real new-allocated `RVec3`) is freed. Inlined `SubRVec3` since its result is also a non-freeable temp. Corrected the stale "internal reference" comment.

- Previously called `JOLT.destroy(this._joltBodyInterface)`. `BodyInterface` is an embedded value member of `PhysicsSystem`, not a heap allocation → interior-pointer free → corruption. Removed.
- PR called `this._joltPhysSystem.GetContactListener()` after `JOLT.destroy(this._joltInterface)` destroying JoltInterface deletes the PhysicsSystem, so `_joltPhysSystem` was dangling -> use-after-free. Fix: capture the listener before destroying `JoltInterface`.

- `JOLT.destroy(joltSettings)` commented out leaking it every construction. Verified `JoltInterface` only copies the filter pointers and `JoltSettings` has no destructor, so it's safe to free.


Dhruv-0-Arora (approved):

azaleacolburn (approved):
Thanks for the fixes!

PepperLola (approved):

PR #1353, opened 2026-06-26
Fix gamepiece early ejection direction `[SYNTH-78]`

RoushilS:
## Task

SYNTH-78

## Symptom

If you eject a gamepiece when while it is still being intaken and hasn't reached it's final desired position, it does not shoot in the direct direction.

## Solution

Set the eject vector to the final desired eject vector rather than the gamepiece's immediate one.

## Verification

To Verify: First, spawn a robot in a field, and configure ejector and intake (make sure to make intake animation take 2 seconds for ease of use). Then intake and immediately eject.

- [ ] Game piece shoots in the right direction when being ejected right after intaking.

---

- [ ] User-facing polish:

### DIFF
--- a/fission/src/mirabuf/EjectableSceneObject.ts
+++ b/fission/src/mirabuf/EjectableSceneObject.ts
@@ -30,6 +30,9 @@ class EjectableSceneObject extends SceneObject {
 
     private _desiredQuatRotation?: THREE.Quaternion
 
+    // Set true once the animation has reached the ejection position at least once.
+    private _reachedEndPosition = false
+
     private static _defaultAnimationDuration = 0.5
 
     public static setAnimationDuration(duration: number) {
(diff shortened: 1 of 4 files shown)


Dhruv-0-Arora (changes requested):
I personally think nothing should happen. In a game, if your ejector is running while you are intaking, then the gamepiece can't even be ejected because of physics. 
So if your assembly is currently intaking, then just don't eject until it is done intaking

RoushilS:
oh is that what the ticket meant? @PepperLola

PepperLola:
> oh is that what the ticket meant? @PepperLola

You've implemented what I initially meant but I actually prefer Dhruv's suggestion as well since that makes more sense in a match

Dhruv-0-Arora:
@RoushilS is this ready for review?

RoushilS:
No, sorry I took a break from this project and have been working on other things as I kept running into bugs when trying to stop/cancel eject behavior after adding the check. I will come back to it soon. The version that simply has the object drop, and the one that has it launch both work.

PR #1354, opened 2026-06-26
Collision Zone Performance Refactor `[SYNTH-132]`

azaleacolburn:
## Task

SYNTH-132

## Symptom

Performance tanks when robots and pieces enter zones. Generally zones cause poor performance.

This is due to the way in which `ZoneSceneObject`s handle collisions. Basically, each one creates a sensor, then subscribes to every collision between all Jolt bodies, this is horribly expensive for two reasons:

1. Not every zone should care about every object, e.g., scoring zones only care about game pieces. So the zones have to filter through a lot of information they don't care about.
2. A lot of Jolt body collisions happen, so for complicated robots like 2471's 2018 bot, each zone has to deal with each body's collision with the zone's sensor.

## Solution

Each sub-class of `ZoneSceneObject` now implements a function called `checkObjectsInZone`, which is responsible for manually going through every object that the zone cares about and checking if they are in the zone, then performing the needed actions given that information.

Instead of listening for collisions, these functions (in `ScoringZoneSceneObject` and `ProtectedZoneSceneObject`) draw bounding boxes around every object they care about (e.g. each robot), then check if that bounding box overlaps with that of the zone.

## Verification

We don't have standardized performance benchmarks, but the lag that used to be present when a robot enters a protected zone is gone. Try this with 2471's 2018 bot, the difference is actually huge (on my machine).

The performance for game pieces in scoring zones is gone, however on some fields (e.g. 2023), there complicated jolt bodies around the scoring zone, which can cause placed bodies to lag slightly. However, other parts of the scoring zone are unaffected.

Tests have been updated to match the new way zones work.

---

- [x] User-facing polish:


RoushilS (approved):
lgtm, fixes the huge performance issue.

azaleacolburn:
Currently blocked by https://github.com/jrouwe/JoltPhysics.js/pull/297, since some necessary methods are not currently exposed on the `OrientedBox` Jolt class.

fission/src/mirabuf/ScoringZoneSceneObject.ts
```diff
@@ -1,165 +1,116 @@
...
+    public override update(): void {
+        super.update()
+    }
 
-        this.unsubscribers.forEach(unsubscribe => unsubscribe())
+    public reset() {
+        this._prevGPs.length = 0
     }
 
+    /// Updates points for alliance and robot when game piece enters this scoring zone
```
AlexD717:
```suggestion
    // Updates points for alliance and robot when game piece enters this scoring zone
```

AlexD717 (changes requested):
The protected zone collisions seem to be triggering when they are not supposed to sometimes (had it happen multiple times but doesn't happen every time)

https://github.com/user-attachments/assets/24609f13-c494-43de-ae12-e511623088ce

fission/src/mirabuf/EjectableSceneObject.ts
```diff
@@ -84,12 +83,6 @@ class EjectableSceneObject extends SceneObject {
 
             World.physicsSystem.disablePhysicsForBody(this._gamePieceBodyId)
 
-            // Remove from any scoring zones
-            const zones = World.sceneRenderer.filterSceneObjects(x => x instanceof ScoringZoneSceneObject)
-            zones.forEach(x => {
-                if (this._gamePieceBodyId) ScoringZoneSceneObject.removeGamepiece(x, this._gamePieceBodyId)
-            })
-
```
rutmanz, line 92: what was the original purpose of this

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -200,6 +201,21 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
         return this.mirabufInstance.parser.assemblyId
     }
 
+    public getBounding(): Jolt.OrientedBox {
+        const box = this.computeBoundingBox()
+
+        const aabb = new JOLT.AABox(convertThreeVector3ToJoltVec3(box.min), convertThreeVector3ToJoltVec3(box.max))
+        const obb = convertAABBToOBB(aabb)
```
rutmanz: Ok so I can guess that OBB is oriented bounding box, but no idea what AABB is, so I would um request that maybe we use the government names instead of these TAFLAs

fission/src/util/Utility.ts
```diff
@@ -40,6 +40,10 @@ export async function hashBuffer(buffer: ArrayBuffer): Promise<string> {
         .join("")
 }
 
+export function forPair<T, U>(listOne: T[], listTwo: U[], predicate: (one: T, two: U) => void): void {
+    listOne.forEach(a => listTwo.forEach(b => predicate(a, b)))
+}
```
rutmanz, line 47: If list1 and list2 are the same, this checks every pair twice. Might be worth filtering list2 by if the element is in list1

fission/src/mirabuf/ProtectedZoneSceneObject.ts
```diff
@@ -31,161 +45,105 @@ class ProtectedZoneSceneObject extends ZoneSceneObject<ProtectedZonePreferences>
...
+
+    /**
+     * Handles collisions between two robots when either one of those robots is in this protected zone
+     */
+    private handleContactPenalty(body1: MirabufSceneObject, body2: MirabufSceneObject) {
+        // Ensures that infinite collisions do not occur
+        if (Date.now() - this._lastRobotCollisionTime < 500) return
+
+        // Find the robot that has the opposite alliance from the zone
+        const opposingRobot = [body1, body2].find(robot => robot.alliance !== this.prefs?.alliance)
```
rutmanz, line 150:
```suggestion
        const opposingRobot = [body1, body2].find(robot => robot.alliance !== this.prefs.alliance)
```

It looks like this is no longer optional (see the function above)

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -149,6 +149,9 @@ class PhysicsSystem extends WorldSystem {
         return this._pauseSet.size > 0
     }
 
+    public get bodies(): readonly Jolt.BodyID[] {
```
rutmanz: oh that's cool I didn't know you could use readonly like this

fission/src/util/Utility.ts
rutmanz: or maybe just having a version of this function for the case that the lists are identical

rutmanz: or maybe it's cheap enough that you don't care

rutmanz (changes requested):
I'm only requesting changes for the acronyms. Also because I could reproduce Alexey's issue with premature contact penalties

fission/src/util/Utility.ts
azaleacolburn: Ya we filter later

fission/src/mirabuf/MirabufSceneObject.ts
azaleacolburn: Okay

fission/src/mirabuf/ScoringZoneSceneObject.ts
azaleacolburn: I think this is fixed

fission/src/mirabuf/EjectableSceneObject.ts
azaleacolburn: It was old game pieces removing themselves from old scoring zones. I don't know why this would ever be needed, but we're doing things in such a way that this couldn't ever be necessary.

Plus, the function being called here wouldn't even work, since it compares objects of different origin by reference.

fission/src/mirabuf/ZoneSceneObject.ts
```diff
@@ -50,126 +27,164 @@ export default abstract class ZoneSceneObject<P extends object> extends SceneObj
...
+        const newTransform = World.physicsSystem.getBody(this.parentBodyId!)!.GetWorldTransform()
+        const transformHasNotUpdated =
+            this._cachedFieldTransformation && newTransform.Equals(this._cachedFieldTransformation)
+
+        // Update translation, rotation, and scale only if the field has moved
+        if (transformHasNotUpdated && !this._deltaTransHasUpdated) return undefined
+
+        if (this._cachedFieldTransformation) JOLT.destroy(this._cachedFieldTransformation)
+        this._cachedFieldTransformation = newTransform
+        this._deltaTransHasUpdated = false
```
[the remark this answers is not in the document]
azaleacolburn: I believe this comment is incorrect and should be removed

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -201,19 +214,9 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
         return this.mirabufInstance.parser.assemblyId
     }
 
-    public getBounding(): Jolt.OrientedBox {
+    public getBounding(): Jolt.AABox {
```
rutmanz: So wait are we doing the thing where a robot 45 deg from a major axis has an extra big box?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -319,25 +322,24 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+        if (targetControls && this.isOwnObject && (this.miraType === MiraType.ROBOT || !targetControls.focusProvider)) {
+            targetControls.focusProvider = this
         }
 
         EventSystem.dispatch("MirabufObjectChangeEvent", this)
     }
 
     // Centered in x-z plane, bottom surface of object
-    public getPositionTransform(vec: THREE.Vector3 = new THREE.Vector3()): THREE.Vector3 {
+    public getXYPositionTransform(vec: THREE.Vector3 = new THREE.Vector3()): THREE.Vector3 {
```
rutmanz: are we using xy or xz here?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -367,7 +374,7 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
 
         // TODO
         // Why are we calling this?
```
rutmanz, line 377: One of life's great mysteries I guess

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -759,6 +766,122 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+            const triangleContext = new JOLT.ShapeGetTriangles(shape, biggest, shape.GetCenterOfMass(), identity, scale)
+            const vertices = new Float32Array(
+                JOLT.HEAP32.buffer,
+                triangleContext.GetVerticesData(),
+                triangleContext.GetVerticesSize() / Float32Array.BYTES_PER_ELEMENT
+            )
+
+            const vertex = new JOLT.RVec3()
+            const transformedVertex = new JOLT.RVec3()
+            for (let i = 0; i < vertices.length; i += 3) {
```
rutmanz, line 817: I would extract this 3 to a constant but have you tuned it? Also can you make the bounding box visualizer part of the debug panel?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -759,6 +766,122 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+                // Transform the vertex into the position it would occupy if the robot were axis aligned
+                vertex.Set(vertices[i], vertices[i + 1], vertices[i + 2])
+                multiplyMat44ByVec3(transformedVertex, vertex, vertexTransform)
```
rutmanz: If the matrix isn't being transformed, this is more like `multiplyVec3ByMat44`

fission/src/mirabuf/MirabufSceneObject.ts
rutmanz: oh you already did that I didn't scroll down enough. Would still be nice just as a debug tool and not part of match mode

rutmanz (changes requested):
Match mode freezes the tab

fission/src/mirabuf/MirabufSceneObject.ts
azaleacolburn: Oh this function shouldn't be used

fission/src/mirabuf/MirabufSceneObject.ts
azaleacolburn: For another PR, this function appears to be externally pure

fission/src/mirabuf/MirabufSceneObject.ts
azaleacolburn: I don't think tuning it is the best option. While we do only care about each component in the end, the vector is being transformed as a single component, so going vertex by vertex works. We could do this in greater multiples of three i suppose though. I'll test that in another pr

azaleacolburn: I don't want to mess with the UI lol, adding an option to debug bounding boxes could be a separate pr

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -766,6 +761,11 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
         }
     }
 
+    private getInverseRotation() {
```
rutmanz: `MirabufSceneObject.getInverseRotation` is not clear enough to tell me what that does. I imagine it makes more sense when you've spent the last like week thinking about this stuff but basically you should either leave a comment or consider a different name that's more descriptive of its purpose

fission/src/mirabuf/ProtectedZoneSceneObject.ts
```diff
@@ -1,191 +1,156 @@
...
 import type MirabufSceneObject from "./MirabufSceneObject"
-import type { RigidNodeAssociate } from "./MirabufSceneObject"
 import { ContactType } from "./ZoneTypes"
 import type { ProtectedZonePreferences } from "@/systems/preferences/PreferenceTypes"
+import MatchMode from "@/systems/match_mode/MatchMode"
+import type Jolt from "@azaleacolburn/jolt-physics"
+import { findListDifference, forPair, renderOrientedBox } from "@/util/Utility"
+import JOLT from "@/util/loading/JoltSyncLoader"
+
+const DEBUG_BOUNDING_BOXES = true
```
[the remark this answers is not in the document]
rutmanz: and we're back

rutmanz (commented):
Image

it's better but still a little funky

fission/src/mirabuf/MirabufSceneObject.ts
azaleacolburn: I mean, it gets the inverse rotation matrix of the root body of the mirabuf scene object. I'll add that as a comment, but I think that it's pretty straight forward.

azaleacolburn:
I don't actually know what to do about the slight offset, like I'm not sure what would cause it, but I'm also not even sure that it matters from a practical perspective.


EDIT:
Okay so I did some testing and it's actually quite far off sometimes, especially when Dozer isn't axis-aligned 3:

I believe the issue may be that we're getting the position transform by computing the axis aligned bounding box, then getting the center of that bounding box with three js. I think the problem then is that this center is slightly off the less axis-aligned the robot is.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -861,12 +865,25 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+        // Get root body transformation
         const rootBody = World.physicsSystem.getBody(this.getRootNodeId()!)!
-        // Do not destroy
+        // NOTE Do not destroy
         const rotation = rootBody.GetRotation()
 
-        const position = convertThreeVector3ToJoltVec3(this.getPositionTransform())
+        // Here, we calculate the vector between the center of the robot when axis-aligned (which is should be initially) and the root node transform
+        // WARNING This requires the robot to be axis-aligned initially. This may not always be true.
+        if (!this._unrotatedRootNodeToCenterPositionTranslation) {
```
rutmanz: make own function

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -761,7 +762,10 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
         }
     }
 
-    private getInverseRotation() {
+    /**
+     * @returns The transformation matrix which corresponds to the reverse of the current spacial rotation of the root node of this scene object, relative to the
```
rutmanz: RELATIVE TO THE

AlexD717 (changes requested):
If you turn on active in sandbox mode on a protected zone, click save, then try to access the configuration again, the website crashes.

https://github.com/user-attachments/assets/15db187f-b036-4e4a-a2c6-2f3f2932105a

Note: You also must have a robot spawned when enabling sandbox mode on a protected zone. Otherwise there is no crash.

AlexD717 (changes requested):
Image

For me it looks like the bounding boxes are still a bit off (when enabling DBEUG_BOUNDING_BOXES), and the collision also behave weirdly.

rutmanz:
image
I would also note that spawning additional robots moves the bounding box of the other robots

azaleacolburn:
Gang I promise the bounding boxes were tightly fitting before 😭

fission/src/mirabuf/ProtectedZoneSceneObject.ts
```diff
@@ -1,191 +1,162 @@
...
-                    this.zoneCollisionRemoved(body2)
-                } else if (idx2 == bodyIdx) {
-                    this.zoneCollisionRemoved(body1)
-                }
+        if (DEBUG_BOUNDING_BOXES) {
+            this.robotBounding?.forEach(m => {
+                World.sceneRenderer.removeObject(m)
+                m.geometry.dispose()
+                const materials = Array.isArray(m.material) ? m.material : [m.material]
+                materials.forEach(mat => mat.dispose())
```
AlexD717: Can you extract this into a helper function and then also call it on dispose? Currently changing the configuration causes the bounding boxes of current robots to stick around.

Image

AlexD717 (changes requested):
Minor debug bounding box duplication glitch, but otherwise lgtm.

AlexD717 (approved):
lgtm

rutmanz (approved):
image

Up to you if you make this a separate ticket, but the boxes are not currently recalculated on joint motion

azaleacolburn:
> Up to you if you make this a separate ticket, but the boxes are not currently recalculated on joint motion

@AlexD717 and I already had a conversation about that this morning. The plan is to handle that when it actually matters in SYNTH-228.

rutmanz:
> > Up to you if you make this a separate ticket, but the boxes are not currently recalculated on joint motion
> 
> @AlexD717 and I already had a conversation about that this morning. The plan is to handle that when it actually matters in [SYNTH-228](https://autodesk.atlassian.net/browse/SYNTH-228).

Technically it's not just for the dimension tracker that it matters, a robot that has a deployable intake is inside a protected zone when the pre-deploy bounding box (bumpers) are not. As long as it gets fixed later/is tracked though

PepperLola (approved):

PR #1355, opened 2026-06-26
ci(bun): migrating fission unit tests to bun [SYNTH-179]

Dhruv-0-Arora:
## Task

FissionUnitTest.yml is currently using npm and not bun.

SYNTH-179

---

- [ ] User-facing polish:


.github/workflows/FissionUnitTest.yml
```diff
@@ -19,10 +19,16 @@ jobs:
...
-      - name: JavaScript Setup
-        uses: actions/setup-node@v4
+
+      - name: Install System Dependencies
+        run: |
+          apt-get update
+          apt-get install -y git-lfs unzip
+
+      - name: Bun Setup
+        uses: oven-sh/setup-bun@v2
```
Dhruv-0-Arora, line 32: This is probably just worse than keeping it where it was. Worth looking at with another set of eyes though

Dhruv-0-Arora:
I'm not sure where our consensus was on this PR. If we want to scrap it, then we should close this PR. Could be worth looking into though.

I believe the reason we use npm was because in deployments we use yarn/npm. @PepperLola knows why

.github/workflows/FissionUnitTest.yml
PepperLola: I think it's fine, it has to run either way. At least this way if it fails it's earlier in the run.

PepperLola (approved):
Since the tests pass and we now use bun to build/deploy, I'm fine with moving the unit tests over to bun as well.

.github/workflows/FissionUnitTest.yml
```diff
@@ -70,10 +74,16 @@ jobs:
...
+      - name: Install System Dependencies
+        run: |
+          apt-get update
+          apt-get install -y unzip
+
+      - name: Bun Setup
+        uses: oven-sh/setup-bun@v2
         with:
-          node-version: 20
+          bun-version: latest
```
Dhruv-0-Arora: this might not be a good idea? @PepperLola

PepperLola: I agree with pinning the bun version; I noticed that when I was reviewing but forgot

AlexD717 (approved):

rutmanz (approved):

PR #1361, opened 2026-06-29
Fix Gamepad custom selection `[SYNTH-152]`

RoushilS:
## Task

SYNTH-152

## Symptom

The custom gamepad selection buttons were blank and unsuable. Additionally, the invert gamepad toggles didn't visually update.

## Solution

Change button behavior in `JoystickButtonSelection` to be owned by the button rather than a value, and configure `useState` for Invert Joystick toggle.

## Verification

- [ ] User can configure custom gamepad buttons. 
- [ ] Invert Joystick toggle reflects internal value.
 
---

- [ ] User-facing polish:


AlexD717 (approved):
Nice job. I found a pre existing input bug that should be fixed in a later PR (SYNTH-185). In addition there is no spacing on the first dropdown (though this also isn't exactly related to your PR).

Image

Dhruv-0-Arora (approved):
Nice

I'm not sure we need the joystickInverted state since it adds a few lines of code where they aren't necessarily needed. However, it does look more consistent with the rest of the code so it shouldn't be a problem

PR #1368, opened 2026-06-30
Snap Field Placement `[SYNTH-94]`

AlexD717:
## Task

SYNTH-94

## Symptom

Currently fields spawn slightly above the ground which is undesirable. I think this is most obvious on the 2025 field, but this can be seen on all fields.
Screenshot 2026-06-30 at 2 38 04 PM

Right now the ground plane is located from -0.19 to 0.01 (y-axis), and fields spawn at 0.1.

## Solution

Have the ground plane go from -0.2 to 0, and fields spawn at 0.

## Verification

- [x] Fields still spawn normally
- [x] There is no longer any gab between fields and the ground plane.

---

- [x] User-facing polish:


RoushilS:
I wonder why this was offset in the past. I imagine it was on purpose.

RoushilS (approved):
Good fix! This also solves the issue of not being able to spawn field after spawning robot without glitching. I will say that `fun field v19` does not snap to bottom, but that field definetely has other issues, lol.

AlexD717:
> Good fix! This also solves the issue of not being able to spawn field after spawning robot without glitching. I will say that `fun field v19` does not snap to bottom, but that field definetely has other issues, lol.

Thanks. However, `fun field v19` does actually snap to its most bottom point, which is the "river" and that is the intended behavior.

PepperLola (approved):

PR #1371, opened 2026-07-01
Allow storing preferences in mirabuf files and refactor preferences system `[SYNTH-194]` and `[SYNTH-190]`

rutmanz:
## Task

SYNTH-194
SYNTH-190
SYNTH-118

## Symptom & Solution

For 194:
- user preferences (not related to robot or field) are stored as top level keys and called global preferences. This is weird from a data structure perspective (especially bc we iterate over them), and this PR moves those to a "User" object
- Remove motor preferences (they are unused)
- Automatic migration of old preferences
- Use mira hashes rather than model name as preference keys.

For 190:
- Remove devtools panel (this was just used for exporting updated mira models, which is now in the configure panel)
- Automatically migrate old stored metadata
- Store all preferences (linked joints, drivetrain properties, etc) instead of just specific types so that one student can configure the robot and share it easily with all teammates.
- Add metadata configuration page to allow students to rename assets. As we try more and more to de-emphasize the exporter, this will let them customize things

image

For 118:
- Update all robot models to include sensible defaults

## Verification

- Run `bun run assetpack` to get the new models
- Import an older dozer with configured stuff (Dozer v10 from the old assetpack) to check migration
- Try making changes, exporting, and importing into a private window, see configuration still there
- Make sure that user-modified preferences aren't overwritten 
---

- [ ] User-facing polish:


rutmanz:
I'm working on getting the checks to pass but this PR is ready for review

fission/bun.lock
rutmanz, line 1: @azaleacolburn do you have any ideas for a long term solution to the autodesk registry getting added to the lockfile?

AlexD717:
> For 118:
Update all robot models to include sensible defaults (still need to do)

Is this still in progress?

rutmanz:
> > For 118:
> 
> Update all robot models to include sensible defaults (still need to do)
> 
> 
> 
> Is this still in progress?

No that's done now

AlexD717:
Maybe it was just me, but I appreciated the ability to view the .mira file configuration in the developer tools panel. When I was developing camera points I used that to manually add some points and verify that the camera functionality worked, before spending a lot of time adding a user facing configuration option. Personally I would prefer if it was just hidden if the user isn't a developer, but I understand not wanting to have a dedicated button in the HUD for what's a pretty niche development use case.

I also have some minor robot nitpicks. I think it would be best to implement them as users should have a great experience with any default robot, but its up to you.
- 2024 Kitbot drives a bit too fast and can easily tip making it feel unrealistic. I would reduce the max velocity to something like 20-25.
- 2471 Robot is the only one with the intake sphere shown by default. For consistency I think it makes sense to either shown on all or hide on all.
- Team 13603 (2020 FTC) doesn't seem to be able to turn for me. I know that we probably want to have at least one FTC bot, but I also don't think we should have a robot that we don't support (almost all FTC bots are mecanum)

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -863,29 +863,70 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+                    ...editor.getUserData("synthesis:field_preferences"),
+                }
+            } else if (!PreferencesSystem.hasRobotPreferences(this.assemblyHash)) {
+                this._robotPreferences = defaultRobotPreferences()
+                editor.migrateDevtoolRobotData(this._robotPreferences)
+                this._robotPreferences = {
+                    ...this._robotPreferences,
+                    ...editor.getUserData("synthesis:robot_preferences"),
+                }
+            }
```
PepperLola: What happens here if there's a field that already has preferences? would it reach the second case and set its `_robotPreferences`?

fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -241,7 +176,21 @@ class PreferencesSystem {
...
         try {
-            this._preferences = { ...defaultGlobalPreferences, ...JSON.parse(loadedPrefs) }
+            const saved: Preferences & UserPreferences = JSON.parse(loadedPrefs)
+            saved[USER_PREFERENCE_KEY] ??= defaultUserPreferences
+            let didMigrate = false
+            for (const key in defaultUserPreferences) {
+                const typedKey = key as UserPreference
+                if (key in saved) {
+                    didMigrate = true
+                    ;(saved[USER_PREFERENCE_KEY] as Record<UserPreference, unknown>)[typedKey] = saved[typedKey]
```
PepperLola: I might be wrong but wouldn't this mutate `defaultUserPreferences`?

fission/src/ui/helpers/UIProviderHelpers.ts
```diff
@@ -85,13 +85,13 @@ export type OpenModalFn = <T, P>(
...
     props?: Omit<ModalProps<P>, "type" | "configured" | "custom"> & Omit<UIScreenCallbacks<T>, "onBeforeAccept">
-) => string
+) => void
 export type OpenPanelFn = <T, P>(
     content: FunctionComponent<PanelImplProps<T, P>>,
     customProps: P,
     parent?: UIScreen<any, any>,
     props?: Omit<PanelProps<P>, "type" | "configured" | "custom"> & Omit<UIScreenCallbacks<T>, "onBeforeAccept">
-) => string
+) => void
```
PepperLola: Why this change? I remember there was a reason I had it return the IDs, though I can't think of anything specific now. If it's gone unused thus far then it's obviously not necessary and I'm fine with removing it.

PepperLola (commented):
See other comments. How well does this work with the brains system? It looks like most of those reference the assemblyName instead of the hash like your changes here do. The only place I can find where this might actually cause an issue is with simMap since it's keyed by assemblyName, but since the brains hold assemblyName I think the whole thing is worth a look. SwerveDriveBehavior also finds objects in the scene based on assemblyName, so that could potentially also be an issue?

fission/src/ui/helpers/UIProviderHelpers.ts
rutmanz: This was me testing for Dhruv's PR, I will revert

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
```diff
@@ -399,18 +398,18 @@ class SynthesisBrain extends Brain {
         ) as SliderStimulus[]
 
         for (let i = 0; i < sliderDrivers.length; i++) {
-            let sequentialConfig = PreferencesSystem.getRobotPreferences(this._assemblyName).sequentialConfig?.find(
-                sc => sc.jointIndex == this._currentJointIndex
-            )
+            let sequentialConfig = PreferencesSystem.getRobotPreferences(
+                this._assembly.assemblyHash
+            ).sequentialConfig?.find(sc => sc.jointIndex == this._currentJointIndex)
```
Dhruv-0-Arora, line 400: I think this causes some data loss, right? Existing users would lose saved configs. It shouldn't be important though.

fission/src/mirabuf/FieldMiraEditor.ts
Dhruv-0-Arora: Out of scope but I don't think this should exist here like this. It should not be in prod stuff like this so either in test or something generic like
<K extends keyof DevtoolMiraData | (string & {})>

fission/src/systems/preferences/PreferencesSystem.ts
Dhruv-0-Arora: Yes, I noticed that too. I think a solution would be this on like 180

```ts
saved[USER_PREFERENCE_KEY] ??= { ...defaultUserPreferences }.
```

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -863,29 +863,70 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+        if (parts) {
+            const editor = new FieldMiraEditor(parts)
+            if (this.miraType === MiraType.FIELD && !PreferencesSystem.hasFieldPreferences(this.assemblyHash)) {
+                this._fieldPreferences = defaultFieldPreferences()
+                editor.migrateDevtoolFieldData(this._fieldPreferences)
+                this._fieldPreferences = {
+                    ...this._fieldPreferences,
+                    ...editor.getUserData("synthesis:field_preferences"),
+                }
+            } else if (!PreferencesSystem.hasRobotPreferences(this.assemblyHash)) {
```
Dhruv-0-Arora:
```suggestion
            } else if (this.miraType === MiraType.ROBOT && !PreferencesSystem.hasRobotPreferences(this.assemblyHash)) {
```
I think this is important since if a field that has preferences already is spawned, it will skip the if statement and fall here otherwise.

Dhruv-0-Arora (changes requested):
❄️

fission/src/mirabuf/FieldMiraEditor.ts
rutmanz: I mean it's compiled out but I agree generally, I'll update the tests to just use real keys

fission/src/systems/simulation/synthesis_brain/SynthesisBrain.ts
rutmanz: Yes, I migrated user preferences but not robot prefs with this system. I think probably not important but we can revisit if someone wants

Dhruv-0-Arora: 🥹 I think its not important

rutmanz:
I have updated the default assets per @AlexD717's feedback, and transitioned the preferences system to be based on the formerly unused UUID field of the Mirabuf. The exporter just sets the UUID to match the asset name, this system migrates old UUIDs to be based on the asset hash (such that if someone keeps importing the same unmigrated model, it keeps getting the same uuid and retains its preferences). All default assets have also been updated to have their UUIDs set and their filenames match their metadata names. This should also address @PepperLola's concern about the brains

I'm going to add back the developer tool panel but stop validating inputs because that's very annoying and it should only be visible to developers

fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -241,7 +176,21 @@ class PreferencesSystem {
         }
 
         try {
-            this._preferences = { ...defaultGlobalPreferences, ...JSON.parse(loadedPrefs) }
+            const saved: Preferences & UserPreferences = JSON.parse(loadedPrefs)
+            saved[USER_PREFERENCE_KEY] ??= defaultUserPreferences()
+            let didMigrate = false
+            for (const key in defaultUserPreferences) {
```
Dhruv-0-Arora:
```suggestion
            for (const key in defaultUserPreferences()) {
```

Dhruv-0-Arora (changes requested):

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -188,16 +192,20 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
-        assemblyName: string,
-        progressHandle?: ProgressHandle,
-        multiplayerOwnerId?: string
-    ) {
+    public get assemblyName() {
+        return this.mirabufInstance.parser.assembly.info?.name ?? "Unknown"
+    }
+
+    public get assemblyId() {
+        console.log(this.mirabufInstance.parser.assembly.info)
```
Dhruv-0-Arora:
```suggestion

```

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -863,29 +867,70 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+                    ...editor.getUserData("synthesis:field_preferences"),
+                }
+            } else if (this.miraType === MiraType.ROBOT && !PreferencesSystem.hasRobotPreferences(this.assemblyId)) {
+                this._robotPreferences = defaultRobotPreferences()
+                editor.migrateDevtoolRobotData(this._robotPreferences)
+                this._robotPreferences = {
+                    ...this._robotPreferences,
+                    ...editor.getUserData("synthesis:robot_preferences"),
+                }
+            }
```
azaleacolburn, line 916: This is pedantic, but:
1. I think using the `satisfies` operator on every object literal would be a good practice.
2. This section is hard for me to read for some reason and I think some whitespace would be helpful, especially before the preferences objects get created.

azaleacolburn: Also, I don't like the amount of repeated structure within this function and within `resetPreferences()`.

I could imagine a really cool solution where some function matches a set of functions, properties, and strings against an enum and then always runs the same structure of statements just with those variable arguments and functions. However, this would probably be unreadable. 

If you have any (better) ideas about how to reduce repeated structure here, I would appreciate it.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -1125,17 +1168,38 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+
+        const cacheInfo = await MirabufCachingService.storeAssemblyInCache(assembly, {
+            miraType: parser.assembly.dynamic ? MiraType.ROBOT : MiraType.FIELD,
+            name: parser.assembly.info?.name ?? "Unknown",
+        })
+
+        if (cacheInfo == null) {
+            globalAddToast("warning", "Migration Error", "Importing failed to save")
+        }
+    }
```
azaleacolburn, line 1183: This probably deserves its own function (you can include or exclude the `if` statement in that.

fission/src/systems/preferences/PreferencesSystem.ts
```diff
@@ -241,7 +176,21 @@ class PreferencesSystem {
...
+            saved[USER_PREFERENCE_KEY] ??= defaultUserPreferences()
+            let didMigrate = false
+            for (const key in defaultUserPreferences()) {
+                const typedKey = key as UserPreference
+                if (key in saved) {
+                    didMigrate = true
+                    ;(saved[USER_PREFERENCE_KEY] as Record<UserPreference, unknown>)[typedKey] = saved[typedKey]
+                    delete saved[typedKey]
+                }
+            }
```
azaleacolburn: This is like the perfect use for a `.filter().map().reduce()`. I unironcally think that would be a lot more readable and clean than this. I think the `reduce` could even be a `forEach` if you just check the length of the `filter`ed and `map`ped list afterwards.

azaleacolburn:
```ts
const keys = Object.keys(defaultUserPreferences())
    .filter(key => key in saved)
    .map(key => key as UserPreference)

keys.forEach(key => {
    const userPreferences = saved[USER_PREFERENCE_KEY] as Record<UserPreference, unknown>
    userPreferences[key] = saved[key]

    delete saved[key]
})

this._preferences = saved
if (keys.length > 0) {
    this.savePreferences()
}
```

fission/src/ui/panels/configuring/assembly-config/configure/AssemblyExport.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+            const a = document.createElement("a")
+            a.href = url
+            a.download = filename
+            document.body.appendChild(a)
+            a.click()
+            setTimeout(() => {
+                document.body.removeChild(a)
+                URL.revokeObjectURL(url)
+            }, 0)
```
azaleacolburn: If we must do this, we should extract it into its own function in its own file.

fission/src/ui/panels/configuring/assembly-config/interfaces/MetadataConfigInterface.tsx
```diff
@@ -0,0 +1,30 @@
...
+type MetadataConfigInterfaceProps = {
+    selectedAssembly: MirabufSceneObject
+}
+
+const MetadataConfigInterface: React.FC<MetadataConfigInterfaceProps> = ({ selectedAssembly }) => {
+    const [name, setName] = useState<string>(selectedAssembly.mirabufInstance.parser.assembly.info?.name ?? "Unknown")
+    useEffect(() => {
+        selectedAssembly.mirabufInstance.parser.assembly.info ??= {}
+        selectedAssembly.mirabufInstance.parser.assembly.info.name = name
+    }, [name, selectedAssembly.mirabufInstance.parser.assembly, selectedAssembly.mirabufInstance.parser.assembly.info])
```
azaleacolburn: Creating a temporary variable to hold the `...parser.assembly` property might be nice from a verbosity perspective.

fission/src/ui/panels/DeveloperToolPanel.tsx
```diff
@@ -83,63 +58,28 @@ const DeveloperToolPanel: React.FC<PanelImplProps<void, void>> = ({ panel }) =>
...
-            //     return
-            // }
+            setKeys(editor.getAllKeys())
 
-            devtoolHandlers[selectedKey].set(activeObj, parsed)
-            PreferencesSystem.savePreferences?.()
+            devtoolHandlers[selectedKey]?.set(activeObj, parsed)
         } catch (_e) {
             setError("Invalid JSON")
         }
```
azaleacolburn: We should probably shrink the scope of the `try` block here to just be around the parsing (or whatever the minimum scope is that can throw an error). 

In reality, we should probably have a utility function that wrappers json parsing and returns some error value instead.

fission/src/util/Utility.ts
```diff
@@ -37,3 +39,20 @@ export async function hashBuffer(buffer: ArrayBuffer): Promise<string> {
...
         .join("")
 }
+
+export function unzipMira(buff: Uint8Array): Uint8Array {
+    // Check if file is gzipped via magic gzip numbers 31 139
+    if (buff[0] == 31 && buff[1] == 139) {
+        return Pako.ungzip(buff)
+    } else {
+        return buff
+    }
```
azaleacolburn, line 49: Ternary

azaleacolburn (changes requested):
I have a few suggestions, feel free to ignore :3

fission/src/ui/panels/configuring/assembly-config/interfaces/MetadataConfigInterface.tsx
rutmanz: I think it's fine and I'm not totally sure how React would like me doing that. If you feel strongly I can reconsider

fission/src/util/Utility.ts
rutmanz: something feels wrong about using a ternary for an expensive operation. Pako.ungzip isn't just a value, it's a process, and putting it in a ternary simplifies that I think

fission/src/systems/preferences/PreferencesSystem.ts
rutmanz: I agree it would be cleaner and I am doing it but does TS optimize away the type-only map?

fission/src/mirabuf/MirabufSceneObject.ts
rutmanz: I agree it could be better but the typing is really obnoxious and I would prefer to make that refactor a separate ticket if you think it's important

azaleacolburn (approved):
woaw

Dhruv-0-Arora (approved):
🔥

AlexD717 (changes requested):
If I add a scoring/protected zone, then click reset (in the configure panel), it doesn't immediately remove that added zone until I either respawn the asset again or enter into the configuration menu.

The toast popup should either tell users they need to respawn it for changes to take effect, or it should immediately update.

AlexD717 (approved):
lgtm

PR #1372, opened 2026-07-01
Spawn Intake/Ejector in the middle of the robot `[SYNTH-204]`

RoushilS:
## Task

SYNTH-204

Make intake/ejector config intially spawn in the center.

## Symptom

When Intake/Ejector is spawned for the first time, it is often distanced from the robot itself (sometimes underground), making the user configuration period take slightly longer. 

## Solution

Set gizmo to spawn at center after doing a check to make sure the transformation delta is not the intial one.

## Verification

---

- [ ] Ejector and Intake config spawn in the middle of the robot
- [ ] This only happens the first time.

- [ ] User-facing polish:

### DIFF
--- a/fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
+++ b/fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
@@ -23,9 +23,8 @@ import {
     convertThreeMatrix4ToArray,
 } from "@/util/TypeConversions"
 
-const IDENTITY_MATRIX = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]
 function isDefaultDelta(delta: number[]): boolean {
-    return delta.length === IDENTITY_MATRIX.length && delta.every((v, i) => v === IDENTITY_MATRIX[i])
+    return !delta || delta.length === 0
 }
 
 // slider constants
--- a/fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
+++ b/fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
@@ -23,6 +23,11 @@ import {
     convertThreeMatrix4ToArray,
 } from "@/util/TypeConversions"
 
+const IDENTITY_MATRIX = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]
+function isDefaultDelta(delta: number[]): boolean {
+    return delta.length === IDENTITY_MATRIX.length && delta.every((v, i) => v === IDENTITY_MATRIX[i])
+}
+
 // slider constants
 const MIN_ZONE_SIZE = 0.1
 const MAX_ZONE_SIZE = 1.0
@@ -163,6 +168,11 @@ const ConfigureGamepiecePickupInterface: React.FC<ConfigPickupProps> = ({ select
                 )
                 const gizmoTransformation = deltaTransformation.premultiply(robotTransformation)
 
+                // / Don't set to center if not default delta, as this will override the user's configuration
+                if (isDefaultDelta(selectedRobot.intakePreferences!.deltaTransformation)) {
+                    gizmoTransformation.setPosition(selectedRobot.getCenter())
+                }
+
                 gizmo.setTransform(gizmoTransformation)
             }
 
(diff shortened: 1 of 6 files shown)


rutmanz (changes requested):
I really don't like the idea of comparing the config to the default to see if it's been user-modified. Perhaps instead make the default array for the robot preferences empty or otherwise qualitatively different to make it clear it hasn't yet been configured, and use your bounding box computation then.

fission/src/util/TypeConversions.ts
```diff
@@ -15,15 +15,29 @@ export function convertThreeToJoltQuat(a: THREE.Euler | THREE.Quaternion | undef
 }
 
 export function convertArrayToThreeMatrix4(arr: number[]) {
-    // DO NOT ask me why retrieving and setting the same EXACT data is done is two DIFFERENT majors
-    // biome-ignore-start format: We would prefer to visualize this as a matrix
```
rutmanz, line 19: This is still true, the old formatting made it much clearer how the matrix works

fission/src/util/TypeConversions.ts
```diff
@@ -15,15 +15,29 @@ export function convertThreeToJoltQuat(a: THREE.Euler | THREE.Quaternion | undef
...
+    // return identity matrix if array is not valid
+    if (!arr || arr.length !== 16) {
+        return new THREE.Matrix4()
+    }
```
rutmanz: I mean I suppose this works, but I was thinking that when the robot preferences are loaded or when the intake zone prefs are accessed it automatically replace it with the centered position.

That would also have the effect of making the zone position be centered even before they open the intake config panel which isn't nice just for consistency.

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureGamepiecePickupInterface.tsx
```diff
@@ -23,6 +23,10 @@ import {
     convertThreeMatrix4ToArray,
 } from "@/util/TypeConversions"
 
+function isDefaultDelta(delta: number[]): boolean {
+    return !delta || delta.length === 0
```
rutmanz: This isn't necessarily a problem but `!delta` will always be false (empty and filled arrays are truthy[*](https://stackoverflow.com/questions/5491605/empty-arrays-seem-to-equal-true-and-false-at-the-same-time)). Unless you're trying to do a null check in which case you should update the method type signature

https://dorey.github.io/JavaScript-Equality-Table/

rutmanz (changes requested):

fission/src/util/TypeConversions.ts
RoushilS: Oh okay, I'll revert it.

fission/src/util/TypeConversions.ts
RoushilS: What does the last sentence mean?

RoushilS: I just want to be sure I'm understanding you.

rutmanz: Basically I don't think it makes sense to have the intake be in one spot before you configure it and a different spot after if you don't move it. Right now it's possible (I think, I'm not able to run it right now) to intake before you configure the intake and the zone is just wherever the identity matrix puts it.

Alternatively, it might make sense to just return early from all the intake and ejector-related functions (the collision checker / button handlers) if the transform array is empty, or even have an explicit "isConfigured" flag in each subsystem so that they don't do anything until someone actually places them somewhere

rutmanz:
It looks like if you press cancel while editing either intake or ejector, both of them reset to this weird position. If you press save instead on first edit it works fine
image

rutmanz (changes requested):
(see above)

rutmanz (commented):
Uh I mean it seems like this works but it's more addressing the symptom than the cause. I'm realizing that #1371 makes handling this much easier, but basically I think the settings need to be saved after `this.endureDefaultZoneTransformations` is called if things are modified. You could either do that with PreferencesSystem or wait for #1371 to get merged (which might take a while)

rutmanz (changes requested):
Problem appears to still be occuring

https://github.com/user-attachments/assets/fc489d6a-4586-44d8-b77d-cd4df26b3a35

RoushilS:
hmm. Is there any chance you could send me that robot? And also test on the other robots, as they all work for me.

AlexD717 (changes requested):
I was able to reproduce @rutmanz issue. The robot being used was `Team 6443 (2019)_v8` (available on this older version of the assetpack with `bun run assetpack`) however any robot not pre configured seems to have this issue. This only happens when going from the MainHUD configure panel, not the right click configure option.

https://github.com/user-attachments/assets/d917e107-bc7f-4f82-be66-f410ef25d816

PR #1360, opened 2026-06-29
[SYNTH-57] Code Sim Gyro and Accelerometer

PepperLola:
SYNTH-57

Previously, the accelerometer outputs didn't really make sense and the gyros didn't send any data to the Java simulator when running the sample projects.

Note: I used Claude for the math/physics; the values it's outputting make sense and seem consistent with what happens in Fission and the math makes sense to me, but that's probably what could benefit from a review the most.

## Verification

Testing information can be found in the code sim file under `docs/`.

---

Before merging, ensure the following criteria are met:

- [x] All acceptance criteria outlined in the ticket are met.
- [x] Necessary test cases have been added and updated.
- [x] A feature toggle or safe disable path has been added (if applicable).
- [x] User-facing polish:
  - Ask: *"Is this ready-looking?"*
- [x] Cross-linking between Jira and GitHub:
  - PR links to the relevant Jira issue.
  - Jira ticket has a comment referencing this PR.

PR #1373, opened 2026-07-01
Add support for basic camera streams [SYNTH-10]

PepperLola:
## Task

SYNTH-10

Added the ability to add cameras to the robot and both preview their streams from within Fission and stream them to the robot code via WebSocket. Since HALSim doesn't support cameras, this involves spinning up another WebSocket server on the robot side and streaming the video feed to it from Fission.

## Verification

Spawn a robot in Fission and set the brain to the WPILib one. Configure cameras on the robot (within robot configuration - place camera, edit settings). Set up code sim by following the document added in #1360 and keep track of the mean brightness (calculated on the robot code side) in the simulator.

---

- [ ] User-facing polish:


AlexD717 (changes requested):
Overall functionality seems pretty good. I am able to add a camera and view its stream. 

Image

However, when I try and edit the camera FPS, width, and height, it doesn't seem to update. The field of view does update.

In addition, the frame received value alternates every frame from false to true (this is very minor so don't bother fixing if its hard).

PepperLola:
> However, when I try and edit the camera FPS, width, and height, it doesn't seem to update. The field of view does update.

Do you mean update them in Fission? or through Shuffleboard? I'll make sure that works when updating in Fission, but I don't plan to support that in Shuffleboard (i.e., the FPS and Resolution inputs in the camera widget).

> In addition, the frame received value alternates every frame from false to true (this is very minor so don't bother fixing if its hard).

This is all on the side of the robot code and is intended behavior, but I could still change it if it's misleading from a code sample point of view. The robot won't receive frames at the same rate as its periodic methods are called, and the frame received value is true only for a tick where there is a new frame. If the robot is updating at 50tps but the camera is streaming at 30fps, even under nominal network conditions they wouldn't line up and you'd end up with some ticks where there is "no frame received" even though everything's working properly. If it makes more sense to remove this from the sample or replace it with a different metric then I can do that.

AlexD717:
> > However, when I try and edit the camera FPS, width, and height, it doesn't seem to update. The field of view does update.
> 
> Do you mean update them in Fission? or through Shuffleboard? I'll make sure that works when updating in Fission, but I don't plan to support that in Shuffleboard (i.e., the FPS and Resolution inputs in the camera widget).
> 

I meant in fission.

> > In addition, the frame received value alternates every frame from false to true (this is very minor so don't bother fixing if its hard).
> 
> This is all on the side of the robot code and is intended behavior, but I could still change it if it's misleading from a code sample point of view. The robot won't receive frames at the same rate as its periodic methods are called, and the frame received value is true only for a tick where there is a new frame. If the robot is updating at 50tps but the camera is streaming at 30fps, even under nominal network conditions they wouldn't line up and you'd end up with some ticks where there is "no frame received" even though everything's working properly. If it makes more sense to remove this from the sample or replace it with a different metric then I can do that.

I wouldn't bother, this was just a minor thing I noticed.

rutmanz:
Doesn't seem to support multiple camera streams on the dashboards

Dhruv-0-Arora (dismissed):
multiple camera views in shuffleboard.
Only one camera currently shows in shuffleboard despite when you have multiple spawned

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -625,6 +646,21 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+        this._cameras.forEach(c => World.sceneRenderer.removeSceneObject(c.id))
+        this._cameras = []
+
+        if (this.miraType !== MiraType.ROBOT) return
+
+        for (const camPref of this.cameraPreferences) {
+            const camera = new RobotCameraSceneObject(this, camPref)
+            this._cameras.push(camera)
+            World.sceneRenderer.registerSceneObject(camera)
+        }
```
azaleacolburn: Would this be better if it used map

fission/src/mirabuf/RobotCameraSceneObject.ts
```diff
@@ -0,0 +1,200 @@
...
+                for (let i = 0; i < base64.length; i++) bytes[i] = base64.charCodeAt(i)
+                sendCameraFrame(device, bytes)
+            }
+        } catch (e) {
+            console.error(`Camera capture failed for '${device}'`, e)
+        } finally {
+            renderer.setRenderTarget(prevTarget)
+            renderer.autoClear = prevAutoClear
+        }
+    }
```
azaleacolburn, line 179: What is the minimum scope that this try-catch needs to have?

fission/src/mirabuf/RobotCameraSceneObject.ts
```diff
@@ -0,0 +1,200 @@
...
+        this._timeSinceCapture += World.currentDeltaT
+        if (this._timeSinceCapture < 1 / fps) return
+        this._timeSinceCapture = 0
+
+        const parentBody = World.physicsSystem.getBody(this._parentBodyId)
+        if (!parentBody) return
+        const worldTransform = this._worldTransform
+            .copy(this._deltaTransformation)
+            .premultiply(convertJoltMat44ToThreeMatrix4(parentBody.GetWorldTransform()))
+            .multiply(FORWARD_FLIP)
```
azaleacolburn, line 147:
```suggestion
            .multiply(FORWARD_FLIP)
            
```

fission/src/mirabuf/RobotCameraSceneObject.ts
```diff
@@ -0,0 +1,200 @@
...
+        } finally {
+            renderer.setRenderTarget(prevTarget)
+            renderer.autoClear = prevAutoClear
+        }
+    }
+
+    // GL pixels are bottom-up; flip rows into the top-down ImageData
+    private flipInto(target: ImageData, source: Uint8Array): void {
+        const w = this._width
+        const h = this._height
```
azaleacolburn: We can just call this height, or just reference the variable itself

fission/src/systems/simulation/wpilib_brain/CameraFrameSocket.ts
```diff
@@ -0,0 +1,46 @@
...
+        ws.addEventListener("close", () => {
+            if (socket === ws) socket = undefined
+        })
+        ws.addEventListener("error", () => {
+            if (socket === ws) socket = undefined
+        })
+        socket = ws
+    } catch {
+        socket = undefined
+    }
```
azaleacolburn, line 27: minimum scope?

azaleacolburn (dismissed):

azaleacolburn:
Really incredible work here! I just had a few nitpicks.

AlexD717 (changes requested):
Could we change the camera UI to look more like the field scoring/protected zones? That way its standardized across our UI instead of having the top buttons for switching cameras.

Image

AlexD717 (changes requested):
test slack notification

AlexD717 (approved):

fission/src/mirabuf/RobotCameraSceneObject.ts
PepperLola: as far as I can tell, everything from `setRenderTarget` to `sendCameraFrame` can throw an exception, so the only think I could move out is the `renderer.autoClear = true`

PepperLola:
> Could we change the camera UI to look more like the field scoring/protected zones? That way its standardized across our UI instead of having the top buttons for switching cameras.

I believe I've refactored it appropriately, though I still plan to improve the styling and test further. If you want you can take a look and see if it works well. Note that when adding or deleting cameras with the camera preview panel open, the canvases will turn black until you save and the physics system is unheld again (I think this is the cause but I will investigate more).

simulation/SyntheSimJava/src/main/java/com/autodesk/synthesis/cscore/CameraServer.java
```diff
@@ -0,0 +1,372 @@
+// Copyright (c) 2009-2026 FIRST and other WPILib contributors All rights reserved.
```
AlexD717, line 1: Should probably mention this in NOTICE.txt as well

AlexD717 (approved):
Some minor UI stuff that I think could be further improved. Up to you if you want to fix them here, or just create another ticket for that (since this ticket was primarily for codesim support).

In our other systems, pressing add new opens that configuration option for the newly added item directly, while for the USB camera panel it just adds a new camera (slight inconsistency, up to you if you want to fix it).

In addition, ideally right clicking on a robot and selecting "preview camera" should open the corresponding robots preview camera (doesn't do that with multiple robots).

simulation/SyntheSimJava/src/main/java/com/autodesk/synthesis/cscore/CameraServer.java
PepperLola: Their license only requires that their copyright notice/this excerpt from their license be included in files that use their code. I figured since it's just this (maybe one other?) file, there's no need to mention it elsewhere?

AlexD717: From a legal perspective, you are probably right, but I also believe it is just considered best practice, that way someone only has to look in one place to see all the licenses. Just something like these files are licensed from WPILib, refer to their header for more information should be fine.

PepperLola: Added a mention of the files to the NOTICE.txt

PepperLola:
@AlexD717 creating a new camera now automatically edits it, and the preview panel now shows only the cameras on the robot that you right clicked on.

AlexD717 (commented):
If you have multiple robots spawned with the same camera name (ex USB Camera 0), Elastic gets confused and sends one frame from one camera and the next from the other resulting in a very bug experience. Just adding a toast warning saying "Having multiple cameras with the same name can lead to unexpected behavior" should be enough. Though since fission handles it fine, you might just trust the people using codesim to figure it out themselves.

Dhruv-0-Arora (approved):

AlexD717 (approved):
Are we going to go through and add top bar icons for all missing configuration options in a later ticket (ex this introduces USB Cameras, one of my PRs introduces robot spawn positions)?

PR #1374, opened 2026-07-02
Merging joints configuration panels `[SYNTH-205]`

Dhruv-0-Arora:
Merging joints panels

SYNTH-205

Screenshot 2026-07-02 at 10 50 35 AM

Screenshot 2026-07-02 at 10 50 49 AM

---

- [ ] User-facing polish:


RoushilS (changes requested):
It works but there is a bit of a mismatch between the UI that's slightly off putting. I don't think the dropdown really adds anything since you won't see it unless you scroll down anyway. 

So maybe remove the dropdown div as well as the black square. And then the format of Joint Sequencing should be the same as Motor configuration.

Dhruv-0-Arora:
I personally like the way it looks. What is the consensus among others?

rutmanz:
Fwiw I agree with Roushil that it feels quite out of place. Especially since the joint config panel names them as they're named from the assembly and the joint sequencing panel names them by their index

PR #1375, opened 2026-07-02
Refactor Constraint Creation `[SYNTH-169]`

azaleacolburn:
## Task

SYNTH-169

Originally, the purpose of this PR was to make sure all the memory allocated by the `createBallConstraint` function were properly managed, which I neglected in SYNTH-132, because I didn't understand most of the logic in the function and was afraid of breaking things.

On further inspection, it turned out that everything was already properly managed, mostly because the lion's share of the allocation were in bodies, which don't actually need to be freed.

Given this, I turned my attention to refactoring and cleaning up the constraint creation functions in `PhysicsSystem.ts`.

## Symptom

The constraint creation functions in `PhysicsSystem.ts` were really confusing, with lots of out-of-order and/or copy-pasted logic.

## Solution

I refactored the constraint functions as much as I could, using all the standard techniques.

## Verification

All tests still pass.

---

- [x] User-facing polish:

### DIFF
--- a/fission/src/systems/physics/ConstraintSettingsUtilities.ts
+++ b/fission/src/systems/physics/ConstraintSettingsUtilities.ts
@@ -41,11 +41,16 @@ function tryGetPerpendicular(vec: Jolt.Vec3, toCheck: Jolt.Vec3): Jolt.Vec3 | un
     if (Math.abs(Math.abs(vec.Dot(toCheck)) - 1.0) < 0.0001) return undefined
 
     const a = vec.Dot(toCheck)
-    return new JOLT.Vec3(
+    const original = new JOLT.Vec3(
         toCheck.GetX() - vec.GetX() * a,
         toCheck.GetY() - vec.GetY() * a,
         toCheck.GetZ() - vec.GetZ() * a
-    ).Normalized()
+    )
+
+    const perp = original.Normalized()
+    JOLT.destroy(original)
+
+    return perp
 }
 
 export function getPerpendicular(vec: Jolt.Vec3): Jolt.Vec3 {
@@ -63,7 +68,8 @@ export function getPerpendicular(vec: Jolt.Vec3): Jolt.Vec3 {
 export function getAxis(freedom: mirabuf.joint.IDOF, versionNum: number = 6): Jolt.Vec3 {
     const miraAxis = freedom.axis! as mirabuf.Vector3
     // No scaling, these are unit vectors
-    const miraAxisX = (versionNum < 5 ? -miraAxis.x : miraAxis.x) ?? 0
+    const x = miraAxis.x ?? 0
+    const miraAxisX = versionNum < 5 ? -x : x
     return new JOLT.Vec3(miraAxisX, miraAxis.y ?? 0, miraAxis.z ?? 0)
 }
 
--- a/fission/src/systems/physics/ConstraintSettingsUtilities.ts
+++ b/fission/src/systems/physics/ConstraintSettingsUtilities.ts
@@ -2,7 +2,7 @@ import Jolt from "@azaleacolburn/jolt-physics"
 import { DOFSpecs } from "./PhysicsSystem"
 import { mirabuf } from "@/proto/mirabuf"
 import JOLT from "@/util/loading/JoltSyncLoader"
-import { convertMirabufVector3ToJoltRVec3 } from "@/util/TypeConversions"
+import { convertMirabufVector3ToJoltRVec3, convertMirabufVector3ToJoltVec3 } from "@/util/TypeConversions"
 
 type LimitSpecs = Omit<DOFSpecs, "friction" | "axis">
 
@@ -129,4 +129,22 @@ export function applySliderLimits(freedom: LimitSpecs, sliderConstraintSettings:
 
     sliderConstraintSettings.mLimitsMax = halfRange
     sliderConstraintSettings.mLimitsMin = -halfRange
+}
+
+export function createDOFSpecs(dofs: mirabuf.joint.IDOF[]): DOFSpecs[] {
+    const axes = dofs
+        .filter(dof => dof.axis)
+        .map(dof => [convertMirabufVector3ToJoltVec3(dof.axis!), dof] as [Jolt.Vec3, mirabuf.joint.IDOF])
+
+    const constraintSpecs: DOFSpecs[] = axes
+        .filter(([_, dof]) => !dof.limits || (dof.limits.upper ?? 0) - (dof.limits.lower ?? 0) > 0.001)
+        .map(([axis, dof]) => {
+            return { ...dof, axis, friction: 0 } satisfies DOFSpecs
+        })
+
+    return constraintSpecs
+}
+
+export function isWheel(jDef: mirabuf.joint.Joint): boolean {
+    return (jDef.info?.name !== "grounded" && (jDef.userData?.data?.wheel ?? "false") === "true") ?? false
 }
--- a/fission/src/systems/physics/ConstraintSettingsUtilities.ts
+++ b/fission/src/systems/physics/ConstraintSettingsUtilities.ts
@@ -1,0 +1,94 @@
+import Jolt from "@azaleacolburn/jolt-physics"
+import { DOFSpecs } from "./PhysicsSystem"
+import { mirabuf } from "@/proto/mirabuf"
+import JOLT from "@/util/loading/JoltSyncLoader"
+
+type LimitSpecs = Omit<DOFSpecs, "friction" | "axis">
+
+function tryGetPerpendicular(vec: Jolt.Vec3, toCheck: Jolt.Vec3): Jolt.Vec3 | undefined {
+    if (Math.abs(Math.abs(vec.Dot(toCheck)) - 1.0) < 0.0001) return undefined
+
+    const a = vec.Dot(toCheck)
+    return new JOLT.Vec3(
+        toCheck.GetX() - vec.GetX() * a,
+        toCheck.GetY() - vec.GetY() * a,
+        toCheck.GetZ() - vec.GetZ() * a
+    ).Normalized()
+}
+
+export function getPerpendicular(vec: Jolt.Vec3): Jolt.Vec3 {
+    const v1 = new JOLT.Vec3(0, 1, 0)
+    const v2 = new JOLT.Vec3(0, 0, 1)
+
+    const perp = tryGetPerpendicular(vec, v1) ?? tryGetPerpendicular(vec, v2)!
+
+    JOLT.destroy(v1)
+    JOLT.destroy(v2)
+
+    return perp
+}
+
+export function getAxis(freedom: mirabuf.joint.IDOF, versionNum: number = 6): Jolt.Vec3 {
+    const miraAxis = freedom.axis! as mirabuf.Vector3
+    // No scaling, these are unit vectors
+    const miraAxisX = (versionNum < 5 ? -miraAxis.x : miraAxis.x) ?? 0
+    return new JOLT.Vec3(miraAxisX, miraAxis.y ?? 0, miraAxis.z ?? 0)
+}
+
+export function setAxes(
+    freedom: mirabuf.joint.IDOF,
+    settings: Jolt.HingeConstraintSettings | Jolt.SliderConstraintSettings,
+    versionNum?: number
+) {
+    const axis = getAxis(freedom, versionNum)
+
+    let constraintAxis: Jolt.Vec3
+    if ("mHingeAxis1" in settings) {
+        constraintAxis = settings.mHingeAxis1 = settings.mHingeAxis2 = axis.Normalized()
+    } else {
+        constraintAxis = settings.mSliderAxis1 = settings.mSliderAxis2 = axis.Normalized()
+    }
+    settings.mNormalAxis1 = settings.mNormalAxis2 = getPerpendicular(constraintAxis)
+
+    JOLT.destroy(axis)
+}
+
+export function applyHingeLimits(freedom: LimitSpecs, hingeSettings: Jolt.HingeConstraintSettings) {
+    // if (!freedom.limits || !freedom.limits?.upper || !freedom.limits?.lower || !freedom.value) return
+    if (!freedom.limits || Math.abs((freedom.limits?.upper ?? 0) - (freedom.limits?.lower ?? 0)) <= 0.001) return
+
+    // Some values that are meant to be exactly PI are perceived as being past it, causing unexpected behavior.
+    // This safety check caps the values to be within [-PI, PI] wth minimal difference in precision.
+    const piSafetyCheck = (v: number) => Math.min(3.14158, Math.max(-3.14158, v))
+
+    const currentPos = piSafetyCheck(freedom.value ?? 0)
+    const upper = piSafetyCheck(freedom.limits.upper ?? 0) - currentPos
+    const lower = piSafetyCheck(freedom.limits.lower ?? 0) - currentPos
+
+    hingeSettings.mLimitsMin = -upper
+    hingeSettings.mLimitsMax = -lower
+}
+
+export function applySliderLimits(freedom: LimitSpecs, sliderConstraintSettings: Jolt.SliderConstraintSettings) {
+    if (!freedom.limits || Math.abs((freedom.limits?.upper ?? 0) - (freedom.limits?.lower ?? 0)) <= 0.001) return
+
+    const currentPos = (freedom.value ?? 0) * 0.01
+    const upper = (freedom.limits!.upper ?? 0) * 0.01 - currentPos
+    const lower = (freedom.limits!.lower ?? 0) * 0.01 - currentPos
+
+    // Calculate mid point
+    const midPoint = (upper + lower) / 2.0
+    const halfRange = Math.abs((upper - lower) / 2.0)
+
+    // Move the anchor points
+    // NOTE
+    // `Add` does modify the "self" vector
+    sliderConstraintSettings.mPoint2.Add(
+        // NOTE
+        // `MulFloat` does not modify its arguments
+        sliderConstraintSettings.mSliderAxis1.MulFloat(midPoint)
+    )
+
+    sliderConstraintSettings.mLimitsMax = halfRange
+    sliderConstraintSettings.mLimitsMin = -halfRange
+}
--- a/fission/src/systems/physics/ConstraintSettingsUtilities.ts
+++ b/fission/src/systems/physics/ConstraintSettingsUtilities.ts
@@ -2,8 +2,40 @@ import Jolt from "@azaleacolburn/jolt-physics"
 import { DOFSpecs } from "./PhysicsSystem"
 import { mirabuf } from "@/proto/mirabuf"
 import JOLT from "@/util/loading/JoltSyncLoader"
+import { convertMirabufVector3ToJoltRVec3 } from "@/util/TypeConversions"
 
 type LimitSpecs = Omit<DOFSpecs, "friction" | "axis">
+
+export function createAnchorPoint(jointInstance: mirabuf.joint.JointInstance, jointDefinition: mirabuf.joint.Joint) {
+    const jointOrigin = jointDefinition.origin
+        ? convertMirabufVector3ToJoltRVec3(jointDefinition.origin)
+        : new JOLT.RVec3(0, 0, 0)
+    // TODO: Offset transformation for robot builder.
+    const jointOriginOffset = jointInstance.offset
+        ? convertMirabufVector3ToJoltRVec3(jointInstance.offset)
+        : new JOLT.RVec3(0, 0, 0)
+
+    const anchorPoint = jointOrigin.AddRVec3(jointOriginOffset)
+
+    JOLT.destroy(jointOrigin)
+    JOLT.destroy(jointOriginOffset)
+
+    return anchorPoint
+}
+
+// Other than `maxTorque`, these controller settings are not being used as of now
+// because `ArcadeDriveBehavior` goes directly to the `WheelDrivers`.
+// `maxTorque` is only used as communication for `WheelDriver` to get maxAcceleration
+export function createVehicleController(maxAcc: number) {
+    const controllerSettings = new JOLT.WheeledVehicleControllerSettings()
+    controllerSettings.mEngine.mMaxTorque = maxAcc
+    controllerSettings.mTransmission.mClutchStrength = 10.0
+    controllerSettings.mTransmission.mGearRatios.clear()
+    controllerSettings.mTransmission.mGearRatios.push_back(2)
+    controllerSettings.mTransmission.mMode = JOLT.ETransmissionMode_Auto
+
+    return controllerSettings
+}
 
 function tryGetPerpendicular(vec: Jolt.Vec3, toCheck: Jolt.Vec3): Jolt.Vec3 | undefined {
     if (Math.abs(Math.abs(vec.Dot(toCheck)) - 1.0) < 0.0001) return undefined
--- a/fission/src/systems/physics/PhysicsSystem.ts
+++ b/fission/src/systems/physics/PhysicsSystem.ts
@@ -458,7 +458,7 @@ class PhysicsSystem extends WorldSystem {
                             ? [bodyA, bodyB]
                             : [bodyB, bodyA]
 
-                        const res = this.createWheelConstraint(
+                        const [fixedConstraint, vehicleConstraint, vehicleListener] = this.createWheelConstraint(
                             jointInst,
                             jDef,
                             maxAcceleration ?? 1.5,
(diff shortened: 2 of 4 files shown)


fission/src/util/TypeConversions.ts
```diff
@@ -115,16 +115,16 @@ export function convertMirabufVector3ToThreeVector3(v: mirabuf.Vector3): THREE.V
     return new THREE.Vector3(v.x / 100.0, v.y / 100.0, v.z / 100.0)
 }
 
-export function convertMirabufVector3ToJoltVec3(v: mirabuf.Vector3): Jolt.Vec3 {
-    return new JOLT.Vec3(v.x / 100.0, v.y / 100.0, v.z / 100.0)
+export function convertMirabufVector3ToJoltVec3(v: mirabuf.Vector3 | mirabuf.IVector3): Jolt.Vec3 {
+    return new JOLT.Vec3(v.x! / 100.0, v.y! / 100.0, v.z! / 100.0)
```
rutmanz: I don't love the non-null assertion, is there a better way to handle it? Also `mirabuf.Vector3` implements `mirabuf.IVector3`.

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -481,6 +492,58 @@ class PhysicsSystem extends WorldSystem {
         })
     }
 
+    private createAnchorPoint(jointInstance: mirabuf.joint.JointInstance, jointDefinition: mirabuf.joint.Joint) {
+        const jointOrigin = jointDefinition.origin
+            ? convertMirabufVector3ToJoltRVec3(jointDefinition.origin as mirabuf.Vector3)
```
rutmanz: Why are you also casting?

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -481,6 +492,58 @@ class PhysicsSystem extends WorldSystem {
...
+
+        const anchorPoint = jointOrigin.AddRVec3(jointOriginOffset)
+
+        JOLT.destroy(jointOrigin)
+        JOLT.destroy(jointOriginOffset)
+
+        return anchorPoint
+    }
+
+    private newConstraint(constraintSettings: GenericConstraintSettings, bodyA: Jolt.Body, bodyB: Jolt.Body) {
```
rutmanz: Because this isn't pure, I might call it `addConstraint` instead to indicate that even if you don't use the returned value it has consequences

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -481,6 +492,58 @@ class PhysicsSystem extends WorldSystem {
...
+
+        JOLT.destroy(constraintSettings)
+
+        return constraint
+    }
+
+    private getAxis(freedom: mirabuf.joint.IDOF, versionNum: number = 6) {
+        const miraAxis = freedom.axis! as mirabuf.Vector3
+        // No scaling, these are unit vectors
+        const miraAxisX = (versionNum < 5 ? -miraAxis.x : miraAxis.x) ?? 0
```
rutmanz: 💀

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -675,163 +696,130 @@ class PhysicsSystem extends WorldSystem {
...
+
+            JOLT.destroy(constraintSpecifications.axis)
+        })
+
         JOLT.destroy(anchorPoint)
-        JOLT.destroy(jointOrigin)
-        JOLT.destroy(jointOriginOffset)
+    }
+
+    private applyHingeLimits(freedom: LimitSpecs, hingeSettings: Jolt.HingeConstraintSettings) {
```
rutmanz: Given that this doesn't reference any class members (as far as I can tell), I think that it should maybe not be a class member. Calling `this.applyHingeLimits()` kind of implies that the hinge limits are being implied to `this`, not the settings

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -675,163 +696,130 @@ class PhysicsSystem extends WorldSystem {
...
+        // Calculate mid point
+        const midPoint = (upper + lower) / 2.0
+        const halfRange = Math.abs((upper - lower) / 2.0)
+
+        // Move the anchor points
+        // NOTE
+        // `Add` does modify the "self" vector
+        sliderConstraintSettings.mPoint2.Add(
+            // NOTE
+            // `MulFloat` does not modify its arguments
```
rutmanz: Classic jolt

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1483,6 +1471,7 @@ class PhysicsSystem extends WorldSystem {
         const body = this._joltBodyInterface.CreateBody(creationSettings)
         this._bodies.push(body.GetID())
 
+        if (destroy) JOLT.destroy(position)
         JOLT.destroy(size)
         JOLT.destroy(rot)
         JOLT.destroy(creationSettings)
```
rutmanz, line 1375: I mean I get why you're only doing a flag for position but maybe be more specific with the parameter name

rutmanz (changes requested):
Uh okay so these are a little pedantic so you don't have to listen to all of them but it's a refactor so I care more

fission/src/util/TypeConversions.ts
azaleacolburn: The alternative would be a null-check. I'm fairly certain that this is a mirabuf skill issue, basically every property in mirabuf is nullable, even though it doesn't make sense to construct a vector with any missing dimensions, so I can't imagine it would ever happen. If a vector like that does exist, it's almost certainly a bug, so crashing isn't the most ridiculous thing. I suppose the alternative would be failing with a toasted error, considering that theoretically a user could try to parse an invalid mirabuf file with fission.

One thing I could do is have all of these functions return undefined on the error path, then just add undefined-checks in every place they're used as well. I don't like any of these solutions though, really, we just shouldn't have vectors with nullable dimensions.

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: I prefer that as well

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: I assume this is a backwards-compatibility thing that we want to keep?

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Would a separate file called something like `ConstraintSettingsUtilities.ts` make sense?

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: It's really bad, actually

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Sure, generally throughout the physics system, there are function that take a `destroy` flag, and what they universally do is destroy any jolt objects that are passed in as arguments.

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Because they were casting before lol

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -499,62 +527,20 @@ class PhysicsSystem extends WorldSystem {
...
-            ? convertMirabufVector3ToJoltRVec3(jointDefinition.origin as mirabuf.Vector3)
-            : new JOLT.RVec3(0, 0, 0)
-        // TODO: Offset transformation for robot builder.
-        const jointOriginOffset = jointInstance.offset
-            ? convertMirabufVector3ToJoltRVec3(jointInstance.offset as mirabuf.Vector3)
-            : new JOLT.RVec3(0, 0, 0)
-
-        // Gets freed with `hingeConstraintSettings`
-        const anchorPoint = jointOrigin.AddRVec3(jointOriginOffset)
+        const anchorPoint = createAnchorPoint(jointInstance, jointDefinition)
```
Dhruv-0-Arora, line 532: wouldn't the vec that is returned (anchorPoint) leak here? It isn't destroyed.

Dhruv-0-Arora: Its in createHingeConstraint, createSliderConstraint, and createWheelConstraint. 
it is properly destroyed in createBallConstraint

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -675,163 +647,90 @@ class PhysicsSystem extends WorldSystem {
...
+
+            this._joltBodyInterface.AddBody(gb.GetID(), JOLT.EActivation_Activate)
+            mechanism.ghostBodies.push(gb.GetID())
+
+            return gb
         }
-        for (let i = 0; i < constraints.length; ++i) {
-            const c = constraints[i]
+
+        const createHingeSettings = (constraint: DOFSpecs) => {
```
Dhruv-0-Arora: This confused me a lot 🫠 
Are you sure its a good idea to have this const fn and also the fn 
```ts
    private createHingeConstraint(
        jointInstance: mirabuf.joint.JointInstance,
        jointDefinition: mirabuf.joint.Joint,
        torque: number,
        bodyA: Jolt.Body,
        bodyB: Jolt.Body,
        versionNum: number
    ): Jolt.Constraint {
```

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -675,163 +647,90 @@ class PhysicsSystem extends WorldSystem {
...
             const hingeSettings = new JOLT.HingeConstraintSettings()
-            hingeSettings.mMaxFrictionTorque = c.friction
-            hingeSettings.mPoint1 = hingeSettings.mPoint2 = convertJoltVec3ToJoltRVec3(anchorPoint, false)
-            hingeSettings.mHingeAxis1 = hingeSettings.mHingeAxis2 = c.axis.Normalized()
+            hingeSettings.mMaxFrictionTorque = constraint.friction
+            hingeSettings.mPoint1 = hingeSettings.mPoint2 = anchorPoint
+
+            const axis = constraint.axis.Normalized()
+            hingeSettings.mHingeAxis1 = hingeSettings.mHingeAxis2 = axis
             hingeSettings.mNormalAxis1 = hingeSettings.mNormalAxis2 = getPerpendicular(hingeSettings.mHingeAxis1)
```
Dhruv-0-Arora, line 698: I believe the original vec from getPerpendicular is being leaked here even though createHingeConstraint is running JOLT.destroy(hingeSettings) 
cuz of Emscripten

Dhruv-0-Arora:
```ts
export function getPerpendicular(vec: Jolt.Vec3): Jolt.Vec3 {
    const v1 = new JOLT.Vec3(0, 1, 0)
    const v2 = new JOLT.Vec3(0, 0, 1)

    const perp = tryGetPerpendicular(vec, v1) ?? tryGetPerpendicular(vec, v2)!

    JOLT.destroy(v1)
    JOLT.destroy(v2)

    return perp
}
```

the perp var here

I might be wrong though

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -675,163 +647,90 @@ class PhysicsSystem extends WorldSystem {
...
+
+        const createHingeSettings = (constraint: DOFSpecs) => {
             const hingeSettings = new JOLT.HingeConstraintSettings()
-            hingeSettings.mMaxFrictionTorque = c.friction
-            hingeSettings.mPoint1 = hingeSettings.mPoint2 = convertJoltVec3ToJoltRVec3(anchorPoint, false)
-            hingeSettings.mHingeAxis1 = hingeSettings.mHingeAxis2 = c.axis.Normalized()
+            hingeSettings.mMaxFrictionTorque = constraint.friction
+            hingeSettings.mPoint1 = hingeSettings.mPoint2 = anchorPoint
+
+            const axis = constraint.axis.Normalized()
```
Dhruv-0-Arora, line 696: this should be destroyed since it creates a new vector

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -640,33 +628,17 @@ class PhysicsSystem extends WorldSystem {
...
+        const axis = getAxis(rotationalFreedom, versionNum).Mul(0.1)
 
         const bounds = bodyWheel.GetShape().GetLocalBounds()
         const radius = (bounds.mMax.GetY() - bounds.mMin.GetY()) / 2.0
 
         const wheelSettings = new JOLT.WheelSettingsWV()
-
-        wheelSettings.mPosition = convertJoltRVec3ToJoltVec3(anchorPoint.AddRVec3(axis.Mul(0.1)))
-
+        wheelSettings.mPosition = convertJoltRVec3ToJoltVec3(anchorPoint.Add(axis))
```
Dhruv-0-Arora, line 641: This isn't being stored in a var so it leaks. The convertJolt... creates new vector and so does .Add(axis)

fission/src/systems/physics/ConstraintSettingsUtilities.ts
```diff
@@ -0,0 +1,126 @@
...
+    JOLT.destroy(v1)
+    JOLT.destroy(v2)
+
+    return perp
+}
+
+export function getAxis(freedom: mirabuf.joint.IDOF, versionNum: number = 6): Jolt.Vec3 {
+    const miraAxis = freedom.axis! as mirabuf.Vector3
+    // No scaling, these are unit vectors
+    const miraAxisX = (versionNum < 5 ? -miraAxis.x : miraAxis.x) ?? 0
```
Dhruv-0-Arora: This can be NaN
```suggestion
    const x = miraAxis.x ?? 0
    const miraAxisX = versionNum < 5 ? -x : x
```

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -675,163 +647,90 @@ class PhysicsSystem extends WorldSystem {
...
-                friction: 0.0,
-                value: rollDof?.value ?? 0,
-                upper: rollDof?.limits ? (rollDof.limits.upper ?? 0) : undefined,
-                lower: rollDof?.limits ? (rollDof.limits.lower ?? 0) : undefined,
+        const axes = dofs.filter(dof => dof.axis).map(dof => convertMirabufVector3ToJoltVec3(dof.axis!))
+        const constraintSpecs: DOFSpecs[] = axes
+            .map((axis, i) => [axis, dofs[i]] as [Jolt.Vec3, mirabuf.joint.IDOF])
+            .filter(([_, dof]) => !dof.limits || (dof.limits.upper ?? 0) - (dof.limits.lower ?? 0) > 0.001)
+            .map(([axis, dof]) => {
+                return { ...dof, axis, friction: 0 } satisfies DOFSpecs
```
Dhruv-0-Arora: I'm not expert but isn't this misaligning the axis? Because axes is created by filtering out DOFs without an axis, its length and indices no longer correspond to the original dofs array. When we map over axes and pair it with dofs[i], won't it mismatch the axes with the wrong DOFs and drop the last ones entirely? I think we should combine the filtering and mapping into a single step instead."
```ts
const constraintSpecs: DOFSpecs[] = dofs
    .filter(dof => dof.axis)
    .filter(dof => !dof.limits || Math.abs((dof.limits.upper ?? 0) - (dof.limits.lower ?? 0)) > 0.001)
    .map(dof => ({
        ...dof,
        axis: convertMirabufVector3ToJoltVec3(dof.axis!),
        friction: 0
    } satisfies DOFSpecs))
```

fission/src/systems/physics/PhysicsSystem.ts
Dhruv-0-Arora: Also my nvim was bugging cuz the two different fn definitions

Dhruv-0-Arora (dismissed):
⛄

fission/src/systems/physics/PhysicsSystem.ts
Dhruv-0-Arora: I mean it not being stored in a var will mean that it is harder to destroy

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -675,163 +647,90 @@ class PhysicsSystem extends WorldSystem {
...
-                axis: yawAxis,
-                friction: 0.0,
-                value: yawDof?.value ?? 0,
-                upper: yawDof?.limits ? (yawDof.limits.upper ?? 0) : undefined,
-                lower: yawDof?.limits ? (yawDof.limits.lower ?? 0) : undefined,
-            })
+        const dofs = jointDefinition.custom?.dofs
+        if (!dofs || dofs.length < 3) {
+            console.warn("Empty degrees-of-freedom in joint definition for ball constraint")
+            return
```
Dhruv-0-Arora, line 676: if you return here, the anchorPoint is never destroyed so it is leaked

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: We can't free it because we set the second point to equal it, at which point it is owned by the constraint settings and should get freed there.

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Um,,,,,,,, maybe this could be named better, but I couldn't think of a better name

azaleacolburn: I mean the functions basically do the same thing, it's just that one operates on `DOFSpecs` while the other operates on a joint instance

azaleacolburn: Like they're basically overloads (but defined in different scopes)

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Erm, I think the `vec` var there gets leaked in one case (somewhere else that isn't mentioned here). 

But I think in the example you showed, the `perp` vector just gets assigned to the normal axes. Is there a place they get overridden and not freed?

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Ya I lowk didn't realize that `Normalized` is stupid like that

azaleacolburn: However, I think you're wrong, both the old `constraint.axis` and the new `axis` both get owned by the `hingeSettings` object

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: Wait what isn't? `anchorPoint.Add(axis)`? Because that value gets freed by `convertJoltRVec3ToJoltVec3`

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: I believe you are correct

azaleacolburn: They can't be one step, but the mapping can come first

azaleacolburn: I think this is actually what we want:

```ts
const axes = dofs
	.filter(dof => dof.axis)
	.map(dof => [convertMirabufVector3ToJoltVec3(dof.axis!), dof] as [Jolt.Vec3, mirabuf.joint.IDOF])

const constraintSpecs: DOFSpecs[] = axes
	.filter(([_, dof]) => !dof.limits || (dof.limits.upper ?? 0) - (dof.limits.lower ?? 0) > 0.001)
	.map(([axis, dof]) => {
		return { ...dof, axis, friction: 0 } satisfies DOFSpecs
	})

```

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: I could make it another function called something like: `createHingeConstraintFromDOFSpecs`

azaleacolburn: I would like to keep it as a const function because it nicely captures `bodyStart` and `bodyNext`

fission/src/systems/physics/PhysicsSystem.ts
Dhruv-0-Arora: I was thinking the value returned by convertJoltRVec3ToJoltVec3()

fission/src/systems/physics/PhysicsSystem.ts
Dhruv-0-Arora: I was thinking that since you are doing
```ts
            hingeSettings.mNormalAxis1 = hingeSettings.mNormalAxis2 = getPerpendicular(hingeSettings.mHingeAxis1)
```
the hingeSettings.mNormals... will be copies of the vec from perpendicular so when hingesettings is destroyed later, the vec returned by getPerpendicular is leaked
in other words, I thought that you might need to do like an assignment for 
```ts
 x = getPerpendicular(hingeSettings.mHingeAxis1)
```
and then JOLT.destroy(x)

except perp instead of x

fission/src/util/TypeConversions.ts
rutmanz: I mean if it's just optional for fun and you don't think it's actually possible to end up in that position then the null assertions are probably just fine. I guess there's not really a graceful way of handling it

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -481,6 +499,16 @@ class PhysicsSystem extends WorldSystem {
         })
     }
 
+    private createConstraint(constraintSettings: GenericConstraintSettings, bodyA: Jolt.Body, bodyB: Jolt.Body) {
```
rutmanz, line 502: addConstraint?

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -574,64 +560,66 @@ class PhysicsSystem extends WorldSystem {
...
+    }
 
-        if (
-            prismaticFreedom.limits &&
-            Math.abs((prismaticFreedom.limits.upper ?? 0) - (prismaticFreedom.limits.lower ?? 0)) > 0.001
-        ) {
-            const currentPos = (prismaticFreedom.value ?? 0) * 0.01
-            const upper = (prismaticFreedom.limits.upper ?? 0) * 0.01 - currentPos
-            const lower = (prismaticFreedom.limits.lower ?? 0) * 0.01 - currentPos
+    private createFixedConstraint(bodyMain: Jolt.Body, bodyWheel: Jolt.Body, anchorPoint: Jolt.RVec3) {
```
rutmanz, line 579: addFixedConstraint?

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -574,64 +560,66 @@ class PhysicsSystem extends WorldSystem {
...
 
-            sliderConstraintSettings.mLimitsMax = halfRange
-            sliderConstraintSettings.mLimitsMin = -halfRange
-        }
+        return fixedConstraint
+    }
 
-        sliderConstraintSettings.mMotorSettings.mMaxForceLimit = maxForce
-        sliderConstraintSettings.mMotorSettings.mMinForceLimit = -maxForce
+    private createVehicleConstraint(wheelSettings: Jolt.WheelSettingsWV, bodyMain: Jolt.Body, maxAcc: number) {
```
rutmanz, line 595: addVehicleConstraint?

fission/src/util/TypeConversions.ts
```diff
@@ -111,20 +111,12 @@ export function convertMirabufTransformToThreeMatrix(m: mirabuf.ITransform): THR
     return new THREE.Matrix4().compose(pos, quat, new THREE.Vector3(1, 1, 1))
 }
 
-export function convertMirabufVector3ToThreeVector3(v: mirabuf.Vector3): THREE.Vector3 {
-    return new THREE.Vector3(v.x / 100.0, v.y / 100.0, v.z / 100.0)
+export function convertMirabufVector3ToJoltVec3(v: mirabuf.IVector3): Jolt.Vec3 {
+    return new JOLT.Vec3(v.x! / 100.0, v.y! / 100.0, v.z! / 100.0)
```
rutmanz, line 115: are mirabuf vectors in cm? 💀

rutmanz (changes requested):
really it's just consistency on the add vs create thing. Otherwise lgtm

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: `create` is the prefix the entire physics system already uses for methods that generate and register some physics object. I don't a reason to change that standard in this PR.

azaleacolburn:
Should I just write a standard for what different prefixes mean in the PhysicsSystem? I can do that as well as go through and standardize everything in another PR.

rutmanz (approved):
lgtm

PepperLola (approved):
There are a couple of TODO comments, not sure if you want to address those before merge.

azaleacolburn:
> There are a couple of TODO comments, not sure if you want to address those before merge.

The only TODO that I added in this PR is much better left for another PR

PR #1370, opened 2026-07-01
feat(UI): Mobile UI Redesign `[SYNTH-26]`

Dhruv-0-Arora:
What it sounds like 👍 

SYNTH-26

---

Before merging, ensure the following criteria are met:

- [ ] All acceptance criteria outlined in the ticket are met.
- [ ] Necessary test cases have been added and updated.
- [ ] A feature toggle or safe disable path has been added (if applicable).
- [ ] User-facing polish:
  - Ask: *"Is this ready-looking?"*
- [ ] Cross-linking between Jira and GitHub:
  - PR links to the relevant Jira issue.
  - Jira ticket has a comment referencing this PR.

PR #1329, opened 2026-06-19
Desktop UI redesign `[SYNTH-142]`

Dhruv-0-Arora:
## Task

This task involves implementing the newly designed UI with a Topbar replacing the current MainHUD implementation. In doing so, we aim to elevate the user experience, decrease friction during configuration workflows, and make Synthesis more intuitive.

The mockup design looks as such:

Screenshot 2026-06-18 at 6 27 26 PM

SYNTH-142

> [!NOTE]
> this pull request is being merged into a feature branch which will contain the full UI refactor before it is merged into `dev`

## Solution

A screenshot of the newly implemented topbar looks like:

Screenshot 2026-07-01 at 12 16 25 AM

## Verification

---

Before merging, ensure the following criteria are met:

- [ ] All acceptance criteria outlined in the ticket are met.
- [ ] Necessary test cases have been added and updated.
- [ ] A feature toggle or safe disable path has been added (if applicable).
- [ ] User-facing polish:
  - Ask: *"Is this ready-looking?"*
- [ ] Cross-linking between Jira and GitHub:
  - PR links to the relevant Jira issue.
  - Jira ticket has a comment referencing this PR.

PR #1377, opened 2026-07-02
MainHUD UI Redesign `[SYNTH-63]`

Dhruv-0-Arora:
SYNTH-63

## Task

SYNTH-142

Implementing the newly designed UI with a Topbar replacing the current MainHUD implementation. In doing so, we aim to elevate the user experience, decrease friction during configuration workflows, and make Synthesis more intuitive.

The mockup design looks as such:

Screenshot 2026-06-18 at 6 27 26 PM

A screenshot of the newly implemented topbar looks like:

Screenshot 2026-07-01 at 12 16 25 AM

## Mobile

SYNTH-26

The redesign also supports mobile:

Screenshot 2026-07-02 at 11 29 51 AM

Screenshot 2026-07-02 at 11 30 09 AM

Screenshot 2026-07-02 at 11 30 29 AM

## Joints

Merging the joint panels

SYNTH-205

Screenshot 2026-07-02 at 10 50 35 AM

Screenshot 2026-07-02 at 10 50 49 AM

---

- [ ] User-facing polish:


azaleacolburn:
I have a few critiques of the UI when the vertical side of the screen is longer than the horizontal side (and the vertical side is quite long):
1. For the top-level view, there should be some sort of faint outline around the buttons (see below)
Screenshot 2026-07-02 at 11 30 02 AM
2. For the assembly configuration view:
a) Options for configuration shouldn't appear until I've selected an assembly, or at least they should be greyed out or something, to indicate that they can't be selected. They also just shouldn't be selectable, instead of having a warning toast pop up when they are selected.
b) The options should be spread out vertically, not stacked in a 2x3 grid.
Screenshot 2026-07-02 at 11 30 24 AM

I'm not sure to what extent my critiques hold any weight (I'm not a UX person) nor to what extent they apply to the entire redesign of the HUD (probably the wrong term) broadly or if they just apply to the vertical version (since that is the version I tested). 

P.S. This isn't really a redesign of the entire UI, so you should probably give the PR a a more specifically name.

fission/src/ui/components/topbar/TopBarConfig.ts
azaleacolburn, line 1: Why is this file name `camelCase` instead of being `PascalCase`?

fission/src/ui/components/topbar/TopBarIcons.tsx
```diff
@@ -0,0 +1,120 @@
+import { Box } from "@mui/material"
+import { type FC, useEffect, useMemo, useRef } from "react"
+
+/**
+ * Topbar icons are SVGs inlined into the bundle at build time and rendered as
+ * real DOM `<svg>` nodes (not `<img src>`). This matters: an `<img>` is a
+ * separate, purgeable resource, and the browser will drop its decoded bitmap
+ * under memory pressure — which the APS login popup reliably triggers by loading
+ * a second copy of the app (Jolt WASM + asset pack). The icon then repaints
```
azaleacolburn: Why does the APS login popup load a second version of the entire application? Is this a hallucination, or something that should concern us?

rutmanz:
It looks like #1329, #1374 and #1370 were merged into this branch before all of their review comments were resolved, and I want to make sure those comments don't get lost / people don't have to bring them up twice

Dhruv-0-Arora:
## #1329 (Desktop Topbar) Comments:

> @rutmanz : my biggest concern is that people won't know what the icons mean intuitively and searching through the hover text to find out isn't exactly user-friendly. ... Not really sure what to do about it other than maybe make one of those in-app tutorial things on first open that walks you through all the main ui elements?

This will hopefully be resolved after we implement onboarding. Onboarding will take users through the process of spawning an assembly and configuring it. It will be a separate PR.

> The selector dropdowns have deadzones that make it behave weirdly.

Resolved.

> The scale just seems kind of big? Like compared to what it used to be and the other ui elements, the top bar just feels like it takes up too much space

My preference was keeping it this size. Making it smaller can be something I implement if another person also agrees with @rutmanz 

> I do think the debug and developer panels need to be accessible so we don't regress feature-wise.

Resolved. They only appear in dev environment on the topbar now.

> [!IMPORTANT]
> @rutmanz : Also the UI feels a little stylistically confused between the MUI style and the new top bar, which is really easy to see with the dropdown open.

I agree. When I was working on re-styling the dropdown in the topbar, it made changes throughout other panels. As a result, I think it is best that we make another task of redesigning the way panels currently are styled. I personally think the way the dropdown look on the topbar for now is fine.

image

## #1374 (Joints Panels Merging) Comments

> @RoushilS : It works but there is a bit of a mismatch between the UI that's slightly off putting. I don't think the dropdown really adds anything since you won't see it unless you scroll down anyway.
> 
> So maybe remove the dropdown div as well as the black square. And then the format of Joint Sequencing should be the same as Motor configuration.

> @rutmanz : Fwiw I agree with Roushil that it feels quite out of place. Especially since the joint config panel names them as they're named from the assembly and the joint sequencing panel names them by their index

I will make the changes 👍 

## #1370 (Mobile UI) Comments:

> @AlexD717 : Any assembly that I try to spawn sets it up as not configured with no configure panel opening.

Resolved.

PepperLola (changes requested):
When the screen gets narrow but not enough for it to say to rotate the screen, the buttons end up overflowing off the screen. I think it would make sense to switch to a dropdown menu when this happens so that all the controls can remain accessible.

azaleacolburn (changes requested):

azaleacolburn (approved):

fission/src/test/bootstrap/AppMount.test.tsx
```diff
@@ -53,6 +46,11 @@ describe("React Mounting", async () => {
     test("App fully mounts through main.tsx", async ({ annotate, skip }) => {
         skip(server.browser == "firefox", "WebGL bug in Github Actions on Firefox")
 
+        // The MainMenu/MainHUD "Singleplayer" gate was removed: the app now boots
```
AlexD717: I don't think it makes sense to write comments for what was removed. Especially like in a year or two, this will just be confusing.

fission/src/ui/components/topbar/UseConfigureAssembly.ts
```diff
@@ -0,0 +1,110 @@
...
+]
+
+export const assemblyLabel = (assembly: MirabufSceneObject): string => {
+    if (assembly.miraType !== MiraType.ROBOT) return assembly.assemblyName
+    const scheme =
+        assembly.multiplayerOwnerName ??
+        InputSystem.brainIndexSchemeMap.get((assembly.brain as SynthesisBrain).brainIndex)?.schemeName ??
+        "-"
+    return `[${scheme}] ${assembly.assemblyName}`
+}
```
AlexD717: This is the same functionality as MirabufSceneObject.ts descriptiveName

AlexD717 (changes requested):
Some minor feedback

fission/src/test/bootstrap/AppMount.test.tsx
Dhruv-0-Arora: Good point. I added that earlier just because I couldn't find anything solid to replace it with.

fission/src/ui/components/topbar/UseConfigureAssembly.ts
Dhruv-0-Arora: Thank you!

azaleacolburn (approved):

fission/src/ui/components/TopBar.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+                {/* Divider line */}
+                <Box sx={{ width: "2px", height: 38, bgcolor: "topBarText.main", opacity: 0.4 }} />
+
+                {appMode === "Configure" && <ConfigureControls />}
+                {appMode === "Gameplay" && <GameplayControls />}
+
+                <Box flexGrow={1} />
+
+                {import.meta.env.DEV && (
```
AlexD717, line 143: I guess this is up for debate, but don't we only want the developer tools to be unavailable. Debug tools should still be there as they provide useful features (ex, drag mode).

AlexD717 (approved):
Apart from this minor issue lgtm.

RoushilS (approved):
lgtm!!

fission/src/ui/components/MobileHUD.tsx
```diff
@@ -0,0 +1,227 @@
...
+    const openMultiplayer = () =>
+        openModal(MultiplayerStartModal, {
+            startWorldCallback: async (name: string, room?: string) => {
+                const isHost = room == null
+                const roomId = room ?? Math.random().toString(10).substring(2, 8)
+                PreferencesSystem.setUserPreference("MultiplayerUsername", name)
+                PreferencesSystem.savePreferences()
+                const success = await MultiplayerSystem.setup(roomId, name, isHost)
+                if (success && isHost) globalAddToast("info", "Room Code", roomId)
+                return success
```
rutmanz: this code duplicates https://github.com/Autodesk/synthesis/pull/1377/changes#diff-647e6ad757bf41ac78f2e7aab4c93296579852d417f815fdc275d7a946c47af9R1

rutmanz (changes requested):
## Looks like the navbar doesn't correctly detect when to go into the dropdown

Ellie's screenshot from Slack
image
And my own reproduction
image

## The feedback on the joints panel was never implemented

image

## It's still massive

I know you asked for some other input on this and nobody ever replied, @azaleacolburn agrees it's way too big. To me, about 75% size is about right.

## Sidebar space seems kind of wasted

This is a lot of the screen being covered for six buttons that could fit in a much smaller space. Is there a reason that it's so wide (compared to like 1 column with less vertical spacing)

image

## Scope

The configuration panel has been kind of co-opted but still works, but you can only get there by pressing the back arrow after entering a configuration mode from the main hud or through the context menu or command palette. There are also some configuration modes that don't appear in the main hud. To me this makes it all feel half-done, and maybe just a "more" button would do the trick in the config subpanel

fission/src/ui/helpers/StartMultiplayerWorld.ts
rutmanz, line 1: not a modal

AlexD717:
Note: There is currently no button for the recently added camera positions. Since I don't want this to be stuck in review forever, I am fine merging this and then later fixing the minor issues.

fission/src/ui/helpers/useIsMobile.ts
```diff
@@ -1,10 +1,38 @@
...
+ * portrait tablets, and small/narrow desktop windows get the MobileHUD. The two
+ * conditions are complementary: the width floor covers ordinary narrow viewports,
+ * and the phone-sized check still catches unusually wide-but-short touch devices.
  */
 export function useIsMobile(): boolean {
-    const theme = useTheme()
-    return useMediaQuery(theme.breakpoints.down("lg"))
+    const isTouch = useIsTouchDevice()
+    const isPhoneSized = useMediaQuery("(max-width: 600px), (max-height: 600px)", { noSsr: true })
+    const isTooNarrowForTopBar = useMediaQuery(`(max-width: ${DESKTOP_MIN_WIDTH - 0.02}px)`, { noSsr: true })
```
rutmanz, line 16: bro bro?

rutmanz (changes requested):
we're so close Dhruv "I don't have one" Arora

Also the naming on the joints

fission/src/ui/helpers/useIsMobile.ts
Dhruv-0-Arora: This is due to min-width and max-width being inclusive - causes overlap issue.
not 0.01 because of safari bug.

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureJointsInterface.tsx
```diff
@@ -0,0 +1,257 @@
...
+    }, [lookingForParent, behavior])
+    useEffect(() => {
+        setHasParent(behavior.parentJointIndex !== undefined)
+    })
+
+    return (
+        <Stack direction="row" textAlign="center" gap={1} key={elementKey}>
+            {hasParent && <Spacer width={10} />}
+            <Tooltip
+                title={hasParent ? "Following Joint " + behavior.parentJointIndex : selectable ? "Set as parent" : ""}
```
rutmanz: so this still references the index (which is now not shown anywhere). I personally don't mind making everything use the index and type but it should be consistent.

Image

rutmanz (changes requested):
---
I think it was better to show a sidebar than to show this. I feel like it's reasonable to use synthesis split screen or otherwise in a narrow way and even if it looks a little worse we shouldn't just stop them from using it (well really it should adjust to have an overflow menu imo).

Image

---

It also looks like you missed my comment on accessing the configuration menu. I think that to avoid issues like @AlexD717 brought up with the camera config button being missing, and with the reality that we will not be making icons for or fitting every configuration subpanel into the top bar/mobile hud, we should simply have a "See more" configuration button that opens the ConfigurationPanel, rather than forcing people to click one of the specific buttons and then click the back arrow

fission/src/ui/components/TopBar.tsx
```diff
@@ -0,0 +1,209 @@
...
+            if (v === deobf("Nzk2NTcz")) {
+                const r = (document as any)[deobf("Njc2NTc0NDU2YzY1NmQ2NTZlNzQ0Mjc5NDk2NA==")](deobf("NzI2ZjZmNzQ="))
+                if (r) {
+                    const w = (document as any)[deobf("NjM3MjY1NjE3NDY1NDU2YzY1NmQ2NTZlNzQ=")](
+                        deobf("NmQ2MTcyNzE3NTY1NjU=")
+                    )
+                    r[deobf("NzA2MTcyNjU2ZTc0NGU2ZjY0NjU=")][deobf("Njk2ZTczNjU3Mjc0NDI2NTY2NmY3MjY1")](w, r)
+                    w[deobf("NjE3MDcwNjU2ZTY0NDM2ODY5NmM2NA==")](r)
+                }
+            }
```
0xda157, line 61: can we please use real variable names and add comment(s) explaining what's happening here

fission/src/ui/components/TopBar.tsx
```diff
@@ -0,0 +1,209 @@
...
+    setAddToast(addToast)
+    setOpenPanel(openPanel)
+    setOpenModal(openModal)
+
+    const [userInfo, setUserInfo] = useState(APS.userInfo)
+    const [modeHovered, setModeHovered] = useState(false)
+    const [modeMenuOpen, setModeMenuOpen] = useState(false)
+
+    useEffect(() => {
+        // biome-ignore-start lint/suspicious/noExplicitAny: allow any
```
0xda157: the biome ignore comments should have real explanations

fission/src/ui/components/ViewCube.tsx
```diff
@@ -8,7 +8,7 @@ import World from "@/systems/World"
 
 interface ViewCubeProps {
     size?: number
-    position?: { top?: number; left?: number; right?: number; bottom?: number }
+    position?: { top?: number | string; left?: number | string; right?: number | string; bottom?: number | string }
```
0xda157, line 11: why would we wants these to be strings? position should be represented by a number no?

fission/src/ui/helpers/useIsMobile.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+// min width desktop topbar needs before stuff overflow keep roughly synced with topbar contents
+export const DESKTOP_MIN_WIDTH = 1000
+
+// true when compact mobile hud should replace desktop topbar
+export function useIsMobile(): boolean {
+    const isTouch = useIsTouchDevice()
+    const isPhoneSized = useMediaQuery("(max-width: 600px), (max-height: 600px)", { noSsr: true })
+    // minus 0.02 makes max width mean strictly below 1000 so it still catch fractional widths like 999.5 from zoom or display scaling
```
0xda157: what does this even mean. subtracting `0.02` would not fix this issue

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureJointsInterface.tsx
```diff
@@ -0,0 +1,266 @@
...
+    const [selectable, setSelectable] = useState(false)
+    const [hasParent, setHasParent] = useState(false)
+    useEffect(() => {
+        setSelectable(
+            lookingForParent !== undefined && lookingForParent !== behavior && behavior.parentJointIndex === undefined
+        )
+    }, [lookingForParent, behavior])
+    useEffect(() => {
+        setHasParent(behavior.parentJointIndex !== undefined)
+    })
```
0xda157: why are we running this on ever re-render?

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureJointsInterface.tsx
```diff
@@ -0,0 +1,266 @@
...
+}
+
+function sortBehaviors(behaviors: SequentialBehaviorPreferences[]): SequentialBehaviorPreferences[] {
+    behaviors.sort((a, b) => a.jointIndex - b.jointIndex)
+
+    const sortedBehaviors: SequentialBehaviorPreferences[] = []
+    behaviors.forEach(b => {
+        if (b.parentJointIndex === undefined) sortedBehaviors.push(b)
+    })
+
```
0xda157: why don't we use filter and the spread operator here?

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureJointsInterface.tsx
```diff
@@ -0,0 +1,266 @@
...
+    const [seqBehaviors, setSeqBehaviors] = useState<SequentialBehaviorPreferences[]>(
+        PreferencesSystem.getRobotPreferences(selectedRobot.assemblyName)?.sequentialConfig ??
+            (selectedRobot.brain as SynthesisBrain).behaviors
+                .filter(b => b instanceof SequenceableBehavior)
+                .map(b => defaultSequentialConfig(b.jointIndex, b instanceof GenericArmBehavior ? "Arm" : "Elevator"))
+    )
+    const [lookingForParent, setLookingForParent] = useState<SequentialBehaviorPreferences | undefined>(undefined)
+
+    const [_, update] = useReducer(x => {
+        setSeqBehaviors(sortBehaviors(seqBehaviors))
```
0xda157, line 181: this is probably out of scope of this pr, but we should not be doing this, it defeats the point of using react. every time we update seqBehaviors, we should just call `setSeqBehaviors` instead of mutating

fission/src/ui/components/TopBar.tsx
Dhruv-0-Arora: this is Julian's obfuscated easter egg so I decided I didn't want to touch it. I think he would believe a comment would defeat the purpose :)

fission/src/ui/components/TopBar.tsx
Dhruv-0-Arora: I added one here 👍

fission/src/ui/components/ViewCube.tsx
Dhruv-0-Arora: I made this change to match React's CSSProperties typing so I could use strings like "50%" or "calc()"
```ts
{showViewCube && <ViewCube position={{ top: "calc(20px + var(--top-bar-height, 0px))", right: 20 }} />}
```

fission/src/ui/helpers/useIsMobile.ts
Dhruv-0-Arora: I changed the comment

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureJointsInterface.tsx
Dhruv-0-Arora: Fixed 👍

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureJointsInterface.tsx
Dhruv-0-Arora: I agree. Since its pre-existing code moved in during this PR, I'll leave it for another PR unless you think I should fix it in this one.,

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureJointsInterface.tsx
Dhruv-0-Arora: fixed

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureJointsInterface.tsx
```diff
@@ -48,15 +48,12 @@ const BehaviorCard: React.FC<BehaviorCardProps> = ({
...
-    const [hasParent, setHasParent] = useState(false)
     useEffect(() => {
         setSelectable(
             lookingForParent !== undefined && lookingForParent !== behavior && behavior.parentJointIndex === undefined
         )
     }, [lookingForParent, behavior])
-    useEffect(() => {
-        setHasParent(behavior.parentJointIndex !== undefined)
-    })
+    const hasParent = behavior.parentJointIndex !== undefined
```
0xda157, line 55: this should probably be reactive to behavior changing

Dhruv-0-Arora: discussed - behavior is mutated in place and places that change parentage call update() which recompute the hasParent const

fission/src/ui/components/TopBar.tsx
Dhruv-0-Arora: ideally we have debug tools but I feel that having it where it is currently can be confusing and add clutter. We should have another PR that changes where this is located and then go from there.

fission/src/ui/components/SplitButtonDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+            onClick={e => setAnchorEl(e.currentTarget)}
+            sx={{ ...HALF_SX, px: 0.25 }}
+        >
+            <Box sx={{ fontSize: 18, display: "flex" }}>
+                <SynthesisIcons.DROPDOWN_CARET />
+            </Box>
+        </IconButton>
+    )
+
+    const half = (button: React.ReactNode, tooltip?: string) => {
```
0xda157: this seems like it should be a proper component

fission/src/ui/components/SplitButtonDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+        const wrapped = <span style={{ display: "flex", height: "100%" }}>{button}</span>
+        return tooltip ? <Tooltip title={tooltip}>{wrapped}</Tooltip> : wrapped
+    }
+
+    return (
+        <>
+            <Stack direction="row" alignItems="stretch" sx={{ height: 34, borderRadius: 1, overflow: "hidden", ...sx }}>
+                {half(iconButton, iconTooltip)}
```
0xda157: yeah this is bad

fission/src/ui/components/SplitButtonDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+            <Menu
+                anchorEl={anchorEl}
+                open={Boolean(anchorEl)}
+                onClose={closeMenu}
+                anchorOrigin={{ vertical: "bottom", horizontal: "left" }}
+                transformOrigin={{ vertical: "top", horizontal: "left" }}
+                {...DROPDOWN_MENU_PROPS}
+                {...menuProps}
+            >
+                {renderMenu(closeMenu)}
```
0xda157:
```suggestion
                <renderMenu closeMenu={closeMenu} />
```

fission/src/ui/components/topbar/ConfigureSplitDropdown.tsx
```diff
@@ -0,0 +1,91 @@
...
+    // fields have no brain to configure
+    const menuOnlyConfigs = isField ? MENU_ONLY_CONFIGS.filter(c => c.mode !== ConfigMode.BRAIN) : MENU_ONLY_CONFIGS
+
+    const configMenuItem = (
+        key: string,
+        icon: React.ReactNode,
+        label: string,
+        mode: ConfigMode,
+        closeMenu: () => void
+    ) => (
```
0xda157: don't nest react components!!!

fission/src/ui/components/SplitButtonDropdown.tsx
Dhruv-0-Arora: fixed

fission/src/ui/components/SplitButtonDropdown.tsx
Dhruv-0-Arora: fixed

fission/src/ui/components/SplitButtonDropdown.tsx
Dhruv-0-Arora: fixed

fission/src/ui/components/topbar/ConfigureSplitDropdown.tsx
Dhruv-0-Arora: fixed

RoushilS (approved):
👍

AlexD717 (approved):
Functionality looks good to me. 

Note: There are some merge conflicts

rutmanz (commented):
I do think that if one of the top buttons opens a screen that is already open and on top, it should close it instead of focusing on it again if possible. Could be its own ticket though

rutmanz (approved):
Dhruv,

It is with heavy heart that I submit this review today. It has been weighing on me for some time, and I don't know how to tell you that this code looks acceptable to me.

Regretfully,
Zachary Rutman

PepperLola (approved):

PR #1381, opened 2026-07-06
Hash-mismatch error on mobile `SYNTH-193`

Dhruv-0-Arora:
## Task

When on mobile, you run into a hash mismatch error (toast) when spawning an assembly for the first time (from default assemblies). This is pretty confusing to users.

## Symptom

crypto.subtle can fail in non-ssl environments and overall can be quite strict.
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle

This implementation will not have any impact on production (likely) because we use ssl, so this is more of a nice to have.

SYNTH-193

---

- [ ] User-facing polish:

### DIFF
--- a/fission/src/mirabuf/MirabufLoader.ts
+++ b/fission/src/mirabuf/MirabufLoader.ts
@@ -212,11 +212,15 @@ class MirabufCachingService {
                 fileSize: miraBuff.byteLength,
             })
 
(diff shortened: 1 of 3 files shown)


Dhruv-0-Arora:
This change is not critical so we don't need it

PR #1384, opened 2026-07-07
Fix Cancel Input Behavior `[SYNTH-185]`

RoushilS:
## Task

SYNTH-185

## Symptom

Previously, when an input was changed, and then canceled, the behavior would not revert. There was also an issue where toggling touch controls and joystick wouldn't save.

## Solution

This PR fixes the problem by implementing a new helper method, `rebindOldBrainSchemes`. Previously, the revert was technically built, but the new input object was not set, meaning the old modified one would still persist. This method goes through and sets the `brainIndexSchemeMap` to the new verison for every object. This method is then called on cancel. As for the gamepad toggle issue, `customized = true` is performed for both.

## Verification

Verify by checking that cancel and save behavior works for input changes as well as input scheme changes.

---

- [ ] User-facing polish:


Dhruv-0-Arora (approved):
works 👍 

I would recommend however that you add some ci
perhaps a test such as 
```ts
InputSchemeManager.resetDefaultSchemes()
    InputSchemeManager.rebindOldBrainSchemes()

    const rebound = InputSystem.brainIndexSchemeMap.get([number])!
    expect(rebound).not.toBe([mutated scheme])
    expect(rebound.schemeName).toBe(ernie)
```

fission/src/systems/input/InputSchemeManager.ts
```diff
@@ -83,6 +83,16 @@ class InputSchemeManager {
         EventSystem.dispatch("InputSchemeChanged", { panelId })
     }
 
+    public static rebindOldBrainSchemes() {
+        const schemesByName = new Map(this.allInputSchemes.map(s => [s.schemeName, s] as const))
+        for (const [brainIndex, scheme] of InputSystem.brainIndexSchemeMap) {
+            const reverted = schemesByName.get(scheme.schemeName)
+            if (reverted && reverted !== scheme) {
```
PepperLola: Since `resetDefaultSchemes` happens right before `rebindOldBrainSchemes` in `ConfigurePanel.tsx`, `this.defaultInputSchemes` (the latter part of `allInputSchemes`) would be overwritten with new copies of the default schemes. This means that `reverted !== schemes` will always be true, so there's not really any point checking it. You could instead check whether the scheme has been customized (`scheme.customized`) or just always revert if there's an old version of the scheme present.

RoushilS:
> works 👍
> 
> I would recommend however that you add some ci perhaps a test such as
> 
> ```ts
> InputSchemeManager.resetDefaultSchemes()
>     InputSchemeManager.rebindOldBrainSchemes()
> 
>     const rebound = InputSystem.brainIndexSchemeMap.get([number])!
>     expect(rebound).not.toBe([mutated scheme])
>     expect(rebound.schemeName).toBe(ernie)
> ```

@Dhruv-0-Arora I will look into this 👍

PepperLola (approved):

PR #1386, opened 2026-07-08
Don't directly call React components `[SYNTH-217]`

0xda157:
## Task

SYNTH-217

## Symptom

React components are being called directly which is very bad pratice:

> Components should only be used in JSX. Don’t call them as regular functions. React should call it.
>
> React must decide when your component function is called [during rendering](https://react.dev/reference/rules/components-and-hooks-must-be-pure#how-does-react-run-your-code).
>
> -- https://react.dev/reference/rules/react-calls-components-and-hooks#never-call-component-functions-directly

## Solution

Invoke React components correctly, refactoring some component's props in some cases.

## Verification

`bun run fmt`, `bun run lint` and `bun run build` all succeed 

---

- [x] User-facing polish:


rutmanz (changes requested):
There are a few more that haven't been converted that I found with this regex `\{[A-Z]\w*\(`

rutmanz:
image

Spacing is wonky, also seeing red with a checkmark is confusing, maybe the dark green was better...

rutmanz (changes requested):
(see above)

rutmanz (approved):

AlexD717 (approved):
lgtm

PepperLola (approved):

PR #1387, opened 2026-07-08
Switch All Icons to SynthesisIcon `[SYNTH-202]`

AlexD717:
## Task

SYNTH-202

## Symptom

Currently some icons are imported from "react-icons" and others from StyledComponents (SynthesisIcons), which is inconsistent. 

## Solution

Import all icons from StyledComponents.

## Verification

- [x] No more importing icons from "react-icons" (other than StyledComponents)
- [x] All icons still appear and work as before

---

- [x] User-facing polish:

### DIFF
--- a/fission/src/ui/components/WPILibConnectionStatus.tsx
+++ b/fission/src/ui/components/WPILibConnectionStatus.tsx
@@ -24,9 +24,9 @@ const WPILibConnectionStatus: React.FC = () => {
             className="select-none absolute right-1 top-1 py-2 px-4 rounded-lg gap-2"
         >
             {status ? (
-                <span className="text-green-500 self-center">{SynthesisIcons.CHECK}</span>
+                <SynthesisIcons.CHECK className="text-green-500 self-center" />
             ) : (
-                <span className="text-cancel-button self-center">{SynthesisIcons.XMARK}</span>
+                <SynthesisIcons.XMARK className="text-cancel-button self-center" />
             )}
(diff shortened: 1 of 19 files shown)


rutmanz (commented):
I think it might make more sense for SynthesisIcons to be aliases of MUI icons (`HAND = FaHandPaper`) and then use them as components (`<SynthesisIcons.HAND />`) so that the properties can be passed through without `cloneElement` and it's clear that they're component

fission/src/ui/components/WPILibConnectionStatus.tsx
```diff
@@ -24,9 +24,13 @@ const WPILibConnectionStatus: React.FC = () => {
             className="select-none absolute right-1 top-1 py-2 px-4 rounded-lg gap-2"
         >
             {status ? (
-                <FaCheck className="text-green-500 self-center" />
+                <span className="text-green-500 self-center">
+                    <SynthesisIcons.CHECK />
```
0xda157: we have to wrap it here? we can't just
```suggestion
                <SynthesisIcons.CHECK className="text-green-500 self-center" />
```

AlexD717: You are right. That was leftover from the original commit. I will go fix that

rutmanz (approved):
LVCALAGITM

0xda157 (approved):
diff LGTM

PR #1388, opened 2026-07-08
Update Vite (partial `[SYNTH-145]`)

0xda157:
## Task

part of SYNTH-145

## Verification

---

- [x] User-facing polish:

### DIFF
--- a/fission/bun.lock
+++ b/fission/bun.lock
@@ -14,7 +14,8 @@
         "@react-three/drei": "^9.122.0",
         "@react-three/fiber": "^8.18.0",
         "@synthesis.adsk/jolt-physics": "^1.0.4",
-        "@vitest/browser": "^3.2.4",
+        "@vitest/browser": "^4.1.10",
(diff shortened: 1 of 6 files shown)


rutmanz (commented):
tests are failing, but also is there a reason to go one dependency at a time rather than all at once? A lot of these things have peer dependencies on different versions so to me it makes sense to try and upgrade everything in one PR

0xda157:
> is there a reason to go one dependency at a time rather than all at once? 

I think that updating our testing code should be separate from updating our actual UI, so that there is less opportunity to miss regressions. 

> A lot of these things have peer dependencies on different versions so to me it makes sense to try and upgrade everything in one PR

Sure it makes sense if the dependencies are peers of each other, then we should update them together, but that isn't the case for vite. it only depends on the vite plugins

AlexD717:
Is this still being worked/developed on or should it be closed or changed to a draft PR? Tests are failing

PR #1392, opened 2026-07-08
Auto-clean Autodesk Registry URLs `[SYNTH-221]`

rutmanz:
## Task

SYNTH-221

## Symptom

As [bun does not support the `omit-lockfile-registry-resolved` npm option](https://github.com/oven-sh/bun/issues/15448), and Autodesk requires the use of the internal artifactory registry, artifactory URLs will be committed to `bun.lock` when new packages are installed or the lockfile is otherwise regenerated. This breaks CI and is undesirable in a public repository.

## Solution

Bun runs the `postinstall` script after packages are installed, which removes artifactory urls using regex. Note that I can't use standard commands like `sed` or `> /dev/null` and instead use npm packages to support non-unix systems

## Verification

if you `bun install <whatever>` while using the Autodesk registry, the bun.lock file still has no artifactory urls. (Make sure to remove the package after)

---

- [X] User-facing polish:

### DIFF
--- a/fission/bun.lock
+++ b/fission/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "synthesis-fission",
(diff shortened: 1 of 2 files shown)


azaleacolburn (approved):
L.G.T.M - This is pretty L.I.V.E

Dhruv-0-Arora (approved):
TWLQSAITILRFM

AlexD717 (approved):
lgtm

PepperLola (approved):

PR #1379, opened 2026-07-03
Driver Station View `[SYNTH-64] & [SYNTH-103]`

AlexD717:
## Task

SYNTH-64

SYNTH-103

## Symptom

Currently, users have no way to easily switching to a fixed field view, such as a driver station view.

## Solution

Implemented the ability to set fixed camera points around the field (through the configure assets panel), save them to mirabuf allowing them to persist, and allow the user to easily select a field view point when focused on a field. I have updated the 2023 and 2026 field to include default positions (also fixed the 2026 field to have a red and blue side instead of both being red). Panning functionality was also fixed, allowing you to adjust your position manually.

## Verification

- [x] Spawning in 2023_v8 field and focusing on it, allows you to select a driver station view in the configure camera panel
- [x] Spawning in the 2026_v2 field and focusing on it, allows you to select a driver station view in the configure camera panel
- [x] Spawning in the 2026_v2 field (no gamepieces) and focusing on it, allows you to select a driver station view in the configure camera panel
- [x] Can configure your own camera points through the configure assets panel
- [x] Can right click and drag at any point to pan (I recommend using a mouse to test this, as touchpads can be a bit unreliable) 
- [x] Can customize pan sensitivity in settings
- [x] Normal robot tracking camera works

---

Before merging, ensure the following criteria are met:

- [x] All acceptance criteria outlined in the ticket are met.
- [x] Necessary test cases have been added and updated.
- [x] A feature toggle or safe disable path has been added (if applicable).
- [x] User-facing polish:
  - Ask: *"Is this ready-looking?"*
- [x] Cross-linking between Jira and GitHub:
  - PR links to the relevant Jira issue.
  - Jira ticket has a comment referencing this PR.
- [ ] PR #1371 should be merged first (and merge conflicts fixed)

PR #1393, opened 2026-07-09
Gamepiece Asset Support `[SYNTH-37]`

BrandonPacewic:
A lesson in scope creep. Big thanks to @azaleacolburn for doing most of the work for me.

## Task

SYNTH-37
Resolves #1190.

## Symptom

Gamepieces are inherently attached to fields. As such, they could not be moved or spawned in isolation.

## Solution

Separate the logic such that `PIECES` becomes its own asset type. Configurable and spawnable by themselves. 

## Verification

No regression on old fields, gamepieces from previous fields are able to be managed, moved, removed, added, and cached all separately from a field.

---

- [ ] User-facing polish:


RoushilS:
Should you be able to focus camera on a gamepiece? If so, the configure camera menu shows up as blank when focused and gamepieces are not included in the dropdown. (which they probably shouldn't be)

AlexD717:
> Should you be able to focus camera on a gamepiece? If so, the configure camera menu shows up as blank when focused and gamepieces are not included in the dropdown. (which they probably shouldn't be)

Note: PR #1379 significantly reworks the camera selection panel, so if you do end up modifying it, expect merge conflicts.

AlexD717 (changes requested):
For some fields gamepieces appear a bit off of their starting location.

Ex, this is on dev for the 2023 field

Image

And this is on this branch, with the gampieces being a bit off their starting locaiton (black cross)

Image

For the 2025 field, the green balls spawn in the right location, but I don't think the coral (grey pipes) are spawning in properly.
Its supposed to look like this, with the green balls on top of the grey pipes

Image

But there aren't any pipes at all

Image

When looking at the field in slow motion, the grey pipes look to be starting to spawn (but spawning halfway in the ground), but then disappear (or fall through the ground) when the algae collides with them.

Image


In addition, drag mode no longer works for gamepieces.

AlexD717 (changes requested):
Image

The 2025 field grey pipes also spawn a static half pipe (it can't be dragged with drag mode, and doesn't collide with robots).

jolt
Dhruv-0-Arora, line 1: I'm not sure whether a merge from dev would fix this but the current submodule is pointing to Azalea's fork and not synthesis.adsk's fork

fission/src/mirabuf/MirabufParser.ts
```diff
@@ -251,6 +444,13 @@ class MirabufParser {
         return node
     }
 
+    private deleteRigidNode(node: RigidNode) {
+        const index = this._rigidNodes.indexOf(node)
+        if (index != -1 && index != null) {
```
Dhruv-0-Arora:
```suggestion
        if (index != -1) {
```
.indexOf() can't return null

fission/src/mirabuf/MirabufParser.ts
```diff
@@ -251,6 +444,13 @@ class MirabufParser {
...
+            this._rigidNodes.splice(index)
```
Dhruv-0-Arora:
```suggestion
            this._rigidNodes.splice(index, 1)
```
I think specifying the number of items to delete would be good practice.
Right now, the use case of this function is when we do 
```ts
const gpRn = this.newRigidNode(GAMEPIECE_SUFFIX)
...
this.deleteRigidNode(gpRn)
```
So you are creating a new rigidNode and then deleting it (so it is always the last one).

But in a future use case where the rigid node being deleted is not the last rigid node, it would silently delete everything after that node too.

fission/src/ui/modals/mirabuf/ImportLocalMirabufModal.tsx
```diff
@@ -61,23 +62,64 @@ const ImportLocalMirabufModal: React.FC<ModalImplProps<void, ImportLocalMirabufP
...
+                                    continue
+                                }
+                                cachedTypeNames.add(typeName)
+
+                                zeroGamePieceInstancePosition(pieceAssembly)
+                                const cacheInfo = await MirabufCachingService.storeAssemblyInCache(pieceAssembly, {
+                                    miraType: MiraType.PIECE,
+                                    name: typeName,
+                                })
+                                if (!cacheInfo) continue
```
Dhruv-0-Arora: This looks to be the same as what is in ImportMirabufModal.tsx
Could you make this a helper?

Dhruv-0-Arora (changes requested):

AlexD717 (changes requested):
Found one more minor bug. Robots are now spawning significantly higher than they previously used to when a field is spawned (while dev tools still list the default spawn position as close to the ground).

Image

fission/src/mirabuf/MirabufParser.ts
AlexD717:
```suggestion
// TODO Account for special versions
```

Doesn't this PR complete the 2nd TODO?

AlexD717 (changes requested):
In the spawn asset panel, on the gamepiece tab, I don't see any cached fields (even though I do see them on the fields tab), and I also don't see the option to spawn any gamepiece.

Image

AlexD717 (changes requested):
It appears that one of your most recent changes broke scoring zone scoring, as dragging a game piece inside of one no longer increases the score. Checking out to a slightly later version has it working.

fission/src/mirabuf/MirabufParser.ts
```diff
@@ -168,38 +192,214 @@ class MirabufParser {
...
+                              e[2],
+                              e[6],
+                              e[10],
+                              e[14] * 100,
+                              e[3],
+                              e[7],
+                              e[11],
+                              e[15],
+                          ],
+                          // biome-ignore-end format: We would prefer to visualize this as a matrix
```
AlexD717, line 366: Maybe I just don't know enough about matrices, but this doesn't seem to be visualized properly

AlexD717 (approved):
lgtm

fission/src/mirabuf/MirabufParser.ts
BrandonPacewic: Good catch, my biome keeps wanting to reformat this matrix for some reason.

0xda157: It's because `biome-ignore-start` can't be used with `format`. This will be fixed with #1452

Dhruv-0-Arora (changes requested):
Gamepieces currently do not stop when you push them around but rather continue oscillating.
No friction?

AlexD717:
In addition, I think it makes sense to have the gampieces not show up in the top bar search (as we discussed yesterday)
Screenshot 2026-08-07 at 8 05 37 AM

PR #1399, opened 2026-07-09
Clean up react bad pratices [SYNTH-224]

0xda157:
## Task

SYNTH-224

## Symptom

many bad practices which make code less readable and in some cases less performant

## Solution

clean up the bad code

## Verification

`bun run fmt`, `bun run lint` and `bun run build` all pass

---

- [x] User-facing polish:


fission/src/ui/components/Label.tsx
```diff
@@ -16,12 +15,10 @@ const sizeToVariant: { [key in LabelSize]: Variant } = {
...
-        <Typography variant={sizeToVariant[size]} {...props}>
-            {children}
-        </Typography>
-    )
-}
+const Label = ({ children, size, ...props }: PropsWithChildren<LabelProps> & TypographyProps): JSX.Element => (
+    <Typography variant={sizeToVariant[size]} {...props}>
+        {children}
+    </Typography>
+)
```
0xda157: I think we should use `=> ( ... )` instead of `=> { return ... }` because `( ... )` more directly means "there is only one statement here." it also visually looks nice imo

fission/src/ui/components/ContextMenu.tsx
```diff
@@ -12,7 +11,7 @@ interface ContextMenuStateData {
     location: [number, number]
 }
 
-const ContextMenu: React.FC = () => {
+const ContextMenu = (): JSX.Element | null => {
```
rutmanz: My IDE is warning that the JSX global namespace is deprecated and should be avoided

fission/src/ui/components/AnalyticsConsent.tsx
```diff
@@ -7,69 +7,67 @@ interface AnalyticsConsentProps {
     onConsent: () => void
 }
 
-const AnalyticsConsent: React.FC<AnalyticsConsentProps> = ({ onConsent, onClose }) => {
-    return (
+const AnalyticsConsent = ({ onConsent, onClose }: AnalyticsConsentProps): JSX.Element => (
```
rutmanz: There are quite a few changes to switch the typing of functional components, what's the theory behind that?

fission/src/util/TypeConversions.ts
```diff
@@ -18,10 +18,22 @@ export function convertArrayToThreeMatrix4(arr: number[]) {
...
+        arr[9],
+        arr[13],
+        arr[2],
+        arr[6],
+        arr[10],
+        arr[14],
+        arr[3],
+        arr[7],
+        arr[11],
+        arr[15]
```
rutmanz: Yeah so this was better before

rutmanz (changes requested):

fission/src/ui/components/AnalyticsConsent.tsx
0xda157: I think that `React.FC<AnalyticsConsentProps>` is too broad and `AnalyticsConsentProps -> React.JSX.Element` is a much more straight forward and narrower type. If we don't use `React.FC` the return type becomes much more useful, for example `(): React.JSX.Element | null` tells someone using their lsp hover that the function may not render, without having to view implementation of the component. As far as I can tell, aside from `React.FC` previously being bad pratice, there's no strong recommendation either way in the react community.

fission/src/ui/components/AnalyticsConsent.tsx
```diff
@@ -7,69 +7,67 @@ interface AnalyticsConsentProps {
     onConsent: () => void
 }
 
-const AnalyticsConsent: React.FC<AnalyticsConsentProps> = ({ onConsent, onClose }) => {
-    return (
+const AnalyticsConsent = ({ onConsent, onClose }: AnalyticsConsentProps): React.JSX.Element => (
```
rutmanz: My IDE is also warning that React should be imported instead of used as a global namespace to make resolution more reliable

rutmanz: And this matters less but stylistically I prefer `React.ReactElement` over `React.JSX.Element` (which is basically an alias)

0xda157:
lowkey decided that `React.FC` doesn't matter that much the remove ternary blocks doesn't really matter and too much of a pain

fission/src/.npmrc
rutmanz, line 1: not again

rutmanz (changes requested):
Otherwise LGTM


For the future though, it's a lot more tedious to re-review a PR if you force push because you can't see what's changed since the last review anymore.

fission/src/ui/panels/simulation/WiringPanel.tsx
```diff
@@ -466,45 +466,41 @@ const WiringPanel: React.FC<PanelImplProps<void, void>> = ({ panel }) => {
         configureScreen(panel!, { title: "Wiring Panel" }, { onBeforeAccept: save })
     }, [save])
 
+    if (!selectedAssembly || !simConfig) return "ERR"
```
Dhruv-0-Arora, line 469: This is pretty inconsequential but you changed "ERRR" to "ERR"

fission/src/ui/components/SelectMenu.tsx
```diff
@@ -159,28 +159,22 @@ const SelectMenu: React.FC<SelectMenuProps> = ({
...
-                                    />
-                                )
-                            })
-                        ) : (
-                            <>
-                                {/** No options available text */}
-                                <Label size="sm">{noOptionsText ?? "No options available!"}</Label>
-                            </>
-                        )}
+                        {options.length <= 0 && <Label size="sm">{noOptionsText ?? "No options available!"}</Label>}
```
Dhruv-0-Arora, line 168:
```suggestion
                        {options.length === 0 && <Label size="sm">{noOptionsText ?? "No options available!"}</Label>}
```

fission/src/ui/panels/configuring/assembly-config/interfaces/inputs/ConfigureInputsInterface.tsx
```diff
@@ -76,67 +76,66 @@ const ConfigureInputsInterface: React.FC<PanelImplProps<void, ConfigurePanelCust
...
+            }}
+            deleteCondition={val => {
+                if (!(val instanceof SchemeSelectionOption)) return false
+
+                return val.scheme.customized
+            }}
+            onAddClicked={() => {
+                openModal(NewInputSchemeModal, undefined)
+            }}
+            defaultSelectedOption={selectedScheme ? schemeOptionMap.get(selectedScheme) : undefined}
```
Dhruv-0-Arora, line 140: this always evaluates to undefined since selectedScheme is guarenteed
This was like this before this PR though so you could consider changing it to 
```suggestion
			defaultSelectedOption={undefined}
```

Dhruv-0-Arora (approved):
Otherwise lgtm

fission/src/ui/components/SelectMenu.tsx
0xda157: Why? unsigned numbers don't exist in typescript, and I think `<=` is more expressive.

fission/src/ui/panels/configuring/assembly-config/interfaces/inputs/ConfigureInputsInterface.tsx
rutmanz: hang on what does this mean

fission/src/ui/panels/configuring/assembly-config/interfaces/inputs/ConfigureInputsInterface.tsx
Dhruv-0-Arora, line 89: @rutmanz this if statement here will make it so selectScheme is always true in the code down below. So the ternary always evaluates to undefined

PR #1403, opened 2026-07-10
Remove joint config options on Codesim

RoushilS:
## Task

SYNTH-234

## Symptom

When the user is doing code simulation and the brain is set to WPILib, the joint configurationand sequence options are visible in the configure assets panel. This can be confusing, and hinder the user as these options are not used during code simulation.

## Solution

Add a check to see if the brain is synthesis before adding the config selection option, in the configure panel.

## Verification

- [ ] Joint selection options are hidden on wpilib brain, but are still functional.

---

- [ ] User-facing polish:

### DIFF
--- a/fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
+++ b/fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
@@ -374,17 +374,20 @@ const ConfigurePanel: React.FC<PanelImplProps<void, ConfigurePanelCustomProps>> 
                         "Configure the robot’s ejector mechanism, which controls the release or expulsion of game pieces."
                     ),
 
-                    new ConfigModeSelectionOption(
-                        "Configure Joints",
-                        ConfigMode.SUBSYSTEMS,
-                        "Set the velocities, torques, and accelerations of your robot's motors."
-                    ),
-
-                    new ConfigModeSelectionOption(
-                        "Sequence Joints",
-                        ConfigMode.SEQUENTIAL,
-                        "Set which joints follow each other. For example, the second stage of an elevator could follow the first, moving in unison with it."
-                    ),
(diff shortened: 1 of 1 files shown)


fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
```diff
@@ -374,17 +374,20 @@ const ConfigurePanel: React.FC<PanelImplProps<void, ConfigurePanelCustomProps>>
...
-                        ConfigMode.SEQUENTIAL,
-                        "Set which joints follow each other. For example, the second stage of an elevator could follow the first, moving in unison with it."
-                    ),
+                    ...(selectedAssembly?.brain?.brainType === "synthesis"
+                        ? [
+                              new ConfigModeSelectionOption(
+                                  "Configure Joints",
+                                  ConfigMode.SUBSYSTEMS,
+                                  "Set the velocities, torques, and accelerations of your robot's motors."
+                              ),
```
PepperLola, line 383: I think we would still want to be able to configure joints in code sim mode so that they could match with the actual robot design, it's just sequencing that doesn't really make sense because the robot code would do that for you.

PepperLola (changes requested):

fission/src/ui/panels/configuring/assembly-config/ConfigurePanel.tsx
rutmanz: (Note that #1377 merges these into a single panel)

RoushilS: In that case, I will hold off on the this PR, for now.

rutmanz:
(converted to draft until #1377 is dealt with just so it doesn't show up in the to-review section, feel free to re-ready it when you want)

RoushilS:
Obsolete

PR #1404, opened 2026-07-10
Destroy Game Piece When Falling of Map `[SYNTH-216]`

AlexD717:
## Task

SYNTH-216

## Symptom

Currently game pieces can be thrown out of the field and they will fall forever (which also breaks the camera if you are focused on a field).

## Solution

Destroy game pieces when their y level is below a certain amount.

## Verification

- [x] Dropping a game piece of the map (ex, using drag mode or ejecting it) results in it despawning
- [x] Robots still spawn back at their starting point (if they are holding a game piece it will respawn with them)

---

- [x] User-facing polish:

### DIFF
--- a/fission/src/mirabuf/MirabufSceneObject.ts
+++ b/fission/src/mirabuf/MirabufSceneObject.ts
@@ -107,6 +107,7 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
     private _robotPreferences: RobotPreferences | undefined
 
     private _ejectables: EjectableSceneObject[] = []
+    private _gamePieces: Array<{ bodyId: Jolt.BodyID; rigidNode: RigidNodeReadOnly }> = []
     private _intakeSensor?: IntakeSensorSceneObject
     private _scoringZones: ScoringZoneSceneObject[] = []
     private _protectedZones: ProtectedZoneSceneObject[] = []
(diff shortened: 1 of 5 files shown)


fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -175,6 +176,37 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+    public destroyGamePiece(bodyId: Jolt.BodyID): void {
+        const key = bodyId.GetIndexAndSequenceNumber()
+        const entry = this._gamePieces.find(g => g.bodyId.GetIndexAndSequenceNumber() === key)
+        if (!entry) return
+
+        // Unlikely that the game piece is in a scoring zone, but if it is, remove it from the zone's contact list
+        const zones = World.sceneRenderer.filterSceneObjects(
+            (x): x is ScoringZoneSceneObject => x instanceof ScoringZoneSceneObject
+        )
+        zones.forEach(zone => ScoringZoneSceneObject.removeGamepiece(zone, bodyId))
```
PepperLola, line 197: `removeGamepiece` currently only actually removes a gamepiece from the scoring zone if `!shouldPointsAccumulate` meaning otherwise the gamepiece body ID would stay referenced even after it's been removed from the scene.

PepperLola (commented):
Wouldn't robots respawning with the gamepiece they're holding be dependent on both the update order of their respective position trackers and whether the root node of the robot was above or below the gamepiece?

Also, gamepiece body IDs never get removed from the field's mechanism's nodeToBody map, meaning anything that would need to reference that later would end up with a bad body ID. It's also used a lot in multiplayer code, which means you could probably end up with ghost gamepieces. I think we should go back over a lot the spawning/configuring/destroying stuff we have and make sure it would all work in multiplayer if we end up deciding to pursue that again.

AlexD717:
> Also, gamepiece body IDs never get removed from the field's mechanism's nodeToBody map, meaning anything that would need to reference that later would end up with a bad body ID. It's also used a lot in multiplayer code, which means you could probably end up with ghost gamepieces. I think we should go back over a lot the spawning/configuring/destroying stuff we have and make sure it would all work in multiplayer if we end up deciding to pursue that again.

There isn't really any easy and clean way to destroy game pieces right now, sync with multiplayer. Removing it from the nodeToBody map causes errors. Rather than doing a scuffed implementation, I am going to hold off until PR #1393 (Game piece asset support) gets merged, which will make this easy and clean to implement.

AlexD717:
Waiting on proper gamepiece implementation

PR #1406, opened 2026-07-11
fix(preferences): scoreboard disabled by default & pref snapshot

Dhruv-0-Arora:
With the UI-refactor, it won't make sense to have the scoreboard enabled by default because switching to the gameplay mode will enable it. Right now it just takes up space and makes testing on mobile difficult.

---

- [ ] User-facing polish:


Dhruv-0-Arora:
squash this pr

fission/src/test/PreferencesSystem.test.ts
```diff
@@ -30,7 +30,7 @@ describe("Preferences System Global Values", () => {
 
         expect(PreferencesSystem.getUserPreference("ZoomSensitivity")).toBe(15)
         expect(PreferencesSystem.getUserPreference("RenderSceneTags")).toBe(true)
-        expect(PreferencesSystem.getUserPreference("RenderScoreboard")).toBe(true)
+        expect(PreferencesSystem.getUserPreference("RenderScoreboard")).toBe(false)
```
rutmanz: lets be so for real we could (should) check that these are the default values, not that they are equal to these literals

rutmanz (changes requested):
otherwise lgtm

fission/src/test/PreferencesSystem.test.ts
Dhruv-0-Arora: Done.

rutmanz (approved):

AlexD717 (approved):
Should be merged after #1377

PR #1407, opened 2026-07-13
Clean Up Invalid Collider Warnings `[SYNTH-222]`

azaleacolburn:
## Task

SYNTH-222

## Symptom

When spawning basically any complex model, a bunch of warnings are printed about invalid colliders. Some are duplicates, and some are just annoying. They have made debugging slightly more annoying for me.

## Solution

I removed one warning, which triggered after one of the other two more specific warnings, so it felt redundant. I put two other warnings behind a constant variable called `DEBUG_COLLIDER_WARNINGS` in `PhysicsSystem.ts`. Additionally, I added a conditional warning to an analogous path that didn't have one before, predicated on the same debug variable.

## Verification

All tests pass.

---

- [x] User-facing polish:

### DIFF
--- a/fission/src/systems/physics/PhysicsSystem.ts
+++ b/fission/src/systems/physics/PhysicsSystem.ts
@@ -1151,9 +1151,12 @@ class PhysicsSystem extends WorldSystem {
         })
 
         if (points.size() < 4) {
+            if (DEBUG_COLLIDER_WARNINGS) console.warn("Could not create convex shape for part")
+
             JOLT.destroy(settings)
             JOLT.destroy(min)
             JOLT.destroy(max)
+
             return
         }
 
--- a/fission/src/systems/physics/PhysicsSystem.ts
+++ b/fission/src/systems/physics/PhysicsSystem.ts
@@ -43,6 +43,8 @@ import {
     setAxes,
 } from "./ConstraintSettingsUtilities"
 
+const DEBUG_COLLIDER_WARNINGS = false
+
 /**
  * Layers used for determining enabled/disabled collisions.
  */
@@ -942,7 +944,6 @@ class PhysicsSystem extends WorldSystem {
                 // const partShapeResult = this.CreateConvexShapeSettingsFromPart(partDefinition)
 
                 if (!partShapeResult) {
-                    console.warn("Skipping collider (no valid shape settings)", debugLabel)
                     return [undefined, undefined]
                 }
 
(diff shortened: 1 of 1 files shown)


AlexD717 (approved):
lgtm

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -36,6 +36,8 @@ import {
     setAxes,
 } from "./ConstraintSettingsUtilities"
 
+const DEBUG_COLLIDER_WARNINGS = false
```
0xda157, line 46: surely there's a better way to manage this. check if we're in dev mode? add an option for debug logging?

azaleacolburn: But I don't want it in dev mode; it should only print when the developer explicitly wants it to print. This is a pattern we use elsewhere, so changing it should be a separate ticket.

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1150,9 +1151,12 @@ class PhysicsSystem extends WorldSystem {
         })
 
         if (points.size() < 4) {
+            if (DEBUG_COLLIDER_WARNINGS) console.warn("Could not create convex shape for part")
```
Dhruv-0-Arora, line 1154:
```suggestion
            if (DEBUG_COLLIDER_WARNINGS) {
           		console.warn("Could not create convex shape for part")
           	}
```

Dhruv-0-Arora (approved):
for consistency's sake

fission/src/systems/physics/PhysicsSystem.ts
0xda157: if this is important enough to enforce, it should be added to the lint/format config, and if it's not that important we shouldn't be nit picking it

fission/src/systems/physics/PhysicsSystem.ts
rutmanz: fwiw I think this is a totally valid choice. Like they could be in a file of their own but it doesn't matter imo

fission/src/systems/physics/PhysicsSystem.ts
rutmanz: my take is it depends on how long the line is and is entirely vibe-based. I personally like it Dhruv's way more but it doesn't really matter and would not be appropriate for a formatter rule

rutmanz (approved):
TM,TLLSCACAIWA

0xda157 (approved):

fission/src/systems/physics/PhysicsSystem.ts
azaleacolburn: My general rule is that if the curly brackets can be omitted, they should be, and it's the job of the formatter to put the body on the next line (indented, of course)

PR #1408, opened 2026-07-13
Remove `.npmrc`

BrandonPacewic:
`.npmrc` got added and it broke our GitHub actions. Since we don't need this right now we can just remove it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
-.vs/
-.vscode/
+.*
 build/
 dist/
 out/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-.*
+.vs/
+.vscode/
 build/
 dist/
 out/
@@ -9,3 +10,4 @@ renderer/
 *.exe
 .idea
 **/.direnv
+.npmrc
--- a/fission/.gitignore
+++ b/fission/.gitignore
@@ -1,4 +1,3 @@
-.*
 public/Downloadables
 package-lock.json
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,5 +1,4 @@
-.vs/
-.vscode/
+.*
```
0xda157: why on earth would we ignore every file prefixed with `.`?? this is bound to cause tons of confusion.  
```suggestion
.vs/
.vscode/
.npmrc
```

BrandonPacewic: This is fairly common and not inherently erroneous. Linux uses this exact pattern.

rutmanz (changes requested):
Given the number of `.*` files we do include (`.gitmodules`, `.envrc`, `.git-blame-ignore-revs`, `.gitattributes`), it does seem kind of odd to chose to ignore all of them by default. At the very least all dotfiles that are checked in should be explicitly excluded from the gitignore imo. My brief perusing of the forums indicates that `git add -f` is somewhat brittle if files get deleted and added back, moved, etc.

I also think that it's nice to be able to see an accurate list of modified files in the git change list (GUI or `git status`), and then chose whether to ignore a file you've added there or not, rather than have it ignored by default and not show up anywhere as a new file that you might want to pay attention to.

I agree that it's not inherently erroneous, but I do think Daisy is right that it is very confusing and probably not the best way of going about this.

rutmanz (approved):
g,tltm

PR #1411, opened 2026-07-13
feat(ui): library modal `SYNTH-146`

Dhruv-0-Arora:
Replaces the old docked "Spawn Asset" side panel with a big centered Library modal

SYNTH-146

## Solution

Assets didn't have a year before, so I added year and thumbnail fields to the asset manifest. The build script (update_manifest.ts) figures out the year by reading it from each file's name (e.g. KitBot (2024) → 2024), and assets with no year fall into an "Other" tab. This metadata flows all the way through to cached assets, so downloads keep their info.

The design has year tabs across the top, and each tab shows that year's robots and its field together in a card grid. Each card has a thumbnail slot (with a placeholder icon until real images get uploaded), a spawn button, a "downloaded" badge, and a delete button. The old Saved / Autodesk Hub / Import-from-File options are still there.

Screenshot 2026-07-13 at 4 22 27 PM

Switched the old panel over to a modal everywhere it gets opened, moved the shared spawn logic into its own file so it isn't duplicated, and fixed the onboarding tour to react to the modal opening instead of the panel.

> [!IMPORTANT]
> Years and thumbnails only show up once the asset pack is rebuilt and redeployed - the committed asset pack wasn't touched yet

---

- [ ] User-facing polish:
Thread structured per-asset `year` and optional `thumbnail` from the
manifest through to cached assets so the asset library can group and
preview assets:

- Widen the manifest schema (ManifestFileEntry) with year/thumbnail.
- Derive year (parsed from filenames at build time) and resolve a
  sibling thumbnail path in update_manifest.ts.
- Expose both on DefaultAssetInfo, resolving the thumbnail to a
  servable URL.
- Persist them onto MirabufCacheInfo via cacheRemote/storeInCache so
  downloaded assets retain the metadata; older cached entries still
  deserialize since the fields are optional.


rutmanz:
as I recall, there is a thumbnail embedded in the mira metadata that is currently unused. Perhaps this would be a good use for it?

PepperLola (commented):
~~I haven't tested it locally yet but, going off of the screenshot: would it make sense to completely remove the "Saved Assets" section and instead make the visual distinction between cached and non-cached assets be that cached ones have a + icon instead of a download icon? I feel like users will generally want to look for cached assets first, so it doesn't really make sense to put them underneath the ones for that year.~~

On second thought it looks like this may already be the case and that area is specifically for assets that don't fit in any year?

PR #1415, opened 2026-07-15
Spawn Location Configuration `[SYNTH-235]`

AlexD717:
## Task

SYNTH-235

## Symptom

Currently we support having custom robot spawn locations (used when starting a match), but no user facing way to edit it (apart from developer tools, which isn't user friendly).

## Solution

Add a configuration options for fields for editing robot spawn locations.

## Verification

- [x] Can edit robot spawn locations for fields (test by starting match mode with "Move Robots to Starting Position" toggled on)
- [x] Camera position editing still works
- [x] Shows a cone indicator for robot positions and fixed rotation camera positions corresponding to their direction.

---

- [x] User-facing polish:


rutmanz:
image
image

I might add a background of some kind and maybe alliance color (attached what I did for protected zones for reference) just to make each spawn position card feel separate. Right now the style feels a little unpolished.

This is definitely an improvement over the old system, I was observing that there is no way to visualize the yaw though. Maybe an arrow like we have for the ejector but a different color? Or maybe like a ghost dozer to give a sense of direction and scale

It also might be nice to have the xyz coordinates editable, just to help people align them if they want. Could also leave that to the devtools interface though, I don't feel strongly about this one

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -354,8 +354,6 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
                 ? fieldLocations[this.alliance][this.station]
                 : fieldLocations?.default
 
-        // TODO
-        // Why are we calling this?
         field?.getPositionTransform(referencePos)
```
rutmanz: wait do you know why we are calling this?

fission/src/ui/panels/configuring/assembly-config/interfaces/FieldPointEditing.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/** Holds a physics pause for as long as the calling component is mounted (e.g. while editing a field-relative point). */
+export function useHoldPhysicsPauseWhileMounted() {
+    useEffect(() => {
+        World.physicsSystem.holdPause(PAUSE_REF_ASSEMBLY_CONFIG)
+        return () => {
+            World.physicsSystem.releasePause(PAUSE_REF_ASSEMBLY_CONFIG)
+        }
+    }, [])
+}
```
rutmanz: ok I really like this refactor. Does it make sense to use random IDs instead of the constants so that if you are for some reason opening multiple panels like this they don't cause any conflicts? Also I might put this (and the function below it) in a more global place, because these same patterns are used in a bunch of other panels and it would be nice to have them use this too

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureSpawnPositionsInterface.tsx
```diff
@@ -0,0 +1,226 @@
...
+}
+
+function setSpawnLocation(locations: SpawnLocations, path: SpawnSlotPath, value: SpawnLocation): SpawnLocations {
+    if (path === "default") return { ...locations, default: value }
+    return { ...locations, [path.alliance]: { ...locations[path.alliance], [path.station]: value } }
+}
+
+function capitalize(word: string): string {
+    return word[0].toUpperCase() + word.slice(1)
+}
```
rutmanz: probably move to a more general utility file

rutmanz (approved):
LGTM, I think some of the refactors could be made more general but up to you. I didn't realize that react hooks could be made reusable like that but it makes a lot of sense and feels much cleaner

fission/src/mirabuf/MirabufSceneObject.ts
AlexD717: the .getCenter call in the function mutates the referencePos vector. This is then needed by setObjectPosition to account for possible field transformations. I will add a comment.

fission/src/ui/panels/configuring/assembly-config/interfaces/FieldPointEditing.ts
AlexD717: Currently, you can only have one configure panel open and only one section open, so its impossible for multiple PAUSE_REF_ASSEMBLY_CONFIG to conflict. We have different pause refs for other panels (`fission/src/systems/physics/PhysicsTypes.ts` if you want to look at all of them). I'm assuming we don't use random ids for easier debugging.

PepperLola (dismissed):
I would like all of the spawn positions to be rendered when in the list view for the configuration, then only render a single one when you're editing one.

fission/src/ui/panels/configuring/assembly-config/interfaces/FieldPointEditing.ts
Dhruv-0-Arora, line 1: I think a test for the content here would be beneficial

fission/src/ui/panels/configuring/assembly-config/interfaces/ConfigureSpawnPositionsInterface.tsx
```diff
@@ -0,0 +1,234 @@
...
+            prefs.spawnLocations = initial
+        })
+    }, [registerCleanupFunction, selectedAssembly])
+
+    const updateLocation = useCallback(
+        (slot: SpawnSlot, updated: SpawnLocation) => {
+            setLocations(prev => {
+                const next = setSpawnLocation(prev, slot.path, updated)
+                next.hasConfiguredLocations = true
+                persist(next, selectedAssembly)
```
Dhruv-0-Arora: persist() being in setLocations is a react anti-pattern

fission/src/ui/panels/configuring/assembly-config/interfaces/FieldPointEditing.ts
```diff
@@ -0,0 +1,135 @@
...
+export function useFieldPointMarkers(
+    selectedField: MirabufSceneObject,
+    points: readonly FieldPointMarker[],
+    facing: "+z" | "-z" = "+z"
+) {
+    useEffect(() => {
+        const fieldRef = selectedField.getXZPositionTransform(new THREE.Vector3())
+
+        const markers = points.map(({ pos, yaw, pitch }) => {
+            const material = createIndicatorMaterial()
```
Dhruv-0-Arora: Wouldn't it be ideal to .dispose() this material?

fission/src/ui/panels/configuring/assembly-config/interfaces/FieldPointEditing.ts
```diff
@@ -0,0 +1,135 @@
...
+const DIRECTION_INDICATOR_COLOR = 0xffcc33
+const INDICATOR_RENDER_ORDER = 1000
+
+function createIndicatorMaterial(): THREE.MeshToonMaterial {
+    const material = World.sceneRenderer.createToonMaterial(DIRECTION_INDICATOR_COLOR)
+    material.depthTest = false
+    material.depthWrite = false
+    material.transparent = true
+    return material
+}
```
Dhruv-0-Arora: This seems to be more of a constant. Would it be better to create a constant material so you don't have to create it every time and just reference the same material instead?

Dhruv-0-Arora (changes requested):

fission/src/ui/panels/configuring/assembly-config/interfaces/FieldPointEditing.ts
AlexD717: It is being disposed? Line 131

```
        return () => {
            markers.forEach(({ group, material, geometries }) => {
                World.sceneRenderer.scene.remove(group)
                geometries.forEach(geometry => geometry.dispose())
                material.gradientMap?.dispose()
                material.dispose()
            })
        }
```

fission/src/ui/panels/configuring/assembly-config/interfaces/FieldPointEditing.ts
AlexD717: Can't be easily done as a pure constant. Changed so that its now created once and then referenced across all 7 spawn points.

AlexD717: Never mind, had a skill issue moment

Dhruv-0-Arora (approved):
I like 🐢 s

PR #1416, opened 2026-07-16
Remove Failing Multiplayer Test `[SYNTH-249]`

azaleacolburn:
## Task

SYNTH-249

## Symptom

Right now, one of the multiplayer tests tries to test two clients actually connecting. This will time-out on Autodesk WiFi because of the firewall, so it isn't really a very helpful test.

## Solution

Remove the test. I'm open to other suggestions though!

## Verification

Everything still builds, tests still pass

---

- [x] User-facing polish:


AlexD717 (approved):
If we still want to test this, you could add a flag to only run it during Github CI (since that doesn't have a firewall), but since we already know that multiplayer doesn't work and are deleting it, I don't think this is particularly important.

Dhruv-0-Arora (approved):
Thank you

rutmanz (approved):
I like @AlexD717's suggestion but doesn't really matter because we'll be updating the system soon anyway so this won't really be applicable

PepperLola (approved):

PR #1418, opened 2026-07-16
fix(physics): jolt static vectors being freed

Dhruv-0-Arora:
## Task

Preventing memory corruption bug.

### Symptom

`Jolt.Vec3().Normalize()` stores the result in a static temp memory allocation for all vectors. This means that if you do
```ts
a = Jolt.Vec3().Normalize()
b = Jolt.Vec3().Normalize() // overriding `a`
```
then the value of a is overridden by the value of b.

this also is the case with the functions

`AddRVec3()` & `Add()`

## Solution

Copying the values into a new Jolt.Vec3() through the new `ownVec3()` function. 

This, however, requires that the return value of functions such as `getPerpendicular()` must be destroyed

```ts
const normalAxis = getPerpendicular(constraintAxis)
settings.mNormalAxis1 = settings.mNormalAxis2 = normalAxis
JOLT.destroy(normalAxis)
```

instead of 

```ts
settings.mNormalAxis1 = settings.mNormalAxis2 = getPerpendicular(constraintAxis)
```
-> this version leaks the vector from getPerpendicular

*More information below and test file to show concept above in more depth*:

<details>

Testing file:

[JoltOwnership.test.ts](https://github.com/user-attachments/files/30247002/JoltOwnership.test.ts)

## Task

Jolt.Vec3 is stored in heap but 
anything a bound method returns by value (`Normalized()`, `AddRVec3()`, `Add()`) is not allocated. the binding keeps one static per method.

How it looks:
```cpp
Vec3* emscripten_bind_Vec3_Normalized_0(Vec3* self) {
    static Vec3 temp;             // one slot, reused forever
    temp = self->Normalized();    // 16 bytes overwritten, no malloc
    return &temp;
}
```

So, we instead will copy the values from this memory and store it in a heap allocated Jolt.Vec3 for safer memory management. This means, however, that these Jolt.Vec3 s will need to be Jolt.destroy(x) ed.

## Symptom

`createAnchorPoint` returned `jointOrigin.AddRVec3(...)` straight through, so callers got a static temp. `PhysicsSystem` freed it in four places: two `JOLT.destroy(anchorPoint)` on the ball constraint path, and two `convertJoltRVec3ToJoltVec3(anchorPoint)` whose `destroy` param defaults to true.

the function `JOLT.getPointer()` shows that `AddRVec3` returns 54992 every call, malloc's block *(heap)* starts around 1157600. all four were freeing a data segment address (meaning they did nothing).

### In addition

settings members are declared by value (`RVec3 mPoint1;`), so setters memcpy and the struct never takes ownership. `JOLT.destroy(settings)` won't free a vector you assigned in, but you can destroy it yourself right after assigning. we were dropping `wheelPos`, the four `urdfWheelBasis` vectors, and `minBounds`/`maxBounds`.

## Solution

1. new `ownVec3`/`ownRVec3` helpers so no static escapes the expression that made it. `createAnchorPoint` and `getPerpendicular` now return owned heap vectors
2. destroy the leaks above, plus `anchorPoint` now that it's real memory
3. `JoltOwnership.test.ts` pins down both halves: fresh address per malloc, recycling on destroy, the static below the heap, 2000 calls yielding one address, the clobber, and structs copying bytes in

## testing

`JoltOwnership.test.ts` passes in chromium + firefox

</details>


fission/src/systems/physics/ConstraintSettingsUtilities.ts
```diff
@@ -15,7 +35,7 @@ export function createAnchorPoint(jointInstance: mirabuf.joint.JointInstance, jo
         ? convertMirabufVector3ToJoltRVec3(jointInstance.offset)
         : new JOLT.RVec3(0, 0, 0)
 
-    const anchorPoint = jointOrigin.AddRVec3(jointOriginOffset)
+    const anchorPoint = ownRVec3(jointOrigin.AddRVec3(jointOriginOffset))
```
azaleacolburn: Why is `anchorPoint` not already owned here?

fission/src/systems/physics/ConstraintSettingsUtilities.ts
```diff
@@ -79,15 +100,20 @@ export function setAxes(
     versionNum?: number
 ) {
     const axis = getAxis(freedom, versionNum)
+    const constraintAxis = ownVec3(axis.Normalized())
 
-    let constraintAxis: Jolt.Vec3
     if ("mHingeAxis1" in settings) {
-        constraintAxis = settings.mHingeAxis1 = settings.mHingeAxis2 = axis.Normalized()
```
azaleacolburn, line 85: I still don't understand how normalized vectors are treated as pre-allocated static variables. Like there are infinitely many of them, it's not like a unit vector or anything. I'm hoping for a more concrete explanation of what's going on here.

fission/src/test/physics/JoltOwnership.test.ts
```diff
@@ -0,0 +1,180 @@
+import { describe, expect, test } from "vitest"
+import { getPerpendicular } from "@/systems/physics/ConstraintSettingsUtilities"
+import JOLT from "@/util/loading/JoltSyncLoader"
+
+/**
+ * what you have to destroy and why
```
azaleacolburn: holy claude

fission/src/test/physics/JoltOwnership.test.ts
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * jolt classes are c++ structs behind emscripten bindings. assigning into one
```
azaleacolburn: They're c++ classes behind emscripten bindings, are they not?

fission/src/test/physics/JoltOwnership.test.ts
```diff
@@ -0,0 +1,180 @@
...
+ *   - `new JOLT.Vec3(...)` (and the `convert*` helpers) is a real malloc, it's yours, destroy it
+ *     or it leaks
+ *   - anything a bound method hands back (`Normalized()`, `AddRVec3()`) is NOT allocated. the
+ *     binding just has one static per method:
+ *
+ *         Vec3* emscripten_bind_Vec3_Normalized_0(Vec3* self) {
+ *             static Vec3 temp;             // one slot, reused forever
+ *             temp = self->Normalized();    // 16 bytes overwritten, no malloc
+ *             return &temp;
+ *         }
```
azaleacolburn: So why is this static, doesn't every `Vec3` have its own normalized vector?

azaleacolburn: It makes sense that its not heap allocated though, i suppose

fission/src/test/physics/JoltOwnership.test.ts
```diff
@@ -0,0 +1,180 @@
...
+ *
+ *     nothing was allocated so nothing leaks and nothing can be freed. destroying one hands free()
+ *     a data segment address malloc never gave out
+ *
+ * rule: destroy what you allocated, not what you were handed
```
azaleacolburn: Okay wait but this doesn't work for things like getting position vectors, right? Like we're given our own copy of those, I thought.

fission/src/systems/physics/ConstraintSettingsUtilities.ts
azaleacolburn: Okay so basically, they just get overridden every time `Normalized()` is called on any vector again 💀

fission/src/test/physics/JoltOwnership.test.ts
azaleacolburn: They get overridden, this is so cursed

azaleacolburn:
Also could you not write your pr descriptions with claude please, it's just like a really annoying style

azaleacolburn:
Also, please use the pr description template in the future

fission/src/systems/physics/ConstraintSettingsUtilities.ts
Dhruv-0-Arora: > anything a bound method returns by value (Normalized(), AddRVec3(), Add()) is not allocated

anything passed through these functions becomes a static temp, so it is not owned. I did a fair but of testing in the ci to console log the memory locations of the variables, and these were the ones that would store it in a static temp.

fission/src/test/physics/JoltOwnership.test.ts
Dhruv-0-Arora: ... that was me?

fission/src/test/physics/JoltOwnership.test.ts
Dhruv-0-Arora: Yes sorry, they are c++ classes - thats a typo
The message is still the same though.

fission/src/test/physics/JoltOwnership.test.ts
Dhruv-0-Arora: I'm not quite sure by what you mean when you say "doesn't every `Vec3` have its own normalized vector"
I might be misunderstanding what you mean but hopefully this outline helps:
```ts
const original = new JOLT.Vec3(1,2,3).     // here we are creating a new Vec3 in the heap. This is not a unit vector and this vec3 needs to be destroyed at some point
const newVec = original.Normalized().           // this is now a unit vector but also is now a static temp.
                                                                     // It is stored outside the heap and stack.
                                                                     // The reason this happens is because of how emscripten binds the
                                                                     // Normalized fn

// pretend original2 exists
const anotherVec = original2.Normalized() // we are now rewriting the same memory address from newVec and effectively destroying the value there.

JOLT.destroy(original) 
```

Reason this is happening is because c++ is returning by value and Typescript can't hold that reference so emscripten is placing the value in a static temp shared across every Vec3 value and is giving a reference to that memory allocation to typescript.

fission/src/test/physics/JoltOwnership.test.ts
Dhruv-0-Arora: For position vectors, getPosition() and getLinearVelocity() also return by value so the binding routes them to the shared static temp.

Heres a script I wrote:

```js
import initJolt from "@synthesis.adsk/jolt-physics/wasm-compat"
const JOLT = await initJolt()
const ptr = o => JOLT.getPointer(o)

// copying over some stuff from the PhysicsSystem
const s = new JOLT.JoltSettings()
const of = new JOLT.ObjectLayerPairFilterTable(1); of.EnableCollision(0, 0)
const bp = new JOLT.BroadPhaseLayerInterfaceTable(1, 1); bp.MapObjectToBroadPhaseLayer(0, new JOLT.BroadPhaseLayer(0))
s.mObjectLayerPairFilter = of; s.mBroadPhaseLayerInterface = bp
s.mObjectVsBroadPhaseLayerFilter = new JOLT.ObjectVsBroadPhaseLayerFilterTable(bp, 1, of, 1)
const bi = new JOLT.JoltInterface(s).GetPhysicsSystem().GetBodyInterface()
const shape = new JOLT.BoxShapeSettings(new JOLT.Vec3(1, 1, 1)).Create().Get()
const body = bi.CreateBody(new JOLT.BodyCreationSettings(shape, new JOLT.RVec3(5, 6, 7), new JOLT.Quat(0, 0, 0, 1), JOLT.EMotionType_Static, 0))
const body2 = bi.CreateBody(new JOLT.BodyCreationSettings(shape, new JOLT.RVec3(5, 6, 7), new JOLT.Quat(0, 0, 0, 1), JOLT.EMotionType_Static, 0))

// GetPosition() hands back the SAME static slot every call
console.log(ptr(body.GetPosition()))
console.log(ptr(body2.GetPosition()), " same address")

// Normalized has a different static slot but the same thing happens
console.log(ptr(new JOLT.Vec3(3, 0, 0).Normalized()))
console.log(ptr(new JOLT.Vec3(0, 3, 0).Normalized()))

```

68896
68896  same address
54416
54416

fission/src/test/physics/JoltOwnership.test.ts
azaleacolburn: No I understood as of my last comment.

fission/src/test/physics/JoltOwnership.test.ts
azaleacolburn: okay

fission/src/systems/physics/ConstraintSettingsUtilities.ts
azaleacolburn: By static temp, you just mean a that there's some data written to a static piece of memory that will get overwritten the next time that static slot is written to?

azaleacolburn: What you're saying is that the result is always written to a piece of static memory, and a pointer to that memory is returned into js memory. Thus it should be cloned onto the wasm heap.

Dhruv-0-Arora: Yup 👍

fission/src/test/physics/JoltOwnership.test.ts
Dhruv-0-Arora: Sorry if it was unclear :( @azaleacolburn

fission/src/test/physics/JoltOwnership.test.ts
Dhruv-0-Arora: Weird, github did not show that last comment when I was replying

BrandonPacewic (commented):
I have mixed feelings about this pr. While it is true that we were not aware of this behavior there is no bug associated with these changes. (Specific to the static alias function return behaviour of course.)

I believe that #1434 will fix all the other memory changes addressed in this pr so those are not needed here (#1434 addresses a much larger scope of issues which is why I'm saying that it should take priority of fixing the other issues listed in this pr).

Adding a small comment to the top of this particular function (`createAnchorPoint`) seems warranted, generally do not want the other changes presented here imo.

With that being said #1434 does add this small comment addressing this side effect so I'm unsure of what to do with this pull request in that case:

https://github.com/Autodesk/synthesis/blob/4968b0490c8cfc4803b521ad8a261921e0b4e8b7/fission/src/systems/physics/ConstraintSettingsUtilities.ts#L9-L11

fission/src/systems/physics/ConstraintSettingsUtilities.ts
```diff
@@ -7,17 +7,16 @@ import { convertMirabufVector3ToJoltRVec3, convertMirabufVector3ToJoltVec3 } fro
...
         : new JOLT.RVec3(0, 0, 0)
     // TODO: Offset transformation for robot builder.
     const jointOriginOffset = jointInstance.offset
-        ? convertMirabufVector3ToJoltRVec3(jointInstance.offset)
-        : new JOLT.RVec3(0, 0, 0)
+        ? convertMirabufVector3ToJoltVec3(jointInstance.offset)
+        : new JOLT.Vec3(0, 0, 0)
 
-    const anchorPoint = jointOrigin.AddRVec3(jointOriginOffset)
+    anchorPoint.Add(jointOriginOffset)
```
Dhruv-0-Arora, line 18: Note: this change from returning a Jolt.RVec3 to a Jolt.Vec3 is not in #1434

rutmanz:
Brandon are you saying this should be closed as superseded by #1434?

PR #1421, opened 2026-07-17
feat(physics): allow gamepieces to sleep & adding simulation debt into timestep

Dhruv-0-Arora:
> [!IMPORTANT]
> The way that gamepieces go to sleep when the field first spawns is something we should look into:
> Currently, they are set to sleep (in this PR) however the MirabufSceneObject instantly wakes them up to move them to their desired location. This will tank performance for the first ~5 seconds while the gamepieces wait to go back to sleep.

SYNTH-257

---

- [ ] User-facing polish:

### DIFF
--- a/fission/src/systems/physics/PhysicsSystem.ts
+++ b/fission/src/systems/physics/PhysicsSystem.ts
@@ -69,6 +69,7 @@ const MAX_COLLISIONS_STEPS = 20
 // step count will scale depending on how many active bodies
 // there currently are
 const ACTIVE_BODY_STEP_BUDGET = 600
+export const MAX_SUBSTEP_PERIOD = 1.0 / 300.0
 // maximum amount of simulation time that can be consumed per render
 // for catching up to slow frame
 const MAX_SIMULATION_TIME_PER_FRAME = 4 * STANDARD_SIMULATION_PERIOD
@@ -1442,7 +1443,11 @@ class PhysicsSystem extends WorldSystem {
             const desiredSteps = Math.ceil(simTime / MAX_COLLISIONS_STEP_PERIOD)
             const activeBodies = this._joltPhysSystem.GetNumActiveBodies(JOLT.EBodyType_RigidBody)
             const affordableSteps = Math.floor(ACTIVE_BODY_STEP_BUDGET / Math.max(activeBodies, 1))
-            const collisionSteps = Math.max(1, Math.min(MAX_COLLISIONS_STEPS, desiredSteps, affordableSteps))
+            const minStableSteps = Math.ceil(simTime / MAX_SUBSTEP_PERIOD)
+            const collisionSteps = Math.max(
+                1,
+                Math.min(MAX_COLLISIONS_STEPS, desiredSteps, Math.max(affordableSteps, minStableSteps))
+            )
             this._joltInterface.Step(simTime, collisionSteps)
         }
 
--- a/fission/src/systems/physics/PhysicsSystem.ts
+++ b/fission/src/systems/physics/PhysicsSystem.ts
@@ -61,6 +61,14 @@ const LAYER_GHOST = 10
 const COUNT_OBJECT_LAYERS = 11
 
 export const STANDARD_SIMULATION_PERIOD = 1.0 / 60.0
+const MAX_COLLISIONS_STEP_PERIOD = 1.0 / 1200.0
+// cap on collision steps per rendered frame
+// when a frame is so slow that the cap is hit, each collision
+// step covers more simulation time
+const MAX_COLLISIONS_STEPS = 20
+// step count will scale depending on how many active bodies
+// there currently are
+const ACTIVE_BODY_STEP_BUDGET = 600
 // maximum amount of simulation time that can be consumed per render
 // for catching up to slow frame
 const MAX_SIMULATION_TIME_PER_FRAME = 4 * STANDARD_SIMULATION_PERIOD
@@ -96,7 +104,9 @@ function computeSphericity(volume: number, area: number): number {
     return volumeEquivalentSphereArea / area
 }
 
-// fixing simulation system in STANDARD_SIMULATION_PERIOD increments
+// simulation period. actual step length tracks the rendered
+// frame time (PhysicsSystem.update), so stable constant keeps
+// driver stiffness independent of construction-time frame rate.
 export function getLastDeltaT(): number {
     return STANDARD_SIMULATION_PERIOD
 }
@@ -152,8 +162,6 @@ class PhysicsSystem extends WorldSystem {
 
     private _pauseSet = new Set<string>()
 
-    private _simulationTimeDebt = 0
-
     private _bodyAssociations: Map<JoltBodyIndexAndSequence, BodyAssociate>
 
     public get isPaused(): boolean {
@@ -190,6 +198,7 @@ class PhysicsSystem extends WorldSystem {
         this._joltPhysSystem.GetPhysicsSettings().mDeterministicSimulation = false
         this._joltPhysSystem.GetPhysicsSettings().mSpeculativeContactDistance = 0.06
         this._joltPhysSystem.GetPhysicsSettings().mPenetrationSlop = 0.005
+        this._joltPhysSystem.GetPhysicsSettings().mTimeBeforeSleep = 0.2
 
         const ground = this.createBox(
             new THREE.Vector3(7.5, 0.1, 7.5),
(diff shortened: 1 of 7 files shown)


fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1077,9 +1085,20 @@ class PhysicsSystem extends WorldSystem {
...
+
                 const body = this._joltBodyInterface.CreateBody(bodySettings)
                 this._joltBodyInterface.AddBody(body.GetID(), JOLT.EActivation_Activate)
-                body.SetAllowSleeping(false)
+
+                // allowing gamepieces to sleep
+                this._joltBodyInterface.AddBody(
+                    body.GetID(),
+                    rn.isGamePiece ? JOLT.EActivation_DontActivate : JOLT.EActivation_Activate
+                )
```
Dhruv-0-Arora, line 1102: This makes gamepieces go to sleep when they are first spawned

AlexD717 (changes requested):
Drag Mode is broken

https://github.com/user-attachments/assets/6a752b42-de83-41bb-a5e5-825be7984643

It appears that you can't use drag mode for objects that are asleep, but the force is still applied once they wake up.

In addition, jittering is a bit worse on my MacBooks 120hz monitor, but still happens on my home desktop (weaker computer and 60hz monitor).

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -99,9 +104,11 @@ function computeSphericity(volume: number, area: number): number {
...
 }
 
-let lastDeltaT = STANDARD_SIMULATION_PERIOD
+// simulation period. actual step length tracks the rendered
+// frame time (PhysicsSystem.update), so stable constant keeps
+// driver stiffness independent of construction-time frame rate.
 export function getLastDeltaT(): number {
-    return lastDeltaT
+    return STANDARD_SIMULATION_PERIOD
 }
```
azaleacolburn: We were doing this before? I thought `lastDeltaT` was supposed to be calculated

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -99,9 +104,11 @@ function computeSphericity(volume: number, area: number): number {
     return volumeEquivalentSphereArea / area
 }
 
-let lastDeltaT = STANDARD_SIMULATION_PERIOD
+// simulation period. actual step length tracks the rendered
+// frame time (PhysicsSystem.update), so stable constant keeps
+// driver stiffness independent of construction-time frame rate.
 export function getLastDeltaT(): number {
-    return lastDeltaT
+    return STANDARD_SIMULATION_PERIOD
```
rutmanz: this should probably be removed. This name implies it is the last delta t and that it changes, when it is actually a constant (that can be referenced directly where this is used)

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1077,9 +1085,21 @@ class PhysicsSystem extends WorldSystem {
                     rn.isDynamic ? JOLT.EMotionType_Dynamic : JOLT.EMotionType_Static,
                     rnLayer
                 )
+                if (rn.isDynamic) {
+                    // prevents fast bodies from phasing through thin static geometry (like ramps on fields)
+                    bodySettings.mMotionQuality = JOLT.EMotionQuality_LinearCast
+                }
```
rutmanz: Dozer's ability to drive on 2026 ramps has not meaningfully changed to my eye. Do you have a reason to suspect this would help?

rutmanz (changes requested):
https://github.com/user-attachments/assets/3e91a110-9e8a-4676-a6b4-2a93dffbe8aa

Like @AlexD717 noticed, Dozer jitters now in a way it doesn't on dev (this is on my work laptop)

fission/src/systems/physics/PhysicsSystem.ts
Dhruv-0-Arora: I believe dozer will physically never be able to drive on the 2026 ramp @rutmanz (because of the geometry of the front shovel thing). I was having issues with the kitbit phasing through the ramp and thats what I was trying to fix.
This isn't a perfect fix though

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1077,9 +1085,21 @@ class PhysicsSystem extends WorldSystem {
...
+
                 const body = this._joltBodyInterface.CreateBody(bodySettings)
-                this._joltBodyInterface.AddBody(body.GetID(), JOLT.EActivation_Activate)
-                body.SetAllowSleeping(false)
+
+                // Game pieces are allowed to sleep and spawn inactive
```
azaleacolburn:
```suggestion
                // Game pieces are allowed to sleep, but are inactive by default
```

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1077,9 +1085,21 @@ class PhysicsSystem extends WorldSystem {
...
+                // they are placed at their resting position by MirabufSceneObject
```
azaleacolburn:
```suggestion
                // they are placed at their initial position by their `MirabufSceneObject`
```

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1077,9 +1085,21 @@ class PhysicsSystem extends WorldSystem {
...
+                // which activates them.
+                this._joltBodyInterface.AddBody(
+                    body.GetID(),
+                    rn.isGamePiece ? JOLT.EActivation_DontActivate : JOLT.EActivation_Activate
+                )
+                if (!rn.isGamePiece) body.SetAllowSleeping(false)
```
azaleacolburn, line 1103: This means that 2024 field performance will still be terrible? Why should only game pieces be allowed to sleep?

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1415,15 +1436,15 @@ class PhysicsSystem extends WorldSystem {
...
-        this._joltInterface.Step(lastDeltaT, substeps)
+        // step the simulation by rendered frame's duration so motion is smooth
+        const simTime = Math.min(deltaT, MAX_SIMULATION_TIME_PER_FRAME)
+        if (simTime > 0) {
+            const desiredSteps = Math.ceil(simTime / MAX_COLLISIONS_STEP_PERIOD)
+            const activeBodies = this._joltPhysSystem.GetNumActiveBodies(JOLT.EBodyType_RigidBody)
+            const affordableSteps = Math.floor(ACTIVE_BODY_STEP_BUDGET / Math.max(activeBodies, 1))
+            const collisionSteps = Math.max(1, Math.min(MAX_COLLISIONS_STEPS, desiredSteps, affordableSteps))
+            this._joltInterface.Step(simTime, collisionSteps)
+        }
```
azaleacolburn: Justification for this change? I imagine it was tuned through trial and error, but some explanation would be helpful for future tuning.

azaleacolburn (changes requested):

fission/src/systems/physics/PhysicsSystem.ts
Dhruv-0-Arora: How do you mean? In addition, the 2024 field performance went from 10 fps for me to 110 fps.

fission/src/systems/physics/PhysicsSystem.ts
Dhruv-0-Arora: Sorry, I tested dozer on the ramp and I understand what you are saying.

I'm not entirely sure how to 100% fix the phasing issue with dozer on that ramp. Its possible it could be a CAD issue.

Dhruv-0-Arora:
@AlexD717 @rutmanz I'm not able to reproduce the robot jitter that you experienced. Does dozer jitter when you initially spawn it for the first time on an empty field?

fission/src/systems/physics/PhysicsSystem.ts
rutmanz: No my question is really what should I test to see the improvement made by this change

Dhruv-0-Arora: I used 2026 kitbot on the new UI when I was testing

AlexD717: In my opinion, this should be in a separate PR. This is mostly unrelated to game piece sleeping, and it will make it a lot easier to test if this is actually helping (or just the higher FPS from the game pieces being at sleep) and the performance loss it is causing.

Dhruv-0-Arora: Good call

fission/src/systems/physics/PhysicsSystem.ts
AlexD717: @azaleacolburn, I think the majority of 2024 fields performance loss is a result of us having notes in the loading station and simulating them.

Screenshot 2026-07-20 092643

This also explains the performance increase from putting the game objects to sleep.

AlexD717:
> @AlexD717 @rutmanz I'm not able to reproduce the robot jitter that you experienced. Does dozer jitter when you initially spawn it for the first time on an empty field?

For me, this really only happens on the 2026 field after I go in and push some balls around (once the fps recovers the jittering starts to slow down and eventually goes away).  If you are struggling to reproduce it try throttling your CPU (it makes the jittering a lot worse for me).

fission/src/systems/physics/PhysicsSystem.ts
Dhruv-0-Arora: I attempted to fix the issue of our current implementation of timestepping causing the physics system to go into slow motion when you have a low framerate. 
This added some jittering bug though so im reverting it and making it its own PR where hopefully I can fix the jittering bug.

PR #1424, opened 2026-07-17
Batch Adding and Removing Bodies `[SYNTH-253]`, et al.

azaleacolburn:
## Task

SYNTH-168 (Symptom)
SYNTH-253 (Cause)

## Symptom

Spawning fields rapidly causes Jolt to crash. According to Jolt documentation, this can happen when there are too many bodies in the broadphase tree arranged in an unoptimized manner.

## Solution

According to the Jolt documentation, there are two ways to ensure optimization, the first is to batch adding bodies, so they can be added and arranged in a more efficient manner. The second is to call `OptimizeBroadPhase()` on the jolt body interface. There's significant overhead to the second option, and its not recommended with single-threaded Jolt, so instead, we batch adding bodies of a `MirabufSceneObject`.

[Source](https://arc.net/l/quote/yuphxwey)

## Verification

- Fission builds
- All tests pass.
- Performance for fields with many bodies is better on my machine.

> [!NOTE]
> Spawning the 2025 field then the 2026 field still causes a memory out of bounds error
---

- [x] User-facing polish:


jolt
rutmanz, line 1: what changed in Jolt

rutmanz (approved):
Seems to work, not sure if I can notice a change but doesn't seem to break anything

AlexD717 (approved):
lgtm. Merge conflicts in jolt

PR #1423, opened 2026-07-17
use uuids for input schemes [SYNTH-251]

0xda157:
## Task

SYNTH-251

## Verification

---

Before merging, ensure the following criteria are met:

- [x] All acceptance criteria outlined in the ticket are met.
- [ ] Necessary test cases have been added and updated.
- [x] A feature toggle or safe disable path has been added (if applicable).
- [x] User-facing polish:
  - Ask: *"Is this ready-looking?"*
- [x] Cross-linking between Jira and GitHub:
  - PR links to the relevant Jira issue.
  - Jira ticket has a comment referencing this PR.

PR #1433, opened 2026-07-22
Improve state management [SYNTH-263]

0xda157:
## Task

SYNTH-263

## Symptom

many of our components include `useReducer`s purely to avoid manage state in the correct react way

## Solution

this pr improve the state handling in many locations, primarily by copying the state using their `set` function instead of directly mutating them.

> [!NOTE]
> There are few more places with very poor state management, but I've deemed them too difficult to improve in this pr. Ideally they will be addressed later.

## Verification

Playwright tests pass, and the UI seems to work as expected.

---

- [x] User-facing polish:


azaleacolburn (approved):

fission/src/ui/panels/configuring/assembly-config/interfaces/inputs/ConfigureSchemeInterface.tsx
```diff
@@ -74,29 +77,27 @@ const ConfigureSchemeInterface: React.FC<ConfigSchemeProps> = ({ selectedScheme,
...
-                        selectedScheme.usesGamepad = false
-                    }
-                    selectedScheme.usesTouchControls = val
-                    selectedScheme.customized = true
+                    setSelectedScheme({
+                        ...selectedScheme,
+                        customized: true,
+                        usesTouchControls: val,
+                        usesGamepad: val ? false : selectedScheme.usesGamepad,
+                    })
```
Dhruv-0-Arora: The old code mutate the selected scheme object in place
this replaces it with a detached copy that never re-enters allInputSchemes

This introduced the bug that even after you save, certain items don't actually save such as:
- toggling use controller / use touch controls
- add joint control
- custom default scheme (though the default scheme will be removed in the ui-redesign.

Dhruv-0-Arora: When I tested this, I just created a custom scheme and checked `use controller` and saved the panel. It didn't save

fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -113,112 +109,113 @@ const ColorEditor: React.FC<{
...
+
+const GeneralTabCheckbox = ({
+    preference,
+    label,
+    ...props
+}: Omit<CheckboxProps, "checked" | "onClick"> & { preference: TypedPreferenceKey<boolean> }) => {
+    const [pref, setPref] = useState(PreferencesSystem.getUserPreference(preference))
+    useEffect(() => {
+        PreferencesSystem.setUserPreference(preference, pref)
+    }, [pref])
```
Dhruv-0-Arora, line 137: This will fire a preferenceChanged event for every preference on mount and twice in dev strictmode

instead, I would suggest
```ts
<Checkbox {...props} label={label} checked={pref} onClick={
setPref(val)
PreferencesSystem.setUserPreferences(preference, val)
} />
```

Dhruv-0-Arora (changes requested):

fission/src/ui/panels/configuring/assembly-config/interfaces/inputs/ConfigureSchemeInterface.tsx
0xda157: it seems like `useStateContext` is lying, `setSelectedScheme` doesn't actually do what we want. I'll work on fixing this.

fission/src/ui/modals/configuring/SettingsModal.tsx
0xda157: > This will fire a preferenceChanged event for every preference on mount and twice in dev strictmode

why is this an issue?

> instead, I would suggest
> 
> ```ts
> <Checkbox {...props} label={label} checked={pref} onClick={
> setPref(val)
> PreferencesSystem.setUserPreferences(preference, val)
> } />
> ```

this is a very strong react anti-pattern. If `setPref` gets called somewhere else, after this change, it will lead to the UI being in an invalid state.

fission/src/ui/panels/configuring/assembly-config/interfaces/inputs/ConfigureSchemeInterface.tsx
0xda157: seems like it isn't really gonna be possible until #1423, I'll revert the commit for now.

fission/src/ui/modals/configuring/SettingsModal.tsx
PepperLola: I agree with @0xda157; there's not much of a cost to the event being fired repeatedly since we have relatively few listeners and preferences, and it avoids ending up with an invalid UI state

Dhruv-0-Arora: So we are reading from the preferences and setting the preferences...

There are 13 preference events being triggered and it seems that 8 do not have listeners:
ReportAnalytics, SubsystemGravity, RenderSceneTag, RenderScoreboard, ShowViewCube

ReportAnalytics & SubsystemGravity might be problematic:
-> ReportAnalytics calls sendMetaData() [eventually] -> sending to google analytics
-> calls GetMotorSettings() and rewrites 2 torque limits on a constraint

@PepperLola @0xda157

0xda157: I'm not sure what's the best way to handle the analytics issue, but getting a preference and then immediately setting it to the same value shouldn't be causing issues aside from that.

0xda157: here's what react has to say on the matter: https://react.dev/learn/synchronizing-with-effects#sending-analytics

AlexD717 (approved):
lgtm

Dhruv-0-Arora (approved):

fission/src/ui/modals/configuring/SettingsModal.tsx
```diff
@@ -113,112 +109,113 @@ const ColorEditor: React.FC<{
...
-                onClick={checked => writePreference("RenderSceneTags", checked)}
-                tooltip="Name tags above robot."
-            />
-            <Checkbox
-                label="Show Scoreboard"
-                checked={PreferencesSystem.getUserPreference("RenderScoreboard")}
-                onClick={checked => {
-                    writePreference("RenderScoreboard", checked)
-                    if (checked) {
-                        // TODO: figure out scoreboard - I think it should be its own component and not a panel
```
BrandonPacewic, line 199: Should purge all the `TODO`s from this codebase.

BrandonPacewic (approved):

PR #1435, opened 2026-07-22
resolve most biome warnings [SYNTH-162]

0xda157:
## Task

SYNTH-162

## Symptom

many warnings, most about dependency array being incorrect 

## Solution

add missing state to dependency arrays, and lift constants out of components. 

## Verification

`bun run build` passes, almost all lint errors are gone, application seems to function as intended.

---

- [x] User-facing polish:


AlexD717 (approved):
Dropped from 120 lint warning on dev to 30 here. All functionality appears to be the same

azaleacolburn (approved):
I didn't notice anything different. Every code change seems to make sense.

fission/biome.json
```diff
@@ -70,7 +70,7 @@
         "noNamespace": "error",
         "useArrayLiterals": "error",
         "useAsConstAssertion": "error",
-        "useComponentExportOnlyModules": "warn",
```
rutmanz, line 73: Why do we not want this rule?

rutmanz (changes requested):
Looks like there are a few more import `type` fixes that can be automatically applied, and looking through the list there was a simple change to resolve one more dependency array issue (src/ui/ThemeProvider.tsx:95) 
```
        setThemeOptions((themeOptions) =>({
            ...themeOptions,
            primary: {
                main: primaryColor,
            },
            secondary: {
                main: secondaryColor,
            },
            blueAlliance: {
                main: blueAllianceColor,
            },
            redAlliance: {
                main: redAllianceColor,
            },
        }))
```

fission/biome.json
0xda157: I'm ignoring it because we use it a lot, there isn't much harm in doing it, and not doing it would make our code less idiomatic imo.

rutmanz: Why do you say there isn't much harm? My understanding is that it prevents HMR for changes to those files / related files

0xda157: hm I guess we should probably not ignore it then. I'll put it back to being a warning and create a ticket for that.

rutmanz (approved):

PR #1436, opened 2026-07-22
Multiple Controller Support `[SYNTH-151]`

RoushilS:
## Task

SYNTH-151

Allow multiple gamepads to each control individual robots

## Symptom

Situation: User wants to simulate full FRC match (six robots) on one computer, not online multiplayer. They discover that Synthesis has a limit of one gamepad, which can only control one robot (theoretically more robots but would be poor ux). They also find that the alternative, having five robots controlled on one keyboard, is infeasible. With this new feature, a computer like a large gaming laptop to a PC with sufficient ports could allow for six controllers.

## Solution

Reworked `Input` and `InputSystem`'s gamepad support to have an array of gamepads that accepts `playerSlots`s. Additionally updated all used gamepad methods to accept a playerSlot. Updated `defaultInputs` to also have a playerSlot parameter, and implemented custom scheme logic. In the `EditInputInterface`, add a simple dropdown for a user to customize gamepad slot, as well as syncing logic. Touched `InputSystemTest` as it was throwing an error.

## Verification

> [!NOTE]
> I have one gamepad to lend to testers, and can find one more, so you don't necessarily need a gamepad of your own to test.

- [ ] Multiple gamepads can be used to control different robots
- [ ] Schemes can be customized and created
- [ ] No unintended functionality is introduced
- [ ] Existing gamepad functionality is not harmed

---

- [ ] User-facing polish:


AlexD717 (changes requested):
While the core functionality is now implemented, I feel like the UI can still be significantly improved.

Having the toggle in the Configure -> Inputs panel is good, but there should also be one in the Spawn Asset panel, alongside the input selection.

Additionally, instead of a simple dropdown, we should use a separate panel or modal to select and easily identify controllers. For example, it could highlight the controller when an input is received. With the current dropdown, I can't tell Controller 1 from Controller 2 without testing them, a problem that gets much worse if you scale up to 6 controllers.

Finally, with the current input selection, if I spawn a robot and assign it the "Jax" input layout (or any other controller layout), it flags an interference, assuming someone else is already using it. As a result, I can't assign that same "Jax" layout to a second controller.

fission/src/ui/panels/configuring/initial-config/InputSchemeSelection.tsx
```diff
@@ -134,6 +182,44 @@ export default function InputSchemeSelection({ brainIndex, onSelect, panelId }:
...
+                            ))}
+                        </Select>
+                    </FormControl>
+                    {/** Live readout of which controllers are supplying input, to identify them */}
+                    <Label size="sm" className="text-center mt-[4pt] mx-[5%]">
+                        {`Current Controller Input: ${
+                            activeControllers.length > 0
+                                ? activeControllers.map(slot => `Controller ${slot + 1}`).join(", ")
+                                : "None"
+                        }`}
```
AlexD717: This text is just slightly too long so it causes the Spawn Asset panel to slightly expand in size when moving a controller.

I would either change it to just "Current Input: Controller 1" or "Current Controller Input: 1"

AlexD717 (changes requested):
Looks a lot better, but there a few more minor requests I have.

If I am trying to spawn a robot that already has a controller assigned, the other controller inputs that have interferences should be shown in a darker green

Image

rutmanz:
(I don't have any controllers to test with)

rutmanz (changes requested):
Image
Two robots with same scheme on different controllers should probably have different names

---

If you can call them something other than controller 1 and controller 2 (like maybe more specifically with the name descriptor of possible)

—-

It looks like it only detects controllers after they give an input, is there a way to query them without this? (other joystick tester sites do this)

AlexD717 (changes requested):
Sorry for not seeing this before, but when you are configuring a specific input scheme (you need to merge dev to fix a bug with opening that panel) only Controller 1 is able to edit it. Since its a global configuration everyone should be able to change it.

In addition, it doesn't seem that @/rutmanz comments about more descriptive controller names was resolved (did you end up discussing that in person and coming to a agreement?). Controllers still show up as Controller 1, Controller 2, etc. rather than Xbox 1, PS5 2

RoushilS:
I believe that Zach and I came to the agreement that descriptive naming was not necessary, @rutmanz feel free to correct me though.

rutmanz:
> I believe that Zach and I came to the agreement that descriptive naming was not necessary, @rutmanz feel free to correct me though. 

I could be wrong but I thought we talked about the auto-detection, which I was wrong about.

I think the descriptive name thing would still be a good idea, but I don't care that much

RoushilS:
> I could be wrong but I thought we talked about the auto-detection
>

Oh, apologies about that. I must have misremembered.

I would be willing to work on it further now if it's necessary, but prefer to skip and revisit later on. This is because when I looked into it, there were quite a few naming/indexing issues that popped up. Since we are coming to the end of the summer and I feel like this is important to get in, would it be okay to omit the descriptive controller naming on this PR? 

Also thanks for the reply!

fission/src/systems/input/InputSystem.ts
```diff
@@ -99,9 +139,20 @@ class InputSystem extends WorldSystem {
...
+
+            // Safely verify the index is a valid number slot before querying rawGamepads
+            if (lookupIndex !== null && lookupIndex !== undefined) {
+                if (rawGamepads[lookupIndex] == null) {
+                    InputSystem.gamepads[lookupIndex] = null
+                } else {
+                    InputSystem.gamepads[lookupIndex] = rawGamepads[lookupIndex]
+                }
+            }
+        }
```
PepperLola:
```suggestion
        for (const lookupIndex of InputSystem._gpIndexes) {
            if (lookupIndex === null) continue;
            InputSystem.gamepads[lookupIndex] = rawGamepads[lookupIndex]
        }
```

I think this should be equivalent? `lookupIndex` should never be undefined because `InputSystem._gpIndexes` is `Gamepad | null`, so we can just do an early continue rather than wrapping the rest of the body in an if. In both cases of the `rawGamepads[lookupIndex]` check, you're setting the value of `InputSystem.gamepads[lookupIndex]` to what you're asserting `rawGamepads[lookupIndex]` so the check is redundant.

fission/src/systems/input/InputSystem.ts
```diff
@@ -163,16 +214,17 @@ class InputSystem extends WorldSystem {
             )
         }
 
-        InputSystem._gpIndex = event.gamepad.index
+        const index = event.gamepad.index
+        InputSystem._gpIndexes[index] = index
```
PepperLola: This now looks entirely redundant. If we're indexing `_gpIndexes` and `gamepads` by `index` and the value within `_gpIndexes` is also `index`, then what's the point of having `_gpIndexes`? It's essentially an identity function - if you already have the index you want to look up in `_gpIndexes`, then getting the value within `_gpIndexes` will just give you that index back.

fission/src/systems/input/InputSystem.ts
```diff
@@ -223,16 +279,34 @@ class InputSystem extends WorldSystem {
         )
     }
 
+    /** @returns An array of all currently connected, active Gamepad objects. */
+    public static getConnectedGamepads(): Gamepad[] {
+        return this._gpIndexes
+            .map(index => (index !== null ? this.gamepads[index] : null))
+            .filter((gamepad): gamepad is Gamepad => gamepad != null)
```
PepperLola:
```suggestion
        return this.gamepads.filter((gamepad): gamepad is Gamepad => gamepad !== null)
```

Is this equivalent?

fission/src/systems/input/InputSystem.ts
```diff
@@ -223,16 +279,34 @@ class InputSystem extends WorldSystem {
...
+    }
+
+    /**
+     * @param {number} playerSlot The logical player slot.
+     * @returns {Gamepad | null} The gamepad in that slot, or null if the slot is unoccupied.
+     */
+    public static getGamepadBySlot(playerSlot: number): Gamepad | null {
+        const rawIndex = InputSystem._gpIndexes[playerSlot]
+        if (rawIndex == null) return null
+        return InputSystem.gamepads[rawIndex] ?? null
```
PepperLola:
```suggestion
        return InputSystem.gamepads[playerSlot]
```

Given how `_gpIndexes` is constructed, `InputSystem._gpIndexes[playerSlot]` will just return `playerSlot`, meaning you can skip the lookup entirely. The `?? null` is also only important if `InputSystem.gamepads[playerSlot]` could return undefined and you want to guarantee the function returns null; in this case I don't think that's possible.

fission/src/ui/panels/configuring/initial-config/InputSchemeSelection.tsx
```diff
@@ -98,23 +98,71 @@ export default function InputSchemeSelection({ brainIndex, onSelect, panelId }:
...
+        // Poll each frame for which controllers are actively supplying input, so users can
+        // physically identify a controller by pressing a button or moving a stick on it.
+        const deadband = 0.15
+        let frame: number
+        let previousKey = ""
+
+        const poll = () => {
+            const active: number[] = []
+            const count = InputSystem.getConnectedPlayerCount()
+            for (let slot = 0; slot < count; slot++) {
```
PepperLola: I don't think this will work. Say 4 players connect controllers, and the second disconnects. `gamepads` will look like `[<gamepad 0>, null, <gamepad 2>, <gamepad 3>]` with gamepad 1 gone now. `poll` will see that 3 gamepads are connected and only then check the first 3 slots, which would be gamepad 0, `null`, and gamepad 2, leaving out gamepad 3 entirely. Instead this should probably just iterate over `InputSystem.gamepads` because you're immediately getting the gamepad anyway.

It would be good to concretely define what a "slot" is - is it the index within `InputSystem.gamepads`? Which, based on how `gamepads` is currently constructed, would also be the index provided by the browser.

fission/src/ui/panels/configuring/initial-config/InputSchemeSelection.tsx
```diff
@@ -160,6 +208,44 @@ export default function InputSchemeSelection({ brainIndex, onSelect, panelId }:
...
+                            labelId="input-scheme-controller-label"
+                            value={selectedSlot}
+                            onChange={e => {
+                                const slot = Number(e.target.value)
+                                setSelectedSlot(slot)
+                                InputSystem.setPlayerSlot(brainIndex, slot)
+                                EventSystem.dispatch("InputSchemeChanged", { panelId })
+                            }}
+                        >
+                            {Array.from({ length: connectedPlayerCount }, (_unused, slot) => (
```
PepperLola: Same issue here as above - if a player in the middle disconnects then `connectedPlayerCount` will decrease and only the first `connectedPlayerCount` controllers will display, regardless of whether they're connected and whether it would be leaving out a controller that is connected.

PepperLola (changes requested):
I left a couple comments - some are style changes but most are functional. I also would like you to remove the SyntheSimFTC changes from the diff since those are unrelated.

PR #1439, opened 2026-07-23
fix(ui): adding name of robot on right click `SYNTH-271`

Dhruv-0-Arora:
SYNTH-271

## Task

Right clicking a robot would title it "A Robot"
Fields would say "A Field"

## Solution

Replace `"A Robot"` with the scheme name (`Ernie` <- no brackets [])
Replace `"A Field"` with `"Field"`

Screenshot 2026-07-22 at 10 41 57 AM

## Verification

Check whether spawned robots have their scheme name. 

# Breaking change

in this new function:
```ts
    public get schemeName() {
        return this.multiplayerOwnerName ?? (this.brain as SynthesisBrain)?.inputSchemeName
    }
```
if a robot is not configured, it will pass back "Not Configured" instead of `undefined` so in the configure assets panel, it will replace `[-] ...` with`[Not Configured] ...`

The purpose of this change was to avoid

---

- [ ] User-facing polish:


fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -204,13 +203,19 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
     }
 
     public get descriptiveName(): string {
-        return `${this.miraType === MiraType.ROBOT ? `[${this.multiplayerOwnerName ?? InputSystem.brainIndexSchemeMap.get((this.brain as SynthesisBrain).brainIndex)?.schemeName ?? "-"}] ` : ""}${this.assemblyName}`
+        return `${this.miraType === MiraType.ROBOT ? `[${this.schemeName}] ` : ""}${this.assemblyName}`
```
AlexD717: Why are we changing descriptiveName to no longer have multiplayer stuff?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -1212,7 +1217,7 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
 
     public getSupplierData(): ContextData {
         const data: ContextData = {
-            title: this.miraType == MiraType.ROBOT ? "A Robot" : "A Field",
+            title: this.miraType == MiraType.ROBOT ? `Robot: ${this.schemeName}` : "Field",
```
AlexD717: I think it makes more sense to use descriptiveName rather than schemeName for consistency.

AlexD717 (changes requested):
Can you add a bit more explanation about why we are going with this name format instead of the already established descriptiveName?

fission/src/mirabuf/MirabufSceneObject.ts
PepperLola: It still does; the default is `this.multiplayerOwnerName` in the `schemeName` getter, then it uses the input scheme name or "Magic" if that's null/undefined

fission/src/mirabuf/MirabufSceneObject.ts
Dhruv-0-Arora: I choose scheme name because descriptive name was too long for the menu

0xda157: schemeName seems like it could be confusing, especially after #1423 is merged

Dhruv-0-Arora: Screenshot 2026-07-24 at 1 15 01 PM

This is what happens otherwise

AlexD717: IMO that looks fine. Also I think its best if you remove the Robot prefix and just have it be like this
image

image

Dhruv-0-Arora: @AlexD717 @0xda157 I changed it to descriptiveName. I did keep the type safety SynthesisBrain check

AlexD717 (approved):
lotm

rutmanz (approved):

BrandonPacewic (approved):

PR #1442, opened 2026-07-23
Don't Pause Audio on Mobile Chromium `[SYNTH-240]`

AlexD717:
## Task

SYNTH-240

## Symptom

On mobile chromium based browser whenever we played a sound effect any music/video playing the background would either stop or have its volume reduces for the sound effect duration which is annoying (since we only ever play sound effects that are short).

## Solution

Switch to using AudioBuffer (designed for short SFX) instead of HTMLAudioElement. This causes Chromium to play this sound over the currently playing audio instead of pausing it. 

For Safari we change the audio session type to `ambient` which tells it to mix this audio with currently playing sound effects (only supported in Safari currently). 

Firefox was never pausing other audio for me, and still isn't.

## Verification

Clicking buttons with music playing in the background doesn't cause the music to pause or lower in volume for the following browsers

- [x] Chrome on Android
- [x] Firefox on Android
- [x] Chrome on IOS
- [x] Firefox on IOS
- [x] Safari on IOS

---

- [x] User-facing polish:


0xda157 (approved):
LGTM, thanks for fixing this. we should test on iOS before merging.

fission/src/systems/sound/SoundPlayer.ts
```diff
@@ -31,36 +37,51 @@ export class SoundPlayer {
...
-                }
-            },
-            { once: true }
-        )
-        return audio.play().catch(error => {
+    public async play(filePath: string): Promise<void> {
+        try {
+            const buffer = await this.getBuffer(filePath)
+            if (this._audioContext.state === "suspended") {
+                console.log("Resuming audio context")
```
Dhruv-0-Arora:
```suggestion
```

fission/src/systems/sound/SoundPlayer.ts
```diff
@@ -31,36 +37,51 @@ export class SoundPlayer {
...
+                await this._audioContext.resume()
+            }
+
+            const gain = this._audioContext.createGain()
+            gain.gain.value = this._currentVolume
+            gain.connect(this._audioContext.destination)
```
Dhruv-0-Arora, line 60: gain node is never disconnected.. this isn't a real leak though

fission/src/systems/sound/SoundPlayer.ts
```diff
@@ -31,36 +37,51 @@ export class SoundPlayer {
...
+
+            const source = this._audioContext.createBufferSource()
+            source.buffer = buffer
+            source.connect(gain)
```
Dhruv-0-Arora, line 64: source never disconnected.. also not really a leak

fission/src/systems/sound/SoundPlayer.ts
```diff
@@ -31,36 +37,51 @@ export class SoundPlayer {
...
+        } catch (error) {
             console.error("Error playing the audio file:", error)
-        })
+        }
     }
 
     private isPlaying(filePath: string): boolean {
-        const audio = this._active.get(filePath)
-        return audio != null && !audio.ended && !audio.paused
+        return this._active.has(filePath)
```
Dhruv-0-Arora, line 85: could be false if clickdown buffer isn't decoded yet?

this._active.set() only happens after `await this.getBuffer(filePath)` now. Potential race condition to consider
preloadSounds though should mean that this is not an issue at all.

Dhruv-0-Arora (approved):
nothing blocking

rutmanz:
I will test on iOS today at home

fission/src/systems/sound/SoundPlayer.ts
AlexD717: I don't think this is a big issue since we are preloading sounds (as you mentioned)

rutmanz:
iOS Safari does not exhibit on dev, no change
iOS Chrome does exhibit on dev, fixed on this branch
iOS Firefox does exhibit on dev, fixed on this branch

rutmanz (approved):
lqgtm

PR #1444, opened 2026-07-23
Fix Nix + Playwright [SYNTH-209]

0xda157:
## Task

SYNTH-209

## Symptom

nix users are unable to get playwright via the nix shell

## Solution

add playwright back and update the flake inputs. likely the issue was the locked version of nixpkgs had an old playwright version

## Verification

things work locally now! I also added a github actions workflow to prevent this from happening in the future.

---

- [x] User-facing polish:


.github/workflows/NixTest.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: cachix/install-nix-action@v31
+
+      - name: Run Playwright Tests
+        working-directory: "fission"
+        run: |
+          nix develop .#fission --command bash -c "bun run assetpack && bun run test"
```
rutmanz: Maybe add a timeout on this or configure to only run one test or test file? I don't think it makes sense to run unit tests twice

rutmanz (changes requested):

azaleacolburn (approved):

azaleacolburn (approved):
Tests run for me

AlexD717 (approved):
lgtm

.github/workflows/NixTest.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: cachix/install-nix-action@v31
+
+      - name: Run Playwright Tests
+        working-directory: "fission"
+        run: |
+          nix develop .#fission --command bash -c \
+            "git lfs pull && unzip public/assetpack.zip -d public && bun install && bun run test run ./src/test/APSTesting.test.ts"
```
rutmanz: If you're just running APS tests, you probably don't need the assetpack. Also does it not make sense use the node_modules cache here?

rutmanz (commented):
I'm just trying to minimize the compute time these tests take, I'm not sure if Autodesk has to pay for it or not but if we're able to do this more efficiently without sacrificing the test quality that would be nice

rutmanz (approved):
tm,lg

.github/workflows/NixTest.yml
BrandonPacewic, line 1: Will this mean that all of our unit tests will run twice on each PR push?

0xda157: one of units tests will, not all

PR #1446, opened 2026-07-24
Mirabuf Scene Object OBB memory fixes `[SYNTH-278]`

BrandonPacewic:
## Task

SYNTH-278

## Symptom

Problems related to how we deal with jolt memory within our object bounding box logic.

## Solution

Used a Asan build to directly confirm where the memory issues were and fixed them.

## Verification

- [ ] Spawn a robot with the COM indicator enabled, confirm it works.
- [ ] Spawn a robot in a field with protected/scoring zones, confirm zone-overlap detection still works and OBB tracks robot position/rotation correctly.

---

- [ ] User-facing polish:


fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -321,8 +314,12 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
         this._basePositionTransform = this.getXZPositionTransform()
 
         if (this.miraType === MiraType.ROBOT) {
-            this.computeFurthestVertices()
-            this.computeUnrotatedRootNodeToCenterPositionTranslation()
+            this._cachedOrientedBoundingBoxDimensions = this.getDimensionsWithoutRotation()
```
azaleacolburn: Why are we doing this. The plan was for `getDimensionsWithoutRotation` to be phased out, not the other way around.

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -132,8 +125,8 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
 
     private _collisionUnsubscriber?: () => void
 
-    private _furthestVertices?: AxisVertices = undefined
-    private _unrotatedRootNodeToCenterPositionTranslation?: Jolt.Vec3 = undefined
+    private _cachedOrientedBoundingBoxDimensions?: { width: number; height: number; depth: number }
+    private _cachedRootToCenterOffset?: THREE.Vector3
```
azaleacolburn: Why are we making this a three object?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -321,8 +314,12 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
 
         if (this.miraType === MiraType.ROBOT) {
-            this.computeFurthestVertices()
-            this.computeUnrotatedRootNodeToCenterPositionTranslation()
+            this._cachedOrientedBoundingBoxDimensions = this.getDimensionsWithoutRotation()
+
+            const rootBody = World.physicsSystem.getBody(this.getRootNodeId()!)!
+            const rootPosition = rootBody.GetPosition() // STATIC_ALIAS
+            const rootPositionThree = new THREE.Vector3(rootPosition.GetX(), rootPosition.GetY(), rootPosition.GetZ())
+            this._cachedRootToCenterOffset = this.getPositionTransform().sub(rootPositionThree)
```
azaleacolburn: What was wrong with this being in its own function?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -562,41 +554,33 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
-                        const oldWeighedCOM = weightedCOM
-
                         const mass = 1 / inverseMass
-                        const com = body.GetCenterOfMassPosition().Mul(mass)
+                        const comPosition = body.GetCenterOfMassPosition() // STATIC_ALIAS
 
-                        weightedCOM = weightedCOM.AddRVec3(com)
+                        weightedCom.x += comPosition.GetX() * mass
+                        weightedCom.y += comPosition.GetY() * mass
+                        weightedCom.z += comPosition.GetZ() * mass
```
azaleacolburn: Aren't there easier ways to add to a three vector?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -767,146 +751,26 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
-     */
-    private computeFurthestVertices(): void {
-        this._furthestVertices = {
-            x: { min: Number.POSITIVE_INFINITY, max: Number.NEGATIVE_INFINITY },
-            y: { min: Number.POSITIVE_INFINITY, max: Number.NEGATIVE_INFINITY },
-            z: { min: Number.POSITIVE_INFINITY, max: Number.NEGATIVE_INFINITY },
-        }
-
-        const inverseRotation = this.getInverseRotationOfBody()
-
```
azaleacolburn, line 829: Why does this function need to be removed?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -132,8 +125,8 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
 
     private _collisionUnsubscriber?: () => void
 
-    private _furthestVertices?: AxisVertices = undefined
-    private _unrotatedRootNodeToCenterPositionTranslation?: Jolt.Vec3 = undefined
+    private _cachedOrientedBoundingBoxDimensions?: { width: number; height: number; depth: number }
```
azaleacolburn: I assume this is just to save having to perform 3 subtractions?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -767,146 +751,26 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
-
-        JOLT.destroy(scale)
-        JOLT.destroy(biggest)
-        JOLT.destroy(identity)
-
-        const mins = [this._furthestVertices.x.min, this._furthestVertices.y.min, this._furthestVertices.z.min]
-        const maxes = [this._furthestVertices.x.max, this._furthestVertices.y.max, this._furthestVertices.z.max]
-        if (mins.some(m => m === Number.POSITIVE_INFINITY) || maxes.some(m => m === Number.NEGATIVE_INFINITY)) {
-            console.warn("Failed to compute furthest vertices")
+        const rotationThree = new THREE.Quaternion(rotation.GetX(), rotation.GetY(), rotation.GetZ(), rotation.GetW())
```
azaleacolburn: Why not use a conversion function?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -767,146 +751,26 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
+        const rootPosition = rootBody.GetPosition() // STATIC_ALIAS
+        const centerThree = new THREE.Vector3(rootPosition.GetX(), rootPosition.GetY(), rootPosition.GetZ())
+        if (this._cachedRootToCenterOffset) {
```
azaleacolburn: Why not just calculate this value in `setup` and then never have to make this check?

fission/src/mirabuf/MirabufSceneObject.ts
```diff
@@ -767,146 +751,26 @@ class MirabufSceneObject extends SceneObject implements ContextSupplier {
...
-        // NOTE Do not destroy
-        const rotation = rootBody.GetRotation()
-
-        // We rotate our vector by the rotation of the root body, otherwise any rotation will mess with the translation
-        const offset = rotation.MulVec3(this._unrotatedRootNodeToCenterPositionTranslation!)
-
-        // Finally, we just offset the root node to get the true center
-        const position = convertJoltRVec3ToJoltVec3(rootBody.GetPosition().Add(offset))
-        const transform = JOLT.Mat44.prototype.sRotationTranslation(rotation, position)
+        const center = convertThreeVector3ToJoltVec3(centerThree)
```
azaleacolburn: Why not just make it a jolt object to begin with?

fission/src/mirabuf/MirabufSceneObject.ts
azaleacolburn: This should probably be a type here

azaleacolburn (changes requested):
I have several thoughts, but most of all I'm just confused about the motivation behind some of these changes.

1. I agree with the memory management changes, thank you for those.
2. I don't like your usage of three objects. I seems to me like you're doing all your math with three objects and then just converting to jolt at the end. Using three objects makes it seem like they are intended for use in the scene rendering system. You're making the code less readable for the sake of not having to clone a jolt quaternion? If anything, jolt quaternion application should be faster enough to make up the difference. Plus, you're optimizing in places that almost certainly aren't bottlenecks.
3. I have some other more specific thoughts that I left as comments.

fission/src/mirabuf/MirabufSceneObject.ts
BrandonPacewic: > I assume this is just to save having to perform 3 subtractions?

Yep.

fission/src/mirabuf/MirabufSceneObject.ts
BrandonPacewic: The rotation inverse is just now computed via quaternion `.invert()`.

azaleacolburn (approved):
We need a better way of handling static aliases besides comments

AlexD717 (approved):
Nothing seems broken

PR #1447, opened 2026-07-24
Mirabuf mass override is a no-op `[SYNTH-281]`

BrandonPacewic:
## Task

SYNTH-281

## Symptom

Mass data from mirabuf was not being applied correctly.

```ts
shape.GetMassProperties().mMass = mass
```

mutates a throwaway copy, never saves the mass override data.

## Solution

The solution is a bit weird because jolt appears to store mass data as a inverse and the only way to really set the mass is to get the inverse and then invert it?

## Verification

Robot and gamepiece mass now actually reflects mirabuf's declared mass instead of falling back to Jolt's density-based estimate. 

---

- [ ] User-facing polish:

### DIFF
--- a/fission/src/systems/physics/PhysicsSystem.ts
+++ b/fission/src/systems/physics/PhysicsSystem.ts
@@ -1063,9 +1063,9 @@ class PhysicsSystem extends WorldSystem {
                             appliedSphereCollider = true
                         }
 
-                        massOverride = totalMass == 0.0 ? 1 : Math.min(totalMass, MAX_GP_MASS)
+                        massOverride = totalMass == 0.0 ? undefined : Math.min(totalMass, MAX_GP_MASS)
                     } else {
-                        massOverride = totalMass == 0.0 ? 1 : totalMass * massMod
+                        massOverride = totalMass == 0.0 ? undefined : totalMass * massMod
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


AlexD717 (approved):
lotm

fission/src/systems/physics/PhysicsSystem.ts
```diff
@@ -1062,10 +1063,9 @@ class PhysicsSystem extends WorldSystem {
                             appliedSphereCollider = true
                         }
 
-                        const mass = totalMass == 0.0 ? 1 : Math.min(totalMass, MAX_GP_MASS)
-                        shape.GetMassProperties().mMass = mass
+                        massOverride = totalMass == 0.0 ? 1 : Math.min(totalMass, MAX_GP_MASS)
                     } else {
-                        shape.GetMassProperties().mMass = totalMass == 0.0 ? 1 : totalMass * massMod
+                        massOverride = totalMass == 0.0 ? 1 : totalMass * massMod
```
rutmanz: why is the mass override 1 if the total mass is 0? https://jrouwe.github.io/JoltPhysics/class_mass_properties.html says that bodySettings.mMassPropertiesOverride.mMass (where this is later used) should be the mass in kg

BrandonPacewic: This predated this pr, it just never worked before and now it does. I updated it such that if we have an undefined mass it defaults to the jolt material density mass which we were already using before.

rutmanz (approved):
thanks bestie

PR #1449, opened 2026-07-24
Bootstrap pip on both platforms

RoushilS:
## Task

SYNTH-283

## Symptom

Launching the exporter on a Windows Fusion install whose bundled Python lacks pip fails immediately with `No module named pip` at the `pip freeze` step, aborting dependency resolution. This surfaces after any Fusion update, since each update lands in a fresh `webdeploy` folder with a clean Python that may not include pip.

## Solution

Generalized the pip bootstrap in `Dependencies.py` to run on both Windows and macOS instead of macOS only, preferring the stdlib `ensurepip` (no network required) and falling back to downloading `get-pip.py`. Pip presence is now detected by running `pip --version` rather than checking for a file on disk, so the check is reliable across platforms.

## Verification

Manually test in-app (reproducing requires a pip-less Fusion Python, which can't be forced on demand); the change is limited to the bootstrap path and preserves existing behavior when pip is already present.

---

- [ ] User-facing polish:


exporter/SynthesisFusionAddin/src/Dependencies.py
```diff
@@ -55,6 +55,40 @@ def executeCommand(*args: str) -> subprocess.CompletedProcess[str]:
...
+        executeCommand(pythonExecutablePath, "-m", "ensurepip", "--upgrade")
+        if pipInstalled(pythonExecutablePath):
+            return
+    except subprocess.CalledProcessError:
+        logger.warning("ensurepip unavailable, falling back to get-pip.py")
+
+    # Fallback: download and run get-pip.py (curl ships with both modern Windows and macOS).
+    pipInstallScriptPath = os.path.join(pythonFolder, "get-pip.py")
+    if not os.path.exists(pipInstallScriptPath):
+        executeCommand("curl", "https://bootstrap.pypa.io/get-pip.py", "-o", pipInstallScriptPath)
```
Dhruv-0-Arora: I would recommend adding the -FsSL flag to curl

Dhruv-0-Arora (approved):
looks good - I didn't test

AlexD717 (approved):
lgtm, can now install Synthesis Exporter on my Windows PC (haven't tested it on a fresh mac though)

exporter/SynthesisFusionAddin/src/Dependencies.py
RoushilS: unfortunately looks like I forgot to push this change

azaleacolburn (approved):

PR #1454, opened 2026-07-28
Add favorites tab to library Modal `[SYNTH-269]`

RoushilS:
## Task

SYNTH-269

## Symptom

Users would have to click through years in the library assembly multiple times, which can be annoying. 

## Solution

Add new tab with frontent favorite display logic, as well as backend default loading of selected assets. Also briefely updated tutorial.

## Verification

- [ ] Users can favorite and unfavorite assets
- [ ] Best assets are favorited by default.
- [ ] Existing functionality preserved

---

- [ ] User-facing polish:


fission/src/mirabuf/DefaultAssetLoader.ts
```diff
@@ -5,8 +5,14 @@ import { API_URL } from "@/util/Consts.ts"
 export type DefaultAssetInfo = Required<Pick<MirabufCacheInfo, "hash" | "remotePath" | "miraType" | "name">> & {
     year?: number
     thumbnail?: string
+    /** Whether this asset is one of our curated picks */
+    defaultFavorite: boolean
 }
 
+const DEFAULT_FAVORITE_NAMES = new Set<string>(["Dozer v11", "KitBot (2026)", "FRC Field 2026 v2"])
+
+const isDefaultFavorite = (filename: string): boolean => DEFAULT_FAVORITE_NAMES.has(filename.replace(/\.mira$/, ""))
```
rutmanz: Why not make this part of the assetpack manifest? Then it would be easier to update / clearer when looking at the rest of the assets. You could still hardcode the names but I think it would make more sense for it to be in the update_manifest file so then it's determined by the asset source and not the frontend (imagine we would want the new kitbot and field to be favorites, but not push a frontend deploy to update that)

fission/src/systems/preferences/PreferenceTypes.ts
```diff
@@ -30,6 +30,8 @@ export type UserPreferences = {
     MultiplayerUsername: string
     MultiplayerClientID: string
     HasSeenOnboardingTour: boolean
+    FavoritedAssemblies: string[]
+    UnfavoritedAssemblies: string[]
```
rutmanz: I think it might make more sense (and make your data structures more efficient) to do `AssemblyFavoriteStatus: Record<string, "favorited"|"unfavorited"|undefined>` so it's all one structure and you can't have an invalid state (like an assembly that is both favorited and unfavorited)

rutmanz (changes requested):
Image

Are the year tabs supposed to be gone?

Also I think it might be better to have the favorites tab always present (either as "Favorites" or with the star icon) and just show "No favorited assets yet!" if empty.

Also it looks like the defaults don't work quite right, when I opened it, nothing from the assetpack was favorited.

Also maybe auto-favorite local uploads because they'll probably be likely to use the models they add themselves

Dhruv-0-Arora:
@rutmanz you have to clear storage if the thumbnails + years are not showing up

rutmanz (commented):
everything else works but the auto-favoriteing of uploads and some defaults (dozer, for example) still isn't happening for me

RoushilS:
I found the issue. You must run 

`git pull && bun run assetpack`

and then restart the bun server.

rutmanz:
> I found the issue. You must run
> 
> `git pull && bun run assetpack`
> 
> and then restart the bun server.

looks like the defaults are there now but uploading doesn't favorite them

Dhruv-0-Arora:
@RoushilS is this ready for review? (ignoring the conflicts)

RoushilS:
No, sorry. I could use some help though if possible. I can't figure out why auto favoriting on default doesn't work. Default assets should be okay though. The starring system is good though,

RoushilS:
I have another, more updated version of this branch that I have been working with more recently. I can force push it to here.

Dhruv-0-Arora:
Yeah, I can take a look

PR #1455, opened 2026-07-28
User can mark components as wheels `[SYNTH-198]`

BrandonPacewic:
## Task

SYNTH-198

## Symptom

If a user did not have a cad model with properly jointed wheels they could not drive or use their robot in Synthesis without first making potentially non-trivial changes to their CAD.

## Solution

Allow for the user to select components on their robot and mark them as wheels after the import step.

> [!note]
> This currently only works for arcade drive.

## Verification

- Import a robot that does not have wheels already assigned, say the [2026 Kitbot via URDF](https://github.com/user-attachments/files/30474539/Kitbot.2026.zip).
- It won't drive on the initial import.
- Select the wheels of the kitbot, be sure to select the outer most component of each wheel.
- Wait for the assembly to be rebuilt, then drive!

---

- [ ] User-facing polish:


AlexD717 (changes requested):
Uploading the given URDF Kitbot and assigning wheels results in the back of it floating up for me (as seen in the attached video).

https://github.com/user-attachments/assets/383bd784-441d-4708-a6ac-fec77f9361ca

Almost the same floating issue happens on the UI selection for me (where I verified that I properly selected the wheels), though Ernie is properly assigned and the console errors don't appear.

PR #1459, opened 2026-07-29
URDF Mesh Decimation `[SYNTH-264]`

BrandonPacewic:
## Task

SYNTH-264

Reduce URDF assembly size and enable caching. 

## Symptom

URDF exports contain extremely detailed meshes. Far higher resolution than we would ever need. Certain vendor provided screws that I identified were over 100k triangles when they should have been just a few hundred. 

This caused robots that we really want to support like [2026 Kitbot (download here)](https://github.com/user-attachments/files/30494860/Kitbot.2026.zip) being quite literally Gigabytes post import. 

## Solution

Two part solution:

1. Reuse byte-identical meshes such that we avoid carrying redundant copies of meshes (as much as reasonably possible). 
2. Fully implement mesh decimation for all 3 mesh data formats for URDF (STL/Obj/glTF).

## Verification

Import a URDF kitbot. It will preform much much better now, both in terms of import time and in terms of memory usage.

These changes for me took the imported triangle count of 2026 Kitbot from 1.2 million to 150 thousand (91.6% reduction) and took the memory usage from 3.11GB to 117MB.

Confirm URDF OPFS caching works as expected.

---

- [ ] User-facing polish:

### DIFF
--- a/fission/src/urdf/MeshDecimation.ts
+++ b/fission/src/urdf/MeshDecimation.ts
@@ -1,7 +1,7 @@
 import { MeshoptSimplifier } from "meshoptimizer"
 import type { ParsedMesh } from "./STLParser"
 
-// STL Mesh Decimation for URDF STL mesh imports.
+// Mesh decimation for URDF STL/OBJ/glTF imports.
 // This decimation step takes the 2026 kitbot import size from 3.11GB to 117MB.
 
 // Every result is checked against the input (see validate()) and the original mesh is returned
@@ -44,6 +44,10 @@ export async function readyMeshDecimation(): Promise<void> {
 interface Welded {
     verts: Float32Array
     indices: Uint32Array
+    /** Per-vertex UV, 2 floats per entry in `verts`. Zero-length when the source mesh carries no UV
+     * (STL, or a glTF/OBJ mesh with no texture coordinates) - skipped rather than carried as zeros
+     * so the STL path does no extra work. */
+    uv: Float32Array
 }
 
 // Collapses bit-identical positions into a compacted vertex buffer and reindexes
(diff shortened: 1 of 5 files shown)


fission/src/urdf/MeshDecimation.ts
```diff
@@ -0,0 +1,670 @@
...
+//
+// These are calibrated values from the 2026 FRC kitbot. These values keep the measured surface
+// within 0.5mm of the original while still having a triangle reduction count of about 87.1%.
+//
+// Bottom rungs exist for hardware like screws. Some of these parts have measured over 100k triangles
+// regardless of how small they are so it's worth waking down to 0.03mm to avoid skipping those meshes.
+const ERROR_BUDGETS_M = [0.0005, 0.00025, 0.000125, 0.0000625, 0.00003125, 0.000015625]
+// ...but never spend more error than this fraction of the part's own bounding-box diagonal, so a
+// 5mm spacer isn't handed a budget the size of itself.
+const RELATIVE_BUDGET_CAP = 0.01
```
rutmanz, line 21: There are a lot of words here but it's kind of hard to tell what this means. What is an error budget and why is it an array of progressively smaller values?

fission/src/urdf/MeshDecimation.ts
```diff
@@ -0,0 +1,670 @@
...
+}
+
+interface Welded {
+    verts: Float32Array
+    indices: Uint32Array
+    uv: Float32Array // Zero length with the source mesh caries no UV
+}
+
+// Collapses bit-identical positions into a compacted vertex buffer and reindexes
+function weldPositions(mesh: ParsedMesh): Welded {
```
rutmanz, line 57: Why is this called welding?

rutmanz (approved):
Still ~2x the size of the exported Mirabuf file (might just be tunable for how decimatey you are), but much much better than before. 


(Mirabuf kitbot is top, URDF kitbot is bottom)
image

I'm not really going to be able to understand everything you're doing with the amount of time I want to spend on this PR, but it seems reasonable and seems to work well.

fission/src/urdf/MeshDecimation.ts
BrandonPacewic: Autodesk tools call this function "Weld Vertices" already. Just using the same terminology for functionally the same thing.

fission/src/urdf/MeshDecimation.ts
BrandonPacewic: An error budget refers to the max allowed amount of surface deviation in meters after a decimation pass. Try the biggest reduction budget first (most likely to fail), if that result fails validation (volume/area/axis checks) then try the next smaller budget. Repeat until either we get a reduction that seems reasonable based on the validation checks or just return the whole mesh untouched if we can't confidently say that the decimation did not screw up the visual of the mesh.

BrandonPacewic:
> Still ~2x the size of the exported Mirabuf file (might just be tunable for how decimatey you are), but much much better than before.

All things considered this is a pretty conservative path to mesh decimation. I'm sure we could tune it and be more permissive with the validation step to get better results. Airing on the side of caution for now.

AlexD717 (approved):
lgtm. Tested with my teams robots in .stl and .gltf format

PR #1462, opened 2026-07-30
Mecanum Drive `[SYNTH-297]`

BrandonPacewic:
## Task

SYNTH-297

## Symptom

Much like Swerve drive Mecanum drive is totally unsupported in the current simulator.

## Solution

Implement Jolt based mecanum drive. This, unlike swerve, is a fresh implementation with jolt physics. I ran into numerous issues around mocking wheel drive angles and friction, all of which is extensively documented.

Since this really just mocks what the mecanum drive physics would look like if you had rollers on your wheels you can assign any drivable robot this drivetrain type; does not depend on any particular cad design like swerve.

## Verification

Spawn in any wheeled robot and simply select `Mecanum` as your drive type. The drive behavior itself will be very similar to swerve; also field centric.

Confirm that all the usual behavior with drive schemes is consistent otherwise.

---

- [ ] User-facing polish:


RoushilS (changes requested):
This works very well! One thing that I would like is the ability to have robot centric driving, as this is what most FTC teams are used to. (0.33% of FRC teams use mecanum vs 88% of FTC teams)

Though this would require some user facing UI changes (possibly add a second dropdown of Field vs Robot oriented under Drivetrain Type in the Configure Drivetrain panel.) The internal change is rather simple. 

it involves replacing this block:

```
const { forward, strafe } = motion
            ? this.toChassisFrame(fieldForward, fieldStrafe, motion.nose)
            : { forward: fieldForward, strafe: fieldStrafe }
```

with this

```
const forward = fieldForward
        const strafe = fieldStrafe
```

if the robot centric option is selected.

RoushilS (approved):

RoushilS:
One note - on light robots (below 15kg) the robot will drift a bit when there is no input. This is not really an issue but I was curious why this is happening. Here is the file of the robot I am using in case anyone wants to see. 
[DT2Jointed_v8.zip](https://github.com/user-attachments/files/30564935/DT2Jointed_v8.zip)

AlexD717 (approved):
lgtm. I did notice that when turning the robot overshoots a bit but then fixes itself (after releasing the arrow key it turns for a few more frames but then turns back). Depending on how hard this is to fix it could be fixed here, in a different ticket, or just not.

PR #1463, opened 2026-07-31
Use UUID-based scene object IDs `[SYNTH-309]`

rutmanz:
## Task

SYNTH-309

## Symptom

Currently scene object IDs are auto-incrementing integers, which makes handling translations in multiplayer exceedingly obnoxious. 

## Solution

This system uses UUIDs instead, meaning no translation is required and scene object IDs match across clients.

## Verification

Theoretically, there should be no behavior change. If you find something that's different, that's a bug.

---

- [X] User-facing polish:

### DIFF
--- a/fission/src/ui/panels/configuring/CameraSelectionPanel.tsx
+++ b/fission/src/ui/panels/configuring/CameraSelectionPanel.tsx
@@ -18,6 +18,7 @@ import { useUIContext } from "@/ui/helpers/UIProviderHelpers"
 import CommandRegistry from "@/ui/components/CommandRegistry"
 import { globalOpenPanel } from "@/ui/components/GlobalUIControls"
 import { MenuItem } from "@mui/material"
+import type { SceneObjectId } from "@/systems/scene/SceneRenderer.ts"
 
 CommandRegistry.get().registerCommand({
     id: "open-camera-config",
@@ -27,7 +28,7 @@ CommandRegistry.get().registerCommand({
     perform: () => import("./CameraSelectionPanel").then(m => globalOpenPanel(m.default, undefined)),
 })
 
-const UNFOCUSED_ID = -1
+const UNFOCUSED_ID = "unfocused" as SceneObjectId
 const CENTER_POINT_INDEX = -1
 
 function getSceneObjects(): MirabufSceneObject[] {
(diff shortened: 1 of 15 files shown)


rutmanz:
Note that this is a blocker for multiplayer, but seemed too unrelated to be in the same PR

AlexD717 (approved):
lgtm

fission/src/util/Utility.ts
```diff
@@ -99,3 +99,11 @@ export function copyVec3(vec: Jolt.Vec3): Jolt.Vec3 {
...
  */
 export const yieldToMain = () => new Promise<void>(resolve => setTimeout(resolve, 0))
+
+export type RecursivePartial<T> = {
+    [P in keyof T]?: T[P] extends (infer U)[]
+        ? RecursivePartial<U>[]
+        : T[P] extends object | undefined
+          ? RecursivePartial<T[P]>
+          : T[P]
+}
```
0xda157, line 109: 😭

fission/src/ui/panels/configuring/CameraSelectionPanel.tsx
```diff
@@ -27,7 +28,7 @@ CommandRegistry.get().registerCommand({
     perform: () => import("./CameraSelectionPanel").then(m => globalOpenPanel(m.default, undefined)),
 })
 
-const UNFOCUSED_ID = -1
+const UNFOCUSED_ID = "unfocused" as SceneObjectId
```
0xda157: seems like a questionable way of handling this to me. I think we should just change the type to be nullable or smth.

fission/src/util/Utility.ts
rutmanz: isn't typescript a beautiful language

fission/src/ui/panels/configuring/CameraSelectionPanel.tsx
```diff
@@ -178,14 +177,14 @@ const FieldViewSettings: React.FC = () => {
                         value={focusedRobotId}
                         onChange={e => {
                             const id = e.target.value as SceneObjectId
-                            const robot = id === UNFOCUSED_ID ? undefined : robots.find(r => r.id === id)
+                            const robot = id == null ? undefined : robots.find(r => r.id === id)
```
0xda157, line 180: why'd we lose an `=` here? using `===` is probably a good idea.

rutmanz: because I needed to set it to undefined and not null to make the typing on menuitem work. I could check ===undefined but that feels less clear

0xda157 (approved):
diff LGTM

PR #1468, opened 2026-07-31
FTC Codesim `[SYNTH-295]`

BrandonPacewic:
## Task

SYNTH-295

Support FTC codesim in the Fission simulator.

## Symptom

I don't think we have ever actually supported any variation of FTC codesim. Now we do.

## Solution

Follows the same type of structure as our FRC codesim although its much simplier. FRC codesim requires and relies on the whole WPILib stack built against the actual manufacturer APIs. The FTC SDK is direct so we can simply setup a clean-room shimmed version of the handful of base classes we want and compile a team's raw `.java` OpMode source file directly with JDK.

This pr sets up just a handful of FTC classes needed to run a very basic TeleOp Arcade drive mode for dozer. It would be very easy for us to add more support later, this is just the initial scaffolding.

This PR does not setup any autonomous mode testing in FTC.

## Verification

You may want to use [this version of Dozer](https://github.com/user-attachments/files/30606786/Dozer.v69.zip) as it has labeled wheel joints.

- Run `cd simulation/SyntheSimFTC && ./gradlew build`.
- Run `./gradlew run --args="--src ../samples/FTCDozerArcadeDriveSample --opmode ExampleDozerArcadeDrive"`.
- Open up the fission dev server while that is running.
- Spawn Dozer in Fission, configure Brain -> FTC.
- Configure Assets -> Simulation -> Wiring Panel.
  - Make the wiring panel look something like this:
image

- You then should be able to drive with `WASD` on simulated code.

---

- [ ] User-facing polish:


PepperLola (commented):
Is there a reason the FTC and FRC code simulation needs to be distinct on the Fission side? especially if they're going to be the same after systemcore. A lot of the system looks like it's a duplicate of the WPILib sim when it doesn't need to be (e.g., the web worker for the websocket looks to be pretty much exactly the same except for the WS endpoint?).

PepperLola (commented):
Also, the roborio config modals are unused and I'm deleting them in my next PR, so there's no need to adapt that for FTC. I think ideally everything should work the same on the Fission side between FTC and FRC (since I don't know what would necessarily need to be different; the configuration is pretty flexible already?).

BrandonPacewic:
> Is there a reason the FTC and FRC code simulation needs to be distinct on the Fission side? especially if they're going to be the same after systemcore. A lot of the system looks like it's a duplicate of the WPILib sim when it doesn't need to be (e.g., the web worker for the websocket looks to be pretty much exactly the same except for the WS endpoint?).

I combined both the web workers, those were mostly duplicated.

I believe everything else here is split correctly. WPI has a whole flow graph of SimInput/SimOutput/SimFlow/Nora-type etc. FTC has none of this and is simply motor power out, gamepad in. No bidirectional flow whatsoever.

Once we know what systemcore is going to look like we can likely move both FTC and FRC codesim to be the same process.

RoushilS (changes requested):
~~I'm having an issue where if on the FTC brain after clearing data, and I go to the joints panel,  the simulation crashes. In the video below I configure wheels, but that isn't even necessary.~~

Nevermind, looks like this happens in other places as well. Making an issue.

https://myshare.autodesk.com/:v:/r/personal/roushil_satta_autodesk_com/Documents/Videos/Clipchamp/Video%20Project%201/Exports/joints_bug.mp4?d=w5bb8da5da2a046cbbc8136c3269f42a1&csf=1&web=1&e=ew69tz

RoushilS (approved):
Very glad to see FTC support in Synthesis!

PepperLola (commented):
As far as I can tell, the only difference between the FTC brain and the WPILib one is that FTC handles gamepad input. When it comes to code sim, I think that should ideally be handled on the robot code side (the split of responsibilities should be the same between FRC and FTC code sim). At that point, we could keep all of the configuration we provide for FRC and use it for FTC, meaning users wouldn't have to learn how to do things separately. It would also be easier to maintain since we wouldn't have to maintain two systems that functionally do the same thing.

PepperLola:
Are the Qualcomm files in SyntheSimFTC derived from the real library? If so, they need to retain the copyright notice

Also, FTCState and WPILibState are exactly the same except the WPILib one uses a "SimAutoReconnect" user preference and exports some constants. I would rather we reuse whatever we can and just change WPILibState to always automatically reconnect

FTCTypes are identical as well except for the added DcMotor, which I think should just be either a PWM or a regular SimDevice so we can reuse WPILib sim stuff.

Since we have complete control over the websocket spec for FTC code sim, it would be nice to have it match the WPILib one (e.g., use <> to denote inputs/outputs, same names, etc.). Ideally we would just treat the DcMotors as PWM devices on the Java side and then nothing in Fission would have to change.

Could we also have a human write the documentation?

Since we don't support configuring the websocket url/port, I don't think it makes sense to let users pass in a port to the op mode runner.

BrandonPacewic:
> Are the Qualcomm files in SyntheSimFTC derived from the real library? If so, they need to retain the copyright notice

These shims were written from a decompiled signature list, the actual source content of these files is all closed source. Since we are effectively reimplementing an API for interop this falls under fair use.

simulation/samples/FTCDozerArcadeDriveSample/org/firstinspires/ftc/teamcode/examples/ExampleDozerAutoDrive.java
```diff
@@ -0,0 +1,65 @@
...
+
+                case DRIVE_BACKWARD:
+                    setAllPower(leftFront, leftMiddle, leftBack, rightFront, rightMiddle, rightBack, -0.5);
+                    break;
+
+                case STOPPED:
+                    setAllPower(leftFront, leftMiddle, leftBack, rightFront, rightMiddle, rightBack, 0.0);
+                    break;
+            }
+        }
```
azaleacolburn: We should just set a variable that we use at the last arg and move the rest of the `setAllPower` function call to be unconditional. We should also use an switch expression (rather than a statement) to set that variable (assuming we're on Java version 14 or higher).

simulation/samples/FTCDozerArcadeDriveSample/org/firstinspires/ftc/teamcode/examples/ExampleDozerAutoDrive.java
```diff
@@ -0,0 +1,65 @@
...
+    // Helper method to reduce code repetition when setting motor power
+    private void setAllPower(DcMotorSimple lf, DcMotorSimple lm, DcMotorSimple lb, 
+                             DcMotorSimple rf, DcMotorSimple rm, DcMotorSimple rb, double power) {
+        lf.setPower(power);
+        lm.setPower(power);
+        lb.setPower(power);
+        rf.setPower(power);
+        rm.setPower(power);
+        rb.setPower(power);
+    }
```
azaleacolburn, line 58: This could take a list of `DcMotorSimple` and then loop over the list, since the same method is called each time.

azaleacolburn (approved):
Everything look great; I have a few nits about the java code, but they're not blocking imo.

fission/src/systems/simulation/wpilib_brain/WPILibState.ts
```diff
@@ -69,6 +71,7 @@ export const receiverTypeMap: { [k in SimType]: NoraTypes | undefined } = {
     [SimType.AI]: NoraTypes.NUMBER,
     [SimType.AO]: undefined,
     [SimType.DRIVERS_STATION]: undefined,
+    [SimType.GAMEPAD]: undefined,
     [SimType.CAMERA]: undefined,
 }
 */
```
PepperLola: Delete this, as the new Nora updates render it obsolete.

fission/src/ui/components/TopBar.tsx
```diff
@@ -183,7 +183,10 @@ const TopBar: React.FC = () => {
 
                     {hasSimBrain() && (
                         <>
-                            <CodeConnectionIndicator />
+                            <CodeConnectionIndicator
+                                label={getSimBrain()?.brainType === "ftc" ? "FTC code connection" : "Code connection"}
```
PepperLola: Do we need to differentiate here? Could they not both be `Code connection`?

fission/src/systems/simulation/wpilib_brain/WPILibState.ts
PepperLola: Delete this, as the new Nora updates render it obsolete.

PepperLola (dismissed):

RoushilS:
~~@PepperLola Would it possible for you to add those reccomendations? Currently my git is in a weird state on another branch.~~

fission/src/ui/components/TopBar.tsx
RoushilS: I will make the change

PR #1471, opened 2026-08-03
Use time-based timer for match mode and improve styling `[SYNTH-315]`

rutmanz:
## Task

SYNTH-315

Predicate for multiplayer

## Symptom

Previously, the match mode timer was decremented manually by 1 second every 1000ish ms (setInterval does not guarantee that it runs exactly at the time interval, just that it doesn't run any more frequently). This meant that matches would be randomly longer than they should be based on how long the event loops took. This is the same issue that occurred with the FMS on Newton at the 2026 World Championship, so I guess by fixing this we are decreasing realism.

## Solution

I used a start time variable to keep track of elapsed times. It still uses setTimeout to trigger behavior, but this is unavoidable in JS. What this change accomplishes is preventing those delays from compounding, so it might be 30ms behind, but it won't be an additional 30ms behind every second.

Also the styling was kind of wonky on match results with the robot breakdown, and it threw a react error, both of which are now fixed.

## Verification

Timer works as expected and matches external timing source, even when slowed down browser via dev tools or 2026 field (he he) 

---

- [X] User-facing polish:

### DIFF
--- a/fission/src/systems/match_mode/MatchMode.ts
+++ b/fission/src/systems/match_mode/MatchMode.ts
@@ -130,7 +130,7 @@ class MatchMode {
     }
 
     async start(broadcast = true, useSpawnPositions: boolean) {
-        const startTime = Date.now() + 500
+        const startTime = Date.now() + 300 // Accounts for time it takes for robots to move to start positions and settle, and for multiplayer state to sync
         if (broadcast && World.multiplayerSystem) {
             await World.multiplayerSystem.broadcast({
                 type: "matchModeState",
@@ -147,12 +147,14 @@ class MatchMode {
         if (useSpawnPositions) {
             World.getOwnRobots().forEach(obj => obj.moveToSpawnLocation())
         }
-        this.autonomousModeStart()
+
         World.scoreTracker.resetScores()
         RobotDimensionTracker.matchStart()
 
         const matchEvent = createMatchEventFromConfig(this._matchModeConfig)
         World.analyticsSystem?.event("Match Start", matchEvent)
+
+        this.runForNext(0, false).then(() => this.autonomousModeStart())
     }
 
     matchEnded() {
--- a/fission/src/systems/match_mode/MatchMode.ts
+++ b/fission/src/systems/match_mode/MatchMode.ts
@@ -46,6 +46,7 @@ class MatchMode {
     private _startTime: number = 0
     private _timeUsed: number = 0
     private _intervalId: number | null = null
+    private _cancelHandler: (() => void) | null = null
 
     // Match Mode Config
     private _matchModeConfig: MatchModeConfig = DefaultMatchModeConfigs.fallbackValues()
@@ -72,6 +73,11 @@ class MatchMode {
         return this._matchModeConfig
     }
 
+    /**
+     * Waits for the specified duration in match time to pass, accounting for delays from prior loop cycles. Resolves when the time has elapsed, rejects if cancelled.
+     * @param duration The duration, in seconds, of the phase
+     * @param updateTimeLeft Whether to dispatch scoreboard update events
+     */
     async runForNext(duration: number, updateTimeLeft: boolean = true) {
         if (this._intervalId !== null) {
             console.warn("Timer already running")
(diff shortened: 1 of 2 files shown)


fission/src/systems/match_mode/MatchMode.ts
```diff
@@ -129,13 +130,20 @@ class MatchMode {
     }
 
     async start(broadcast = true, useSpawnPositions: boolean) {
-        if (broadcast) {
-            await World.multiplayerSystem?.broadcast({
+        const startTime = Date.now() + 500
```
AlexD717: Can you add a comment explaining what the `+ 500` is for? If its just to account for multiplayer latency I think it would make more sense to pass in the start time in the `MatchModeStateData` broadcast.

AlexD717 (commented):
If we are relying on `Date.now()` doesn't that introduce a possible unsync scenario if a devices clock is unsynchronized? Especially for mobile devices I don't believe its uncommon for the time to drift by like a second.

rutmanz:
> If we are relying on `Date.now()` doesn't that introduce a possible unsync scenario if a devices clock is unsynchronized? Especially for mobile devices I don't believe its uncommon for the time to drift by like a second.

Yes, it does introduce that for multiplayer, so when the multiplayer side of this gets implemented it will need to account for clock offset. This is a separate PR from Multiplayer just bc it's mostly unrelated to multiplayer functionality and can be reviewed separately

fission/src/systems/match_mode/MatchMode.ts
rutmanz: This is for making sure everything else has time to happen before match mode starts. I've added a comment explaining, but this can't be accomplished by doing it in the start time passed to other devices bc if the message arrives 100ms late, the auto code will start 100ms late even if the timer is accurate, which isn't really desirable

AlexD717 (approved):
Match mode time timer was ticking a tad too tardy so this should make the match mode time timer more timely and time based by tying the match time timer to true real time to track the match time timely.

azaleacolburn (approved):

Dhruv-0-Arora:
this isn't super relevant to this PR, but match mode does not have a cancel button. To cancel match mode, you need to use the command palatte and use the toggle button there.
Also, when I cancel a match mode while in between auto and teleop, the teleop continues (match mode doesn't cancel).

fission/src/systems/match_mode/MatchMode.ts
```diff
@@ -72,53 +72,54 @@ class MatchMode {
...
+                this.endgameStart()
+            }
+        }, 200)
+
+        const remainingTime = this._startTime + this._timeUsed - Date.now() + duration * 1000
+        return new Promise<void>(res => setTimeout(res, remainingTime)).finally(() => {
+            this._timeUsed += duration * 1000
             clearInterval(this._intervalId as number)
+            this._intervalId = null
         })
```
Dhruv-0-Arora: Canceling a match mode doesn't actually stop the match from finishing.

You can cancel it with the command palette

Dhruv-0-Arora (changes requested):

fission/src/systems/match_mode/MatchMode.ts
```diff
@@ -129,27 +130,35 @@ class MatchMode {
...
+                data: {
+                    event: "start",
+                    config: this._matchModeConfig,
+                    moveRobots: useSpawnPositions,
+                },
             })
-            console.log("sent multiplayer")
         }
+
+        this._startTime = startTime
```
Dhruv-0-Arora, line 162: Do you think it would be good practice to reset `this._timeUsed` here?

fission/src/systems/match_mode/MatchMode.ts
```diff
@@ -72,53 +72,54 @@ class MatchMode {
...
+
+            if (timeLeft >= 0 && updateTimeLeft) {
+                EventSystem.dispatch("TimeChangedEvent", { time: timeLeft })
+            }
+
+            // Checks if endgame has started
+            if (
+                this._matchModeType === MatchModeType.TELEOP &&
+                timeLeft <= this._matchModeConfig.endgameTime &&
+                !this._endgame
```
Dhruv-0-Arora, line 100: this._endgame isn't reset to false anywhere -> bug: when I start a 2nd match without refreshing, then endgame doesn't happen again

Dhruv-0-Arora (changes requested):

AlexD717:
> this isn't super relevant to this PR, but match mode does not have a cancel button. 

#1473 will introduce a match mode cancel button

Dhruv-0-Arora (approved):

BrandonPacewic (approved):

PR #1483, opened 2026-08-06
Properly Scaled Unstick Force `[SYNTH-323]`

AlexD717:
## Task

SYNTH-323

## Symptom

Unstick force is intended as a slight bump to help get robots unstuck (when they are slightly glitching through the ground), however, it used AddForce which scales with FPS and results in unreasonably high forces at low FPS.

## Solution

Switch to using AddImpulse which isn't depended on FPS, and change defaults to reasonable values (run `bun run assetpack` to update).

## Verification

- [x] Can still apply unstick force
- [x] Applied force is the same no matter the FPS
- [x] Applied force is large enough to cause a slight jump making robots unstuck, but not large enough to have them go flying

---

- [x] User-facing polish:


elliefre (changes requested):
I can't dismiss this, however there is no issue, there was issues on my end not with the change.

elliefre (approved):
This is good, ignore previous request for changes.

Dhruv-0-Arora (changes requested):
Image

existing preferences in local storage will cause the robot to fly up on unstick, forcing users to have to change the unstick. Maybe add a migration thing?

Dhruv-0-Arora (approved):
🔥

rutmanz (commented):
Maybe change the label of "Unstick Force" in the UI (Configure > Joints > Drivetrain)? It's not really a force anymore

AlexD717:
Renamed to "Unstick Strength", which also made the migration logic obsolete (since it just defaults to the default value if there is no unstickStrength key)

rutmanz (approved):
lgtm

PR #1484, opened 2026-08-06
Resync gizmo body on parent object move `[SYNTH-334]`

BrandonPacewic:
## Task

SYNTH-334

## Symptom

`GizmoSceneObject` caches ceach rigid body's transform relative to the gizmo once at construction. If the parent object moves afterward through something other than the gizmo that cache goes stale and causes the gizmo to be out of sync with its assigned component.

## Solution

Allow for the ability to sync the gizmo to its parent for usage when you want to be able to move the parent component and the attached gizmo at the same time.

## Verification

Gizmo behaves normally. We currently do not have any callers that need this functionality. But we will for MAM. Since that pr stack is already going to be quite large I decided to move this into its own chunk.

---

- [ ] User-facing polish:


fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -68,22 +68,47 @@ class GizmoSceneObject extends SceneObject {
...
+            )
+            const relativeTransform = worldTransform.premultiply(gizmoTransformInv)
+            this._relativeTransformations!.set(rn.id, relativeTransform)
+        })
+    }
+
+    /**
+     * Re-seats the gizmo on its parent after something other than this gizmo moved the parent.
+     */
+    public syncToParent() {
```
azaleacolburn, line 103: Might want to mark this as unused so people know not to remove it.

fission/src/systems/scene/GizmoSceneObject.ts
```diff
@@ -68,22 +68,47 @@ class GizmoSceneObject extends SceneObject {
...
-            })
-        }
+        if (this._parentObject) this.bakeRelativeTransformations()
+    }
+
+    /**
+     * Records where every body of the parent sits relative to the gizmo, which is what dragging then
+     * moves them by.
+     *
+     * Re-baked rather than updated in place: the offsets only stay valid for as long as the parent moves
```
azaleacolburn, line 78: Does the term "bake" have a technical meaning? Would "calculate" not be clearer?

azaleacolburn (approved):

AlexD717 (approved):
lgtm

fission/src/systems/scene/GizmoSceneObject.ts
PepperLola: "Baking" is a pretty common term in the 3d animation space to refer to precomputing something

PR #1487, opened 2026-08-07
feat(git): biome pre-commit hook

Dhruv-0-Arora:
## Task

A git hook to notify you when you try to commit something that you have a formatting error in the files that you are trying to commit. To install just `bun i`

- will append a `sh /fission/scripts/hooks/pre-commit` to your /.git/hooks/pre-commit

To **bypass**, run `git commit --no-verify` when committing

> [!NOTE]
> Works on git worktrees only if that worktree contains the /fission/scripts/hooks/pre-commit file

> [!WARNING]
> Install script doesn't work on windows when using powershell and cmd. Fails cleanly

SYNTH-336

## Symptom

There a quite a few `chore(formatting): ...` commits right now. This PR aims to reduce the number of those.

## Verification

Create a formatting mistake in a file and try to commit the formatting mistake.

---

- [ ] User-facing polish:


fission/package.json
```diff
@@ -27,6 +27,7 @@
     "assetpack:update": "bun update_manifest.ts && cd public && zip -FS -r assetpack.zip Downloadables -x '**/.*' -x '**/__MACOSX'",
     "assetpack:merge": "git checkout --theirs public/assetpack.zip && rm -rf public/Downloadables && tar -xf public/assetpack.zip -C public/ && git checkout --ours public/assetpack.zip && tar -xf public/assetpack.zip -C public/ && bun run assetpack:update",
     "playwright:install": "bun x playwright install",
+    "hooks:install": "bun scripts/hooks/install.ts",
```
0xda157:
```suggestion
    "hooks:install": "install -Dm755 scripts/hooks/pre-commit $(git rev-parse --git-common-dir)/hooks",
```

Dhruv-0-Arora: This wouldn't work on windows @0xda157

Dhruv-0-Arora:
```suggestion
    "hooks:install": "cp scripts/hooks/pre-commit \"$(git rev-parse --git-common-dir)/hooks/pre-commit\" && chmod 755 \"$(git rev-parse --git-common-dir)/hooks/pre-commit\"",
```
This could work

0xda157: oh right forget we support windows.

fission/package.json
```diff
@@ -27,6 +27,7 @@
     "assetpack:update": "bun update_manifest.ts && cd public && zip -FS -r assetpack.zip Downloadables -x '**/.*' -x '**/__MACOSX'",
     "assetpack:merge": "git checkout --theirs public/assetpack.zip && rm -rf public/Downloadables && tar -xf public/assetpack.zip -C public/ && git checkout --ours public/assetpack.zip && tar -xf public/assetpack.zip -C public/ && bun run assetpack:update",
     "playwright:install": "bun x playwright install",
+    "hooks:install": "git config core.hooksPath fission/scripts/hooks",
```
Dhruv-0-Arora: This implementation supports windows but imo is a worse way to do this over
```
"hooks:install": "cp scripts/hooks/pre-commit \"$(git rev-parse --git-common-dir)/hooks/pre-commit\" && chmod 755 \"$(git rev-parse --g
it-common-dir)/hooks/pre-commit\"",
```
this version doesn't work on windows

rutmanz: I don't think it's appropriate to change the hooks directory, I have my own checkout hooks that I use that I don't think should be committed to version control and don't want globally. Better to just append "./scripts/hooks/pre-commit" with appropriate dir control to the pre-commit file.

Alternatively there are like dozens of NPM libraries for doing exactly this

fission/README.md
```diff
@@ -132,3 +132,4 @@ These core systems make up the lion's share of the fission source code. Each sys
 | `assetpack`          | Downloads the assetpack and unzips/installs it in the correct location.                                                                         |
 | `assetpack:update`   | Downloads the assetpack and unzips/installs it in the correct location, replacing the old directory if it exists.                               |
 | `playwright:install` | Downloads the Playwright browsers.                                                                                                              |
+| `hooks:install`      | Points `core.hooksPath` at `fission/scripts/hooks`, enabling the Biome pre-commit hook. Run automatically by `init`.                            |
```
rutmanz: this is a little thing I like to call factually inaccurate.

https://github.com/Autodesk/synthesis/blob/5aafdb8d861d97aa8bcd00d781dff34785e1bb53/fission/package.json#L10

rutmanz (changes requested):

fission/README.md
Dhruv-0-Arora: I removed it 🤦‍♂️

fission/package.json
```diff
@@ -27,6 +27,7 @@
     "assetpack:update": "bun update_manifest.ts && cd public && zip -FS -r assetpack.zip Downloadables -x '**/.*' -x '**/__MACOSX'",
     "assetpack:merge": "git checkout --theirs public/assetpack.zip && rm -rf public/Downloadables && tar -xf public/assetpack.zip -C public/ && git checkout --ours public/assetpack.zip && tar -xf public/assetpack.zip -C public/ && bun run assetpack:update",
     "playwright:install": "bun x playwright install",
+    "prepare": "f=$(git rev-parse --git-path hooks/pre-commit 2>/dev/null) || exit 0; [ -e \"$f\" ] || printf '#!/bin/sh\\n' > \"$f\"; grep -q fission/scripts/hooks/pre-commit \"$f\" || printf 'sh ./fission/scripts/hooks/pre-commit\\n' >> \"$f\"; chmod +x \"$f\"",
```
0xda157, line 30: ??? this is a terrible one liner. make this it's own script or simplify it.

Dhruv-0-Arora: I would personally disagree - I feel that a script is unnecessary and all this line does is append
`sh ./fission/scripts/hooks/pre-commit` to .git/hooks/pre-commit if it doesn't already exist.

rutmanz: If you're using bash if statements it should probably be an actual file imo. Having everything on one line with a bunch of string escaping and no whitespace is not exactly a nice experience

Dhruv-0-Arora: 👍 im going to change it. Julian also agreed with you guys

fission/scripts/hooks/install.sh
```diff
@@ -0,0 +1,40 @@
...
+set -e
+
+# git runs hooks from the repo root so the path is relative to the root
+HOOK_COMMAND='sh ./fission/scripts/hooks/pre-commit'
+
+if ! git rev-parse --git-dir >/dev/null 2>&1; then
+  exit 0
+fi
+
+hooks_dir=$(git config --get core.hooksPath 2>/dev/null || true)
```
0xda157, line 14: hooks_dir=$(true) ?

fission/scripts/hooks/install.sh
```diff
@@ -0,0 +1,40 @@
...
+  hook_file="$hooks_dir/pre-commit"
+else
+  # worktree-safe
+  hook_file=$(git rev-parse --git-path hooks/pre-commit 2>/dev/null) || exit 0
+fi
+
+mkdir -p "$(dirname "$hook_file")"
+
+if [ ! -e "$hook_file" ]; then
+  printf '#!/bin/sh\n' >"$hook_file"
```
0xda157, line 29: we should always use `/bin/sh` or always use `/usr/bin/env sh`

0xda157, line 29:
```suggestion
  echo "#!/bin/sh" >"$hook_file"
```

fission/scripts/hooks/install.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+chmod +x "$hook_file"
+
+# hook already installed
+if grep -q 'fission/scripts/hooks/pre-commit' "$hook_file"; then
+  exit 0
+fi
+
+printf '%s\n' "$HOOK_COMMAND" >>"$hook_file"
```
0xda157, line 38:
```suggestion
echo "$HOOK_COMMAND" >>"$hook_file"
```

0xda157 (changes requested):
this breaks if you checkout a branch that doesn't include this PR. we should copy the script to pre-commit, not insert a script that calls it.

fission/scripts/hooks/pre-commit
```diff
@@ -0,0 +1,33 @@
...
+
+staged_list=$(mktemp)
+trap 'rm -f "$staged_list"' EXIT
+git diff --cached --name-only --diff-filter=ACMR --relative -z -- . >"$staged_list"
+
+if [ ! -s "$staged_list" ]; then
+  exit 0
+fi
+
+if ! xargs -0 bun x biome check --no-errors-on-unmatched --files-ignore-unknown=true <"$staged_list"; then
```
PepperLola, line 27: This will read the staged files from disk? In which case it's possible that someone could stage something that's formatted correctly, make a change to the file with bad formatting (but not stage) and then they'd be blocked from committing even though the staged change is okay.

This could also happen the other way around - someone stages changes with bad formatting and then runs the formatter but doesn't stage the correctly formatted files. Then they'd be allowed to commit even though the format is bad.

fission/scripts/hooks/install.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env sh
+#
+# Installs the fission Biome pre-commit hook.
+
+set -e
+
+# git runs hooks from the repo root so the path is relative to the root
+HOOK_COMMAND='sh ./fission/scripts/hooks/pre-commit'
```
PepperLola, line 8: Farther down you have a `hook_file` case for worktrees and explicitly mention that it's worktree-safe, but  wouldn't this assume that the root directory of the repository contains `fission` which has `scripts/hooks`?

PepperLola (changes requested):
I left a couple comments. If the pre-commit hook is installed automatically on `bun i` would we want a way to disable the installation for CI? Could just be an env var we specify in the workflow file or something.

fission/scripts/hooks/install.sh
Dhruv-0-Arora: yes, but I made the assumption that the root folder will contain the scripts/hooks. the pre-commit file will be in the root folder anyways so I think it is the best solution. Trying to navigate into the folder of the worktree would be complicated.

PepperLola: It looks like the command is run from the root folder of the worktree (so `./fission` will exist). The comment made it seem like it would run from the directory that contains the worktrees, in which case `./fission` would never exist (unless someone made a worktree called `fission`. I tested the commit hook and it properly blocked a commit with bad formatting, so it appears this is not an issue. The other comment still stands though.

Dhruv-0-Arora:
This PR is going to require a bit more work that will be done on a later date:
it requires solidifying the logic of how we handle reading the formatting of the staged file rather than the file from disk
biome doesn't have a great mechanism for this yet

PR #1488, opened 2026-08-07
Improve UI Blocking and Exclusion Rules `[SYNTH-338]`

rutmanz:
## Task

SYNTH-338

## Symptom

While the "blocking" attribute prevents other panels from being opened, there is still a lot that can happen in the UI that isn't necessarily desirable (you can join multiplayer while in the initial assembly config panel, for example). I noticed this when making the UI for joint selection, as that needs to be a completely blocking operation

There was also an exclusion rule to prevent config and mira import panels from being open at the same time, but this was based on the file name and extremely scuffed.

## Solution

UI Provider now has a `blockState` attribute, which is used by
- Top bar buttons / drop downs
- Context menu
- Command palette
- MobileHUD drawer
- Panel / modal opening (although theoretically this shouldn't matter because the UI can't be used but it's a good fallback)

## Verification

Spawn an assembly and don't close the initial config panel (this is the only panel with the blocking attribute currently), try to do something meaningful and hopefully you can't.

---

- [X] User-facing polish:


AlexD717 (approved):
lotm

fission/src/ui/UIProvider.tsx
Dhruv-0-Arora:
```suggestion
        [panels, addToast, blockedState]
```

Dhruv-0-Arora: So, if you add this to your dependency array it seems, then if you create a new inputScheme, the blockedState will infinitely regenerate.

fission/src/ui/UIProvider.tsx
```diff
@@ -105,6 +107,20 @@ export const UIProvider: React.FC<UIProviderProps> = ({ children }) => {
         return false
     }
 
+    const blockState: UIBlockState = useMemo(() => {
+        const blockingPanel = panels.find(p => p.props.blocking)
```
Dhruv-0-Arora, line 110:
```suggestion
        const blockingPanel = panels.find(p => p.props.blocking)
        console.count("here")
```
I tested with this here.

Dhruv-0-Arora: In addition, opening a panel fires this twice; while
opening modal fires once
closing panel fires once
closing modal fires once

This is probably irrelevant though / strict mode maybe

Dhruv-0-Arora (changes requested):
this is 🔥 change
This is from our standup discussion:

rutmanz:
re-requested @AlexD717 because I did some more sketchy stuff and I want you to check it again

rutmanz:
oops sorry for force push I didn't realize the github rebase button would do that

fission/src/ui/UIProvider.tsx
```diff
@@ -227,25 +247,32 @@ export const UIProvider: React.FC<UIProviderProps> = ({ children }) => {
...
+                        resolve()
+                        return
+                    }
+                    setConfigureWaiters(v => [...v, cb])
+                }
+                setConfigureWaiters(v => [...v, cb])
+            })
+            const exclusiveGroup = panel.props.exclusiveGroup
+            if (exclusiveGroup != null) {
+                const existing = panels.find(p => p.props.exclusiveGroup == exclusiveGroup)
```
AlexD717: Not really a fan of this. Why can't we just move this check down to the configureScreen const and check there, rather than doing the waiters thing.

AlexD717 (changes requested):
Clicking on a disable dropdown plays a sound effect (which feels wierd)

fission/src/ui/UIProvider.tsx
rutmanz: uh because I wasn't being very smart yeah that's a great point

Dhruv-0-Arora (changes requested):
If you go to 

Configure Assets
Inputs
'+' -> new input scheme

Then you run into an infinite render loop.

fission/src/ui/UIProvider.tsx
```diff
@@ -105,6 +107,20 @@ export const UIProvider: React.FC<UIProviderProps> = ({ children }) => {
...
+            return {
+                blocked: true,
+                blockMessage:
+                    blockingPanel?.props?.blockingMessage ?? "Close the current panel before opening another.",
+            }
+        }
+        return {
+            blocked: false,
+        }
+    }, [refreshDep, panels])
```
Dhruv-0-Arora, line 121:
```suggestion
    const blockingPanel = panels.find(p => p.props.blocking)
    const blocked = blockingPanel != null
    const blockMessage = blockingPanel?.props.blockingMessage ?? "Close the current panel before opening another."
    const blockState: UIBlockState = useMemo(
        () => (blocked ? { blocked: true, blockMessage } : { blocked: false }),
        [blocked, blockMessage]
    )

```

Dhruv-0-Arora: Making the blockState change when the block changes worked for me

rutmanz:
@Dhruv-0-Arora I'm not able to replicate, that works fine for me

Dhruv-0-Arora (approved):
ok

AlexD717 (approved):
lgtm, though build test is failing on CI

PR #1490, opened 2026-08-07
refactor(physics): createAnchorPoint() returns Vec3 `SYNTH-341`

Dhruv-0-Arora:
## Task

This is a refactor made in #1418 which was closed.

SYNTH-341

## Symptom

Currently in dev, createAnchorPoint() returns a RVec3 which is a __static alias__. There is inherently nothing wrong with this, but it could allow for problematic behaviors down the line such as -> `Jolt.destroy(anchorPoint)`

> `anchorPoint` being a static alias means that if the function Jolt.AddRVec3() is run, the anchorPoint value is overwritten and the `anchorPoint` reference points to the new value.

## Solution

Rather than creating 2 `new Jolt.RVec3` for them only to be destroyed in the function scope:
- create 1 `new Jolt.RVec3` -> gets **destroyed** in function scope.
- create 1 `new Jolt.Vec3` -> passing ownership to function caller.

> Jolt.RVec3 is actually fundamentally the same cpp data type as Jolt.Vec3 (both are the same byte size). Under the hood, JS wraps them differently though. Below is why:
<details>

In CMakeLists.txt in JoltPhysics.js, double precision is turned off:
`option(DOUBLE_PRECISION "Compile the library in double precision mode" OFF)`

In Jolt/Math/Real.h:
```cpp
#ifdef JPH_DOUBLE_PRECISION

...

#else

// Define real to float
using Real = float;
using Real3 = Float3;
using RVec3 = Vec3;
```

Because DOUBLE_PRECISION is off, `RVec3 = Vec3`

</details>

---

- [ ] User-facing polish:

### DIFF
--- a/fission/src/systems/physics/ConstraintSettingsUtilities.ts
+++ b/fission/src/systems/physics/ConstraintSettingsUtilities.ts
@@ -6,19 +6,17 @@ import { convertMirabufVector3ToJoltRVec3, convertMirabufVector3ToJoltVec3 } fro
 
 type LimitSpecs = Omit<DOFSpecs, "friction" | "axis">
 
-// Returns a STATIC_ALIAS `RVec3.AddRVec3()`'s scratch buffer.
 export function createAnchorPoint(jointInstance: mirabuf.joint.JointInstance, jointDefinition: mirabuf.joint.Joint) {
-    const jointOrigin = jointDefinition.origin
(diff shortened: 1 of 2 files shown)


azaleacolburn (approved):

AlexD717 (approved):
lgtm

PR #1491, opened 2026-08-07
Ftc codesim autonomous support

RoushilS:
## Task

SYNTH-327

## Symptom

Users had no way to test autonomous simulation in FTC.

## Solution

Created autonomous state, which folds initialization and enabled into one state, and added simulation panel ui, so that autonomous and teleop functionality could be separated. Added opMode slots, opModeCandidate, and opMode discovery, that enables finding and switching between opmodes with robust failure cases. There is also now more detailed command line information. Finally, there is a sample autonomous program added.

## Verification

To verify, use the [Jointed version of Dozer](https://github.com/user-attachments/files/30846407/Dozer.v69.zip), and follow the instructions in CODE_Simulation.md

---

- [ ] User-facing polish:


RoushilS:
Addressed @PepperLola 's IRL Feedback.

PepperLola (changes requested):
Please upgrade gradle to match the version defined [here](https://github.com/FIRST-Tech-Challenge/FtcRobotController/blob/26cd1fdd2a3c4b26173d9ff33a3279c27d1c7ad1/gradle/wrapper/gradle-wrapper.properties#L3) so that newer versions of Java will work. Beyond that I think everything looks good; I'm hoping we can eventually come up with a better (unified) way of enabling auto vs teleop

PepperLola:
Actually, it might be better to update gradle to the same version as in SyntheSimJava to keep things consistent; I don't think it should cause any issues for this library's gradle version to be newer than the FtcRobotController one?

PepperLola (approved):

PepperLola (changes requested):
I assume the FTC java library README is from a previous commit in the parent ftc codesim branch - if so, please consolidate it to just the instructions/command to run to start the simulator. The "Enable FTC Teleop" button should also only display when the FTC brain is enabled and not also for WPILib.

PepperLola (approved):

PR #1511, opened 2026-08-17
Performance Benchmarking `[SYNTH-58]`

azaleacolburn:
## Task

SYNTH-58

## Symptom

No real way to verify performance improvements or check for regressions.

## Solution

Use `vitest bench` to run benchmarks for the:

- MirabufParser
- PhysicsSystem
- SceneRendere

I would be happy to add benchmarks for other systems if need be.

In order to compare performance across commits, simply:

```
git checkout [COMMIT_ONE]
bun run bench:baseline
git checkout [COMMIT_TWO]
bun run bench:check
```

## Verification

`bun run bench` runs and outputs reasonable times
`bun run bench:baseline` writes chromium performance to `src/bench/baseline.json`
`bun run bench:check` notes nearly identical performance to the cached baseline without any changes

---

- [x] User-facing polish:


fission/src/bench/baseline.json
0xda157, line 1: benchmarks are incredibly device specific, why are we including this in the repo?

fission/src/bench/baseline.json
```diff
@@ -0,0 +1,339 @@
+{
+  "files": [
+    {
+      "filepath": "/Users/colbura/synthesis/fission/src/bench/MirabufParser.bench.ts",
```
0xda157: `/Users/colbura` hmm...

0xda157 (dismissed):

fission/src/bench/baseline.json
azaleacolburn: Oh that's a problem, this file should probably be ignored. Also This PR wasn't marked for review.

fission/src/bench/baseline.json
azaleacolburn: Because this PR wasn't marked for review

fission/src/bench/PhysicsSystem.bench.ts
```diff
@@ -0,0 +1,128 @@
...
+    bench("rayCast — miss", () => {
+        raycastSystem.rayCast(RAY_FROM, RAY_MISS_DIR, false)
+    })
+})
+
+describe("PhysicsSystem — assembly spawn", () => {
+    // createMechanismFromParser allocates a LayerReserve (one of 8 robot slots).
+    // destroyMechanism removes bodies and constraints but does not release the slot,
+    // so we do it manually to keep the pool from exhausting across iterations.
+    bench("spawn + destroy Dozer (7 bodies, 6 joints)", () => {
```
PepperLola: Why include the number of bodies and joints? This feels like it could easily drift if we ever update the assembly. If we want to print this diagnostic information then would it make sense to load it from the mechanism?

PepperLola (commented):
Some of the comments seem a verbose or unnecessary but otherwise looks good.

AlexD717 (approved):
lgtm, some of the comments look a bit unnecessary but that doesn't really matter

fission/src/bench/PhysicsSystem.bench.ts
azaleacolburn: I agree, this should be changed

PepperLola (approved):

